PR #217, opened 2014-08-19
server._is_satisfied should not travel all the paths.

damz:
We just found out that the way `ProtocolGraphWalker._is_satisfied` is currently implemented is very inefficient, and that would be the understatement of the year.

The implementation travels all the paths in the commit graph that lead from the want to one have.

It is usually not a problem, but we have a relatively small repository (less than 400 commits), with an arguably weird history of cross-merges, that has `2.1e22` paths between its HEAD commit and its root commit.

A really good start is to maintain the set of already known nodes (i.e. nodes that are either pending or already visited).

### DIFF
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -443,6 +443,7 @@ def _find_shallow(store, heads, depth):
 def _want_satisfied(store, haves, want, earliest):
     o = store[want]
     pending = collections.deque([o])
+    known = set([want])
     while pending:
         commit = pending.popleft()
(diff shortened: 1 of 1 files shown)


damz:
Rebased. We have been running this on the Platform.sh Git service forever now. There are several other functions in the code base exploring the commit graph and using a similar set (for example: `dulwich.object_store.ObjectStoreGraphWalker`).

Please merge.

jelmer:
Merged, at last..

PR #231, opened 2014-11-05
Fix redundant parse

keis:
Before this retrieving a loose tree object from a repository would cause it be parsed three times. objects stored in a pack did not show this behaviour.

Setting the raw string no longer implies parsing the data which has the possibly problematic side effect that errors will only occur when you actually try to use the object.


jelmer:
Thanks for the report; I've fixed this but in a slightly simpler manner, and added a test to prevent regressions.

PR #241, opened 2014-12-28
Remove leading '/' from path in FileSystemBackend

valberg:
When doing `git clone git://localhost/foo.git` I get an `Invalid path /foo.git` exception even though `self.root = '/home/git/repositories'`. This is because the path given is `/foo.git` and that makes the call to `os.path.join` have no effect.

Don't know if I'm using git clone the wrong way, but this PR fixes this issue when using `dulwich daemon`.

### DIFF
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -169,6 +169,10 @@ class FileSystemBackend(Backend):
         self.root = (os.path.abspath(root) + "/").replace("//", "/")
 
     def open_repository(self, path):
+        # If path starts with a '/' the os.path.join function has no effect, so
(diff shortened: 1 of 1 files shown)


jelmer:
There is no requirement that a path with a leading / should fail, but if you specify a path with a leading slash it should match the local absolute path.

E.g. this should work:

git://localhost/bar
git://localhost//full/path/to/bar

Jelmer

jelmer:
I'm open to discussion about changing the current behaviour, but please provide matching tests if you make changes to the behaviour.

jelmer:
closing for now, feel free to reopen

PR #243, opened 2015-01-12
Use a block cache in _find_content_rename_candidates().

miwilliams:
This will improve the performance of calls to changes_with_renames(), (which calls _find_content_rename_candidates), by avoiding re-hashing in _similarity_score().


jelmer:
Thanks, Mike!

Issue #247, opened 2015-01-21
TypeError when fetching empty updates into MemoryRepo


hweemiin:
Attempt to fetch updates from remote repository into a MemoryRepo, when there are no updates (MemoryRepo already up-to-date), following error occurs:
``` 
Traceback (most recent call last):
  File "E:\test\python\test.py", line 30, in <module> do_fetch(repo)
  File "E:\test\python\test.py", line 22, in do_fetch client.fetch(path, repo)
  File "C:\Python27\lib\site-packages\dulwich\client.py", line 223, in fetch commit()
  File "C:\Python27\lib\site-packages\dulwich\object_store.py", line 752, in commit for obj in PackInflater.for_pack_data(p):
  File "C:\Python27\lib\site-packages\dulwich\pack.py", line 1238, in for_pack_data for unpacked in pack_data._iter_unpacked():
  File "C:\Python27\lib\site-packages\dulwich\pack.py", line 1097, in _iter_unpacked for _ in xrange(self._num_objects):
TypeError: an integer is required
```
Here is the code to reproduce the issue:
```
repo = MemoryRepo()
client, path = get_transport_and_path('https://github.com/jelmer/dulwich.git')
client.fetch(path, repo)
client, path = get_transport_and_path('https://github.com/jelmer/dulwich.git')
client.fetch(path, repo)
```

PR #248, opened 2015-01-21
Fix issue #247 'TypeError when fetching empty updates into MemoryRepo'

hweemiin:
Added check for _num_objects is None when iterating pack data.


jelmer:
This was merged.

PR #252, opened 2015-02-19
clean up clone target on failure

samstav:
If 'target' does not already exist, the clone()
function will create it. If the clone() function
creates the dir but the clone ultimately fails,
the dir should be cleaned up.

...

Checking this with git version 2.1.2, and it looks
like `git clone` will actually remove the target dir
on failure *even if* it was there before the command
was run. That doesn't sound right, but this could be
changed to behave identically to git.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -49,6 +49,7 @@ __docformat__ = 'restructuredText'
 
 from collections import namedtuple
 import os
+import shutil
 import sys
 import time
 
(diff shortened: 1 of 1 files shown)


jelmer:
On Wed, Feb 18, 2015 at 08:36:23PM -0800, Sam Stavinoha wrote:
> If 'target' does not already exist, the clone()
> function will create it. If the clone() function
> creates the dir but the clone ultimately fails,
> the dir should be cleaned up.
> 
> ...
> 
> Checking this with git version 2.1.2, and it looks
> like `git clone` will actually remove the target dir
> on failure *even if* it was there before the command
> was run. That doesn't sound right, but this could be
> changed to behave identically to git.

Thanks for the patch. I agree dulwich shouldn't be removing a directory
it hasn't created.  Can you please also add a test for the new behaviour?

Cheers,

Jelmer

jelmer:
Closing this as the issue seems to have stalled. Please reopen when you followup.

PR #253, opened 2015-02-19
Implement LocalGitClient.send_pack

diryboy:
Currently hg-git push/pull to/from local git repo will fail if no git was installed, implemented this so hg-git can use LocalGitClient for local scheme instead of SubprocessGitClient

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -679,7 +679,29 @@ class LocalGitClient(GitClient):
         :raises UpdateRefsError: if the server supports report-status
                                  and rejects ref updates
         """
-        raise NotImplementedError(self.send_pack)
+        from dulwich.repo import Repo
+        from dulwich.protocol import ZERO_SHA
+        
+        target = Repo(path)
+        old_refs = target.get_refs()
+        if not old_refs:
+            old_refs["capabilities^{}"] = ZERO_SHA
+        new_refs = determine_wants(old_refs)
+
+        have = [sha1 for sha1 in old_refs.values() if sha1 != ZERO_SHA]
+        want = []
+        for refname in set(new_refs.keys() + old_refs.keys()):
+            old_sha1 = old_refs.get(refname, ZERO_SHA)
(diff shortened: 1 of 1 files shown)


dulwich/client.py
```diff
@@ -679,7 +679,29 @@ def send_pack(self, path, determine_wants, generate_pack_contents,
...
                                  and rejects ref updates
         """
-        raise NotImplementedError(self.send_pack)
+        from dulwich.repo import Repo
+        from dulwich.protocol import ZERO_SHA
+        
+        target = Repo(path)
+        old_refs = target.get_refs()
+        if not old_refs:
+            old_refs["capabilities^{}"] = ZERO_SHA
```
jelmer: This is protocol specific and shouldn't be necessary ^

jelmer:
Thanks!

Can you please also add a test or two for this new function? You should be able to reuse some of the existing send_pack tests in dulwich/tests/test_client.py.

dulwich/client.py
diryboy: I basically copied the implementation of [HttpGitClient.send_pack](https://github.com/jelmer/dulwich/blob/master/dulwich/client.py#L1007), the remote response containing this special key directly is passed to `determine_wants` next line, which is a callback passed in from outside, existing code from hg-git already depends on this key exists.

So removing the key need other types of clients do the same, and hg-git need to patch `determine_wants`.

jelmer: If the capabilities^ ref is leaked into the refs returned to the user somewhere, that is a bug. It is protocol specific, and an abstraction leak. I'll file a bug about this, and will file a bug with hg-git too about its dependency on this behaviour.

diryboy:
Wait, sorry I actually have questions when writing test, did not managed to save some time to write to you.

The problem of this implementation is that it does not update the refs in the target repo, leaving the new objects added garbage collectable.

jelmer:
sorry, that wasn't actually meant to go out yet..

PR #257, opened 2015-03-11
Resolve delta refs when pulling into a MemoryRepo.

### LINKED ISSUE #256: Pulling to a MemoryRepo (after initial clone) fails
What's the correct way to pull from a remote to a MemoryRepo? The client.fetch() approach suggested in porcelain.py seems to fail in a variety of ways. 

* Cloning then immediately pulling:
```
> import dulwich.repo, dulwich.client
> repo = dulwich.repo.MemoryRepo()
> client, path = dulwich.client.get_transport_and_path('https://github.com/jelmer/dulwich.git')
> branches = client.fetch(path, repo)
> # Reusing client here throws an HTTP 403, but that's easy to work around.
> client, path = dulwich.client.get_transport_and_path('https://github.com/jelmer/dulwich.git')
> branches = client.fetch(path, repo)
  File ...
    branches = client.fetch(path, repo)
  File "dulwich/client.py", line 223, in fetch
    commit()
  File "dulwich/object_store.py", line 748, in fetch
    for obj in PackInflater.for_pack_data(p):
  File "dulwich/pack.py", line 1238, in fetch
    for unpacked in pack_data._iter_unpacked():
  File "dulwich/pack.py", line 1097, in fetch
    for _ in xrange(self._num_objects):
TypeError: an integer is required
```
* Cloning, applying a commit, pushing then pulling. I don't have a succinct reproduction here, but it succeeds up to and including the push, then fails on the pull with this traceback:
```
  File ...
    branches = client.fetch(path, repo, determine_wants=DetermineWants)
  File "dulwich/client.py", line 223, in fetch
    commit()
  File "dulwich/object_store.py", line 748, in commit
    for obj in PackInflater.for_pack_data(p):
  File "dulwich/pack.py", line 1264, in _walk_all_chains
    for result in self._walk_ref_chains():
  File "dulwich/pack.py", line 1274, in _walk_ref_chains
    self._ensure_no_pending()
  File "dulwich/pack.py", line 1270, in _ensure_no_pending
    raise KeyError([sha_to_hex(s) for s in self._pending_ref])
KeyError: ['6f31271c006a661440d89d3d524332595dfc6970']
# This is the SHA of a tree that was committed locally and pushed earlier.
```

All this is on v0.9.8.

max99x:
Fixes #256.

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -745,7 +745,7 @@ class MemoryObjectStore(BaseObjectStore):
         def commit():
             p = PackData.from_file(BytesIO(f.getvalue()), f.tell())
             f.close()
-            for obj in PackInflater.for_pack_data(p):
+            for obj in PackInflater.for_pack_data(p, self.get_raw):
                 self._data[obj.id] = obj
         def abort():
             pass


max99x:
It looks like packs with deltas are supposed to be passed to add_thin_pack(), instead of add_pack(), but I only see it being called in tests and in the server code, which leaves me confused - shouldn't the client support them too? It seems to work with the above fix which resolve refs in add_pack().

jelmer:
On Wed, Mar 11, 2015 at 03:43:01PM -0700, Max Shawabkeh wrote:
> It looks like packs with deltas are supposed to be passed to add_thin_pack(), instead of add_pack(), but I only see it being called in tests and in the server code, which leaves me confused - shouldn't the client support them too? It seems to work with the above fix which resolve refs in add_pack().

Merged, thanks.

Jelmer

PR #261, opened 2015-03-25
Skip three tests needing --no-single-branch for older git versions

metatoaster:
The --[no-]single-branch] flag only available in git version 1.7.10.

Running the tests on a slightly older version of git failed on tests that called the binary with those flags. This patch marks those tests to be skipped using the provided utilities.

See:

http://git-scm.com/docs/git-clone/1.7.9.7
http://git-scm.com/docs/git-clone/1.7.10


jelmer:
Merged, thanks!

Issue #88, opened 2013-04-09
http protocol error: git fetch_pack: expected ACK/NAK, got 'dul-daemon says what'


asaladin:
Hi,

This description follows the previous bug report on launchpad: https://bugs.launchpad.net/dulwich/+bug/953916

I've been able to reproduce the same error on a public git repository, so it can now easily be reproduced. 
I can reproduce this problem with at least two versions of dulwich: the current pypi version and the latest github master version at 2a8548be3b1fd4a (Dec 04, 2012). 

It happens during a git pull operation of a repository served by http.
Here are the steps to reproduce:

```
$ cd /tmp 
$ git clone https://github.com/asaladin/testdul_repo.git ptools
$ git clone https://github.com/asaladin/testdul_git2.git git2

$ cat test_dulwich/test_dulwich.py
##################
from dulwich.repo import Repo
from dulwich.web import HTTPGitApplication
from dulwich import server as dulserver

repo_path = "/tmp/ptools"

_d = {'/repo' : Repo(repo_path)}
backend = dulserver.DictBackend(_d)
gitserve = HTTPGitApplication(backend)

def app(environ, start_response):
   return gitserve(environ, start_response)
##################

#run on another shell with: $ gunicorn test_dulwich:app
$ cd git2
$ git pull http://localhost:8000/repo 
fatal: git fetch_pack: expected ACK/NAK, got 'dul-daemon says what'
```

cheers,

jelmer:
Fixed in master.

PR #262, opened 2015-03-28
Complete fix for issue #88

metatoaster:
The gist of this gnarly problem is that it looks like the implementation over HTTP behaves as-if the `no-done` capability was enabled but unspecified so the reference client would make a second round trip back with more `haves` but then it doesn't know what to do with the `PACK` that was already sent. Alternatively there are cases where the side-band was used when it was expecting ACK/NAK. There were multiple problems with the code base that made this issue rather hard to stomp out, and here are the fixes (some verbatim from commit messages).

- `MultiAckDetailedGraphWalkerImpl` rebuilt to ensure output generated
  matches the reference implementation, like the git-http-backend. (Even though this is not strictly necessary, repeatedly calling `all_wants_specified` doesn't really change much given that there's only one place where the commits can be added in, which is by the object store through the `ack` method.

- Correction to the communication after all `have` lines have been processed. As a freebie I got `no-done` capability implemented since that's what it looked like before.

- There is another issue where attempts to pull an unrelated repo, the
  various combination of the previous flaws made it difficult to send
  the correct number of NAKs.  Removing the walker.send_ack call from
  the walker implementation and just ensure the ack is the only method
  that will call send_ack simplifies the process of correcting this.

- Added various state variables to the handler to track whether the `done` token is expected.

- Added a couple methods to deal with the handling of the above state variables, which the walker implementations themselves have access to through a method provided by the generic protocol walker class

- Added a method to the graph walker that takes in the state variables provided by the handler to delegate the dealing of the final ACK/NAK line that ends the section to permit the PACK section to begin, which then gets delegated to the Impl instances (which also have been normalized to address this).

- Finally, cleaned up the test cases and added further tests where relevant. Naturally 
one of the earlier commits introduce some test examples that demonstrates this problem.

### DIFF
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -282,11 +282,15 @@ class UploadPackHandler(Handler):
         graph_walker = ProtocolGraphWalker(self, self.repo.object_store,
             self.repo.get_peeled)
         objects_iter = self.repo.fetch_objects(
-          graph_walker.determine_wants, graph_walker, self.progress,
-          get_tagged=self.get_tagged)
+            graph_walker.determine_wants, graph_walker, self.progress,
+            get_tagged=self.get_tagged)
 
         # Did the process short-circuit (e.g. in a stateless RPC call)? Note
         # that the client still expects a 0-object pack in most cases.
+        # Also, if it also happens that the object_iter is instantiated
+        # with a graph walker with an implementation that talks over the
+        # wire (which is this instance of this class) this will actually
+        # iterate through everything and write things out to the wire.
         if len(objects_iter) == 0:
             return
 
@@ -686,6 +690,11 @@ class MultiAckDetailedGraphWalkerImpl(object):
             if command is None:
                 self.walker.send_nak()
                 if self.walker.http_req:
+                    # why special case?  if this was "stateless" as in
+                    # HTTP there would be nothing left to be read and
+                    # None would have returned later... so I guess this
+                    # is a short circuit (also to make the test case
+                    # as written happy).
                     return None
                 continue
             elif command == 'done':
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -660,6 +660,8 @@ class BaseGraphWalkerImpl(object):
             # Okay we are not actually done then since the walker picked
             # up no haves.  This is usually triggered when client attempts
             # to pull from a source that has no common base_commit.
+            # See: test_server.MultiAckDetailedGraphWalkerImplTestCase.\
+            #          test_multi_ack_stateless_nodone
             return False
 
         self.post_nodone_check()
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -242,7 +242,6 @@ class UploadPackHandler(Handler):
         # data (such as side-band, see the progress method here).
         self._processing_have_lines = False
 
-    @property
     def _done_required(self):
         return not self.has_capability("no-done")
 
@@ -317,7 +316,7 @@ class UploadPackHandler(Handler):
         self._processing_have_lines = False
 
         if not graph_walker.handle_done(
-                self._done_required, self._done_received):
+                self._done_required(), self._done_received):
             return
 
         self.progress("dul-daemon says what\n")
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -239,7 +239,7 @@ class UploadPackHandler(Handler):
         self.advertise_refs = advertise_refs
         # A state variable for denoting that the have list is still
         # being processed, and the client is not accepting any other
-        # side-band data.
+        # data (such as side-band, see the progress method here).
         self._processing_have_lines = False
 
     @property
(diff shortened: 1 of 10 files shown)


metatoaster:
It should pass now, fixed that one minor error with automatic merges not working because lack of default user name and email on the Travis git installation (go figure). However `apt-get` threw a fit so the tests failed for the most part for the second round, but one of them did pass on Travis.

metatoaster:
Pushed one more minor change to clarify some comments and minor clean up to trigger the build again. As the title doesn't get parsed for references to other issues I have to say again that this provides a fix for issue #88.

jelmer:
Thanks for working on this, and the patch! I'll take some time next week to properly review it, but I've added some first comments below.

dulwich/server.py
```diff
@@ -232,18 +237,26 @@ def __init__(self, backend, args, proto, http_req=None,
         self.repo = backend.open_repository(args[0])
         self._graph_walker = None
         self.advertise_refs = advertise_refs
+        # A state variable for denoting that the have list is still
+        # being processed, and the client is not accepting any other
+        # data (such as side-band, see the progress method here).
+        self._processing_have_lines = False
+
+    @property
+    def _done_required(self):
```
jelmer: This should not be a property - we don't have properties for any of the other capabilities.

dulwich/tests/compat/server_utils.py
```diff
@@ -237,6 +237,54 @@ def test_fetch_full_depth_into_shallow_clone_from_dulwich(self):
...
+                        cwd=self._client_repo.path)
+        # the merge commit will be automatic so hashes are different,
+        # however it being a working directory and done with git it has
+        # a rather predictable result, assuming a default strategy.
+        self.assertEqual(len(os.listdir(self._client_repo.path)), 8)
+
+    def test_push_to_dulwich_issue_88_standard(self):
+        # Same thing, but we reverse the role of the server/client
+        # and do a push instead.
+        self._source_repo = import_repo('issue88_expect_ack_nak_client.export')
```
jelmer: Please avoid adding more fast-import repositories, and build repository using the utility functions instead.

fast-export streams are hard to read and debug.

dulwich/server.py
metatoaster: You mean the `_done_required` flag that is used only internally here that reports whether the done capability is enabled for this particular request? Fine, changing it to a method call then.

dulwich/tests/compat/server_utils.py
```diff
@@ -237,6 +237,54 @@ def test_fetch_full_depth_into_shallow_clone_from_dulwich(self):
...
+    def test_fetch_from_dulwich_issue_88_alternative(self):
+        # likewise, but the case where the two repos have no common parent
+        self._source_repo = import_repo('issue88_expect_ack_nak_other.export')
+        self.addCleanup(tear_down_repo, self._source_repo)
+        self._client_repo = import_repo('issue88_expect_ack_nak_client.export',
+                                        as_working_copy=True)
+        self.addCleanup(tear_down_repo, self._client_repo)
+        port = self._start_server(self._source_repo)
+
+        run_git_or_fail(['config', 'user.email', 'user@example.com'],
```
jelmer: Why does this need a merge? I'd rather avoid stuff that creates new commits using git proper, and it doesn't seem to be necessary here.

dulwich/tests/compat/server_utils.py
```diff
@@ -237,6 +237,54 @@ def test_fetch_full_depth_into_shallow_clone_from_dulwich(self):
         self.assertEqual([], _get_shallow(clone))
         self.assertReposEqual(clone, self._source_repo)
 
+    def test_fetch_from_dulwich_issue_88_standard(self):
```
jelmer: FWIW the compat tests are only supposed to test compatibility with C git, they're not the right place for integration tests, which should probably be in dulwich/tests/test_server.py

dulwich/tests/compat/server_utils.py
metatoaster: This is needed because there are cases where merges of two unrelated repositories happen, and is needed to debug the issue of extra NAKs (or not enough of them I forgot which) depending on the no-done capability.  I supposed this can be replaced with a fetch if it can reproduce the same issue.

dulwich/tests/compat/server_utils.py
metatoaster: This bug was not easy to debug and it needs the very minimum 16 + 1 commits to generate enough `have` lines to trigger the bug in dulwich which in turn cause standard git client to error out on push.  Perhaps if there is a way to automatically convert that into the utility function calls that can reproduce this bug without the features added, because I've already put far more time into this than I ever wanted to.

dulwich/tests/compat/server_utils.py
jelmer: That'd be great - fetch should do the same thing as pull, protocol-wise

dulwich/tests/compat/server_utils.py
jelmer: I do really appreciate your taking the time to debug and fix this. This is also precisely why I care about readability of the code. :-)

The utility functions should be fairly easy to use. See dulwich.tests.utils, in particular make_commit() and build_commit_graph()

For an example, see e.g. dulwich/tests/test_missing_obj_finder.py:75

dulwich/tests/compat/server_utils.py
metatoaster: Actually not NAKs, but PAKs, but yeah pushing new changes for this.

dulwich/tests/compat/server_utils.py
metatoaster: Do not that I have made changes to `dulwich/tests/test_server.py` that corrects the responses to the expected answers in 145c6ddb80d0eb7f8be03dc0e0c0f5dfa0b75d30 and added further tests commit 3c276c4edceb5243f933dbc38fbe69b4e5ee8c70 to precisely test the behaviour with multiack-detailed and no-done. However reviewing my own code again I think I may have missed one behaviour which I must test now.

jelmer:
Merged, in modified form. Thanks!

jelmer:
I've left the .export files in for now, but we should really remove those in favor of a couple of lines of Python code that just creates those commits.

PR #264, opened 2015-03-31
Most `ShaFile` objects don't cache their SHA properly

damz:
The `self._needs_serialization` flag is not cleared properly inside `Tree._serialize`, `Commit._serialize` and `Tag._serialize`, which forces those objects to compute their SHA over and over again. It is actually sounds more logical to clear the flag directly inside `ShaFile.as_raw_chunks()`, which is responsible for calling `ShaFile._serialize()`.

There are no tests covering this that I can see, so I added some.


jelmer:
Merged, thanks!

Issue #263, opened 2015-03-30
UnicodeDecodeError in 0.10


kiilerix:
I see unicode issues with dulwich 0.10 and later when creating a repo with a unicode name. (Linux, Python 2.7.) 

```
...
  File ".../site-packages/dulwich/repo.py", line 899, in init_bare
    return cls._init_maybe_bare(path, True)
  File ".../site-packages/dulwich/repo.py", line 871, in _init_maybe_bare
    ret.refs.set_symbolic_ref("HEAD", "refs/heads/master")
  File ".../site-packages/dulwich/refs.py", line 544, in set_symbolic_ref
    filename = self.refpath(name)
  File ".../site-packages/dulwich/refs.py", line 434, in refpath
    return os.path.join(self.path, name)
  File ".../posixpath.py", line 80, in join
    path += '/' + b
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 9: ordinal not in range(128)
```

refpath's "name" used to be 'HEAD' but after https://github.com/jelmer/dulwich/commit/2a69962f25bab502f266eda1e5d9eb567830f091#diff-68db177312d4b6a6396be97b3def4771R433 it is u'HEAD' while self.path according to https://github.com/jelmer/dulwich/commit/72d224fc0faae6f08a6b64cbb2ebce50ce59ad0c is a utf-8 encoded bytestring - these two string types are incompatible and cannot be path joined.

PR #268, opened 2015-04-16
Fix for Issue 263 

garyvdm:
This adds a test for Issue #263, and revert the changes which caused the bug.

The test is very superficial. For more extensive testing, I've found I can set the `TEMP` env var to a path with unicode chars in it and run the tests.

Any way, the test does error without the fix, an so will hopefully catch regressions like this in the future.

Looking to the future, I would like to maybe ensure that the `Repo._controldir`  and maybe also `Repo.path` are bytes in `Repo.__init__`, so that we can handle `Repo(u'...')` and `Repo(b'...')` in python3.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -106,8 +106,6 @@ def hex_to_filename(path, hex):
     # os.path.join accepts bytes or unicode, but all args must be of the same
     # type. Make sure that hex which is expected to be bytes, is the same type
     # as path.
-    if getattr(path, 'encode', None) is not None:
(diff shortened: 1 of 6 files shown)


jelmer:
This is adding back skipIfPY3 in a couple of places, does it really need to?

garyvdm:
I was adding back the skipIfPY3 because the temporary fix was to revert the python3 fixes, which broke the relevant tests in python3.

I'm working on a more permanent fix now.

garyvdm:
Superseded by PR /271

PR #270, opened 2015-04-22
Appveyor: Windows CI

garyvdm:
This adds an appveyor config file. To get it running, you (Jelmer) will need to create and account on appveyor (which is free of charge for FLOSS projects) and add a project for dulwich. 

This is what the output will look like: https://ci.appveyor.com/project/garyvdm/dulwich

This also disables some things that were preventing the tests from completing.  Fixes for these issues will still needed in the future.

In the future, we can use appveyor to build binaries for windows. (Issue 222)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,14 @@
+environment:
+  matrix:
+    - PYTHON: "C:\\Python27"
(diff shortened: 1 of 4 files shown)


jelmer:
Thanks :)

PR #271, opened 2015-04-22
Change the way methods that take a path argument behave. 

### LINKED ISSUE #263: UnicodeDecodeError in 0.10

garyvdm:
Fixes #263

For a number of user facing methods, make it possible to pass in paths in as either a unicode string, or bytes. These include:

* `Repo.__init__`
* `Repo.init`
* `Repo.init_bare`
* `index.build_index_from_tree`
* `index.get_unstaged_changes`

The repo.Repo.path attribute will remain unchanged. The Repo._controldir attribute and a new Repo._path_bytes attribute are ensured to be bytes.

For a number of internal methods, it now requires a bytes path rather than
a unicode string. These include:

* `objects.hex_to_filename`
* `objects.filename_to_hex`
* `_GitFile.__init__`
* `ShellHook.__init__` (and subclassed of ShellHook)
* `DiskObjectStore.__init__`
* `DiskRefsContainer.__init__`
* `Repo._put_named_file`
* `Repo.get_named_file`
* `pack.write_pack`
* `Pack.__init__`
* etc...

For the repository tests, I've doubled up the test so that they all run with both a unicode, and bytes path for their root. There my be other places where this needs to happen.

P.S. When I started out this, I thought it would be a small patch. That was wishful thinking.

### DIFF
--- a/dulwich/file.py
+++ b/dulwich/file.py
@@ -104,7 +104,7 @@ class _GitFile(object):
                      'truncate', 'write', 'writelines')
     def __init__(self, filename, mode, bufsize):
         self._filename = filename
-        self._lockfilename = '%s.lock' % self._filename
+        self._lockfilename = self._filename + b'.lock'
         fd = os.open(self._lockfilename,
             os.O_RDWR | os.O_CREAT | os.O_EXCL | getattr(os, "O_BINARY", 0))
         self._file = os.fdopen(fd, mode, bufsize)
--- a/dulwich/hooks.py
+++ b/dulwich/hooks.py
@@ -100,7 +100,7 @@ class PreCommitShellHook(ShellHook):
     """pre-commit shell hook"""
 
     def __init__(self, controldir):
-        filepath = os.path.join(controldir, 'hooks', 'pre-commit')
+        filepath = os.path.join(controldir, b'hooks', b'pre-commit')
 
         ShellHook.__init__(self, 'pre-commit', filepath, 0)
 
@@ -109,7 +109,7 @@ class PostCommitShellHook(ShellHook):
     """post-commit shell hook"""
 
     def __init__(self, controldir):
-        filepath = os.path.join(controldir, 'hooks', 'post-commit')
+        filepath = os.path.join(controldir, b'hooks', b'post-commit')
 
         ShellHook.__init__(self, 'post-commit', filepath, 0)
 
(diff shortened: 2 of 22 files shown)


jelmer:
How does this work on Windows, where paths aren't bytestrings?

garyvdm:
Provided that your byte sting is correctly encoded, it just works. (I assume that python decode the byte sting in the background, but I'm not sure.)

    C:\Users\GaryvdM\test>C:\Python27\python.exe
    Python 2.7.9 (default, Dec 10 2014, 12:24:55) [MSC v.1500 32 bit (Intel)] on win32
    >>> import sys
    >>> sys.getfilesystemencoding()
    'mbcs'
    >>> path = u'délwíçh'.encode('mbcs')
    >>> path
    'd\xe9lw\xed\xe7h'
    >>> with open(path, 'w') as f:
    ...     f.write('foo')
    ...
    >>> exit()
    
    C:\Users\GaryvdM\test>dir
     Volume in drive C has no label.
     Volume Serial Number is 2ECB-CF15
    
     Directory of C:\Users\GaryvdM\test
    
    2015-04-23  10:07 AM    <DIR>          .
    2015-04-23  10:07 AM    <DIR>          ..
    2015-04-23  10:07 AM                 3 délwíçh
                   1 File(s)              3 bytes
                   2 Dir(s)  216 798 289 920 bytes free

(Tested with 2.7 & 3.4)

PR #272, opened 2015-04-22
Buffer the subpocess.stdout files for SubprocessGitClient on python3.

garyvdm:
We have been getting intermittent `GitProtocolError:  Length of pkt read....` errors on Python 3. e.g.: https://travis-ci.org/jelmer/dulwich/jobs/59521569 .  

I can reliably reproduces this by running multiple instances of the test suite from pycharm. I think it happens when the system IO queues are long.

This is due to a change in the way `subprocess.POpen(bufsize=0` works in python3 vs python2. See. http://bugs.python.org/issue17488 

We still need a `bufsize=0` on the stdin, but need the stdout to be buffered. This patch dose this by wrapping the stderr in a io.BufferedReader.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -38,7 +38,7 @@ Known capabilities that are not supported:
 
 __docformat__ = 'restructuredText'
 
-from io import BytesIO
+from io import BytesIO, BufferedReader
 import dulwich
 import select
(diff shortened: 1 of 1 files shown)


jelmer:
Merged, thanks!

PR #275, opened 2015-04-28
Removing .git dirs requires a bit more effort on windows.

garyvdm:
shutils.rmtree on .git dirs fails on windows due to the read only attribute being set, so we need to do some additional work to unset the attribute.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -725,7 +725,7 @@ class LocalGitClient(GitClient):
 
         with closing(Repo(path)) as target:
             old_refs = target.get_refs()
-            new_refs = determine_wants(old_refs)
+            new_refs = determine_wants(dict(old_refs))
 
             have = [sha1 for sha1 in old_refs.values() if sha1 != ZERO_SHA]
             want = []
@@ -779,10 +779,6 @@ class LocalGitClient(GitClient):
             if objects_iter is None:
                 return
             write_pack_objects(ProtocolFile(None, pack_data), objects_iter)
-
-
-# What Git client to use for local access
-default_local_git_client_cls = SubprocessGitClient
 
 
 class SSHVendor(object):
@@ -1133,17 +1129,23 @@ class HttpGitClient(GitClient):
         finally:
             resp.close()
 
+# What Git client to use for local access
+default_local_git_client_cls = LocalGitClient
 
-def get_transport_and_path_from_url(url, config=None, **kwargs):
+def get_transport_and_path_from_url(url, local_git_client_cls=None, config=None, **kwargs):
     """Obtain a git client from a URL.
 
     :param url: URL to open
+    :param local_git_client_cls: Class to use for local (file) urls.
     :param config: Optional config object
     :param thin_packs: Whether or not thin packs should be retrieved
     :param report_activity: Optional callback for reporting transport
         activity.
     :return: Tuple with client instance and relative path.
     """
+    if local_git_client_cls is None:
+        local_git_client_cls = default_local_git_client_cls
+
     parsed = urlparse.urlparse(url)
     if parsed.scheme == 'git':
         return (TCPGitClient(parsed.hostname, port=parsed.port, **kwargs),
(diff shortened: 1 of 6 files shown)


jelmer:
This is specific to the compat tests I think? Dulwich itself doesn't create read-only directories. Please move it to dulwich/tests/compat/utils.py

jelmer:
Please use a different name, as this is doing something different than shutil.rmtree.

garyvdm:
> This is specific to the compat tests I think? Dulwich itself doesn't create read-only directories. Please move it to dulwich/tests/compat/utils.py

Sorry, I was painting with a very broad brush. I've removed a bunch of unnecessary changes. It is still need in some of the non compat tests, e.g. `dulwich.tests.test_porcelain.PushTests.test_simple`, so I've moved the new method to `dulwich/tests/utils.py` rather than `dulwich/tests/compat/utils.py`. 

> Please use a different name, as this is doing something different than shutil.rmtree.

Done.

garyvdm:
Sorry. Made a rebase error, and included other stuff. Fixed now.

dulwich/tests/test_porcelain.py
```diff
@@ -47,7 +48,7 @@ class PorcelainTestCase(TestCase):
     def setUp(self):
         super(TestCase, self).setUp()
         repo_dir = tempfile.mkdtemp()
-        self.addCleanup(shutil.rmtree, repo_dir)
+        self.addCleanup(rmtree_ro, repo_dir)
```
jelmer: why is this necessary? dulwich doesn't mark any files readonly AFAIK.

garyvdm: It's needed for `test_porcelain.PushTests.test_simple` which calls a `SubprocessGitClient` when it calls `push`. 

We can change the `default_local_git_client_cls` to `LocalGitClient` for this test, but I don't think there is any harm in using `rmtree_ro` here.

garyvdm:
Opps, I seem to have confused travis, and it's not testing the right sha. Here is a test result of the correct sha on my account: https://travis-ci.org/garyvdm/dulwich/builds/60458018

jelmer:
Thanks for the updates. I think we should update the default push to use the LocalGitClient by default as well. The noncampat testsuite should still work and behave the same independent of whether C git is installed.

On 28 April 2015 11:54:17 pm BST, Gary van der Merwe <notifications@github.com> wrote:
>> This is specific to the compat tests I think? Dulwich itself doesn't
>create read-only directories. Please move it to
>dulwich/tests/compat/utils.py
>
>Sorry, I was painting with a very broad brush. I've removed a bunch of
>unnecessary changes. It is still need in some of the non compat tests,
>e.g. `dulwich.tests.test_porcelain.PushTests.test_simple`, so I've
>moved the new method to `dulwich/tests/utils.py` rather than
>`dulwich/tests/compat/utils.py`. 
>
>> Please use a different name, as this is doing something different
>than shutil.rmtree.
>
>Done.
>
>
>
>---
>https://github.com/jelmer/dulwich/pull/275#issuecomment-97250478

Jelmer Vernooij <jelmer@jelmer.co.uk> - https://jelmer.co.uk/

garyvdm:
> I think we should update the default push to use the LocalGitClient by default as well.

Done.

`rmtree_ro` is still in `tests/utils.py` and not `tests/compat/utils.py` as `dulwich.tests.compat.test_repository.ObjectStoreTestCase` uses `test.utils.tear_down_repo`.

jelmer:
On Thu, Apr 30, 2015 at 02:59:59AM -0700, Gary van der Merwe wrote:
> > I think we should update the default push to use the LocalGitClient by default as well.
> 
> Done.
Thanks!

> `rmtree_ro` is still in `tests/utils.py` and not `tests/compat/utils.py` as `dulwich.tests.compat.test_repository.ObjectStoreTestCase` uses `test.utils.tear_down_repo`.  
Fair enough. 

I'll merge this evening.

dulwich/client.py
```diff
@@ -704,7 +704,7 @@ def send_pack(self, path, determine_wants, generate_pack_contents,
 
         target = Repo(path)
         old_refs = target.get_refs()
-        new_refs = determine_wants(old_refs)
+        new_refs = determine_wants(dict(old_refs))
```
jelmer: why is this necessary?

dulwich/tests/test_client.py
```diff
@@ -408,6 +408,12 @@ def test_http(self):
         self.assertTrue(isinstance(c, HttpGitClient))
         self.assertEqual('/jelmer/dulwich', path)
 
+    def test_local_specified_cls(self):
+        c, path = get_transport_and_path(
+            'foo.bar/baz', local_git_client_cls=SubprocessGitClient)
```
jelmer: won't this break on systems without C git installed?

dulwich/tests/test_porcelain.py
```diff
@@ -465,12 +465,12 @@ def test_simple(self):
         # Check that the target and source
         r_clone = Repo(clone_path)
 
+        self.assertEqual(r_clone[b'HEAD'].id, self.repo[refs_path].id)
```
jelmer: why does this ordering need to change?

jelmer:
Thanks, I've merged the first commit - added comments to the second.

dulwich/tests/test_client.py
garyvdm: No. This test passes because we are just asserting on the type of client returned. It never calls any methods on the clients, so the client never calls git.

This commit actually fixes a bunch of tests from `dulwich.tests.test_porcelain` that were failing when git was not installed because the default client was `SubprocessGitClient`, and we were not checking if it was installed with `require_git_version`.

dulwich/tests/test_porcelain.py
garyvdm: The test was failing at one point, and the failure message was a confusing `IndexError` from the `change = list(tree_changes(...))[0]` line because there were no changes.

The change in order is so we first check that the push happened by checking the `HEAD` changed, so that the failure message is more understandable.

garyvdm:
I've rebased the `Switched `default_local_git_client_cls` to `LocalGitClient`.` commit.

garyvdm:
Rebased onto master.

garyvdm:
Bump. There is still a commit from this branch not yet merged.

garyvdm:
Thanks.

PR #273, opened 2015-04-28
Close Repo/ObjectStore/Pack objects

garyvdm:
This is the first step of https://lists.launchpad.net/dulwich-users/msg00832.html. Please see that for motivation of this PR.

PR #278, opened 2015-05-03
Close Repo/ObjectStore/Pack objects alternative

garyvdm:
This PR is an alternative to PR #273. It dose not make ObjectStore and Repo object context managers. Rather it uses contextlib.closing.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -48,6 +48,10 @@ Differences in behaviour are considered bugs.
 __docformat__ = 'restructuredText'
 
 from collections import namedtuple
+from contextlib import (
+    closing,
+    contextmanager,
+)
 import os
 import sys
 import time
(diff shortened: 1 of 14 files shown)


dulwich/porcelain.py
```diff
@@ -222,17 +228,17 @@ def add(repo=".", paths=None):
...
-        # If nothing is specified, add all non-ignored files.
-        paths = []
-        for dirpath, dirnames, filenames in os.walk(r.path):
-            # Skip .git and below.
-            if '.git' in dirnames:
-                dirnames.remove('.git')
-            for filename in filenames:
-                paths.append(os.path.join(dirpath[len(r.path)+1:], filename))
-    r.stage(paths)
+    with closing(open_repo(repo)) as r:
```
jelmer: this kind of thing breaks if the user passes in an existing Repo object (which open_repo() supports), as it will close the repo for them.

dulwich/porcelain.py
```diff
@@ -200,6 +201,11 @@ def clone(source, target=None, bare=False, checkout=None, errstream=sys.stdout,
 
     if not os.path.exists(target):
         os.mkdir(target)
+
+    # XXX This is a bit messy. Who must close this repo, and what happens if
```
jelmer: This should be fine - just wrap the code below in "try / except" and close the repo if an exception is raised (and re-raise it).

garyvdm:
Comments have been addressed.

dulwich/porcelain.py
```diff
@@ -200,17 +220,22 @@ def clone(source, target=None, bare=False, checkout=None, errstream=sys.stdout,
...
-        r.reset_index()
+    try:
+        remote_refs = client.fetch(host_path, r,
+            determine_wants=r.object_store.determine_wants_all,
+            progress=errstream.write)
+        r[b"HEAD"] = remote_refs[b"HEAD"]
+        if checkout:
+            errstream.write(b'Checking out HEAD')
+            r.reset_index()
+    except Exception:
```
jelmer: specifically in this case you probably want "except:" rather than "except Exception", so you also catch KeyboardInterrupt, etc.

garyvdm:
> specifically in this case you probably want "except:" rather than "except Exception", so you also catch KeyboardInterrupt, etc.

Ah, yes. Will do a patch for this.

> jelmer merged commit ff831aa into jelmer:master

Thanks!

PR #283, opened 2015-05-05
Support git.bat on Windows

userzimmermann:
I have only __git.bat__ in my Windows PATH, which made dulwich always complain about the missing __git__ executable :) I changed the `Popen` call to prepend `['cmd', '/c']` on Windows. Now __git.bat__ works as well as __git.exe__

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -663,7 +663,14 @@ class SubprocessGitClient(TraditionalGitClient):
 
     git = ['git']
     if sys.platform == 'win32': # support .exe, .bat and .cmd
-        git = ['cmd', '/c'] + git
+        try: # to avoid overhead
+            import win32api
+        except ImportError: # run through cmd.exe with some overhead
+            git = ['cmd', '/c'] + git
+        else:
+            status, git = win32api.FindExecutable('git')
+            #TODO: need to check status? (exc is raised if not found)
+            git = [git]
 
     def _connect(self, service, path):
         import subprocess
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -46,8 +46,6 @@ import socket
 import subprocess
 import sys
 
-WIN = sys.platform.startswith('win')
-
 try:
     import urllib2
     import urlparse
@@ -663,11 +661,13 @@ class SubprocessGitClient(TraditionalGitClient):
             del kwargs['stderr']
         TraditionalGitClient.__init__(self, *args, **kwargs)
 
+    git = ['git']
+    if sys.platform == 'win32': # support .exe, .bat and .cmd
+        git = ['cmd', '/c'] + git
+
     def _connect(self, service, path):
         import subprocess
-        argv = ['git', service, path]
-        if WIN: # support git.exe and git.bat
-            argv = ['cmd', '/c'] + argv
+        argv = self.git + [service, path]
         p = SubprocessWrapper(
             subprocess.Popen(argv, bufsize=0, stdin=subprocess.PIPE,
                              stdout=subprocess.PIPE,
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -46,6 +46,8 @@ import socket
 import subprocess
 import sys
 
+WIN = sys.platform.startswith('win')
+
 try:
     import urllib2
     import urlparse
@@ -664,6 +666,8 @@ class SubprocessGitClient(TraditionalGitClient):
     def _connect(self, service, path):
         import subprocess
         argv = ['git', service, path]
+        if WIN: # support git.exe and git.bat
+            argv = ['cmd', '/c'] + argv
         p = SubprocessWrapper(
             subprocess.Popen(argv, bufsize=0, stdin=subprocess.PIPE,
                              stdout=subprocess.PIPE,


dulwich/client.py
```diff
@@ -46,6 +46,8 @@
 import subprocess
 import sys
 
+WIN = sys.platform.startswith('win')
```
jelmer: please move this code to below, where it is actually used and please check for just 'win32' as that seems to be the only relevant value.

dulwich/client.py
```diff
@@ -664,6 +666,8 @@ def __init__(self, *args, **kwargs):
     def _connect(self, service, path):
         import subprocess
         argv = ['git', service, path]
+        if WIN: # support git.exe and git.bat
+            argv = ['cmd', '/c'] + argv
```
jelmer: Is there any performance overhead involved with running this through cmd?

jelmer:
Thanks for the patch! Some comments inline.

garyvdm:
Would you consider the option of adding the location of git.exe into the PATH variable? The git windows installer gives you the option to do this: 
screenshot from 2015-05-06 08-50-30

userzimmermann:
I moved all the Windows logic to `SubprocessGitClient` and separated it a bit from the `_connect` method.

Regarding `sys.platform`: I always use `.startswith('win')`, thinking about possibly different values which might come with future Python and Windows versions. But that doesn't matter currently. I just changed it :)

There is a performance overhead:

    %timeit call(['git.exe'])
    10 loops, best of 3: 60.1 ms per loop

    %timeit call(['git.bat'])
    10 loops, best of 3: 80.4 ms per loop

    %timeit call(['cmd', '/c', 'git.exe'])
    10 loops, best of 3: 75.2 ms per loop

    %timeit call(['cmd', '/c', 'git.bat'])
    10 loops, best of 3: 79 ms per loop

But the extra time is fix and so the actual relative impact becomes really low on running real __git__ subcommands.

@garyvdm I use the embedded portable __git__ from SourceTree, which has a __git.bat__ wrapper instead of the small __git.exe__ wrapper of the standard Windows __git__ distribution. I also added some custom logic to __git.bat__. And there are often reasons to wrap binaries with custom scripts. On Windows that usually ends up in some __.bat__ or __.cmd__, which are unfortunately not implicitly supported by `subprocess`. So I think it's generally a good idea to call external commands via `cmd /c` on Windows.

jelmer:
Hmmm, 25% is quite a lot - especially since the only point of SubprocessGitClient is speed. Perhaps this should rather be something we let the user override? 

Jelmer



On 6 May 2015 5:31:16 pm BST, Stefan Zimmermann <notifications@github.com> wrote:
>I moved all the Windows logic to `SubprocessGitClient` and separated it
>a bit from the `_connect` method.
>
>Regarding `sys.platform`: I always use `.startswith('win')`, thinking
>about possibly different values which might come with future Python and
>Windows versions. But that doesn't matter currently. I just changed it
>:)
>
>There is a performance overhead:
>
>    %timeit call(['git.exe'])
>    10 loops, best of 3: 60.1 ms per loop
>
>    %timeit call(['git.bat'])
>    10 loops, best of 3: 80.4 ms per loop
>
>    %timeit call(['cmd', '/c', 'git.exe'])
>    10 loops, best of 3: 75.2 ms per loop
>
>    %timeit call(['cmd', '/c', 'git.bat'])
>    10 loops, best of 3: 79 ms per loop
>
>But the extra time is fix and so the actual relative impact becomes
>really low on running real __git__ subcommands.
>
>@garyvdm I use the embedded portable __git__ from SourceTree, which has
>a __git.bat__ wrapper instead of the small __git.exe__ wrapper of the
>standard Windows __git__ distribution. I also added some custom logic
>to __git.bat__. And there are often reasons to wrap binaries with
>custom scripts. On Windows that usually ends up in some __.bat__ or
>__.cmd__, which are unfortunately not implicitly supported by
>`subprocess`. So I think it's generally a good idea to call external
>commands via `cmd /c` on Windows.
>
>---
>https://github.com/jelmer/dulwich/pull/283#issuecomment-99530482

Jelmer Vernooij <jelmer@jelmer.co.uk> - https://jelmer.co.uk/

userzimmermann:
I found something very nice! :)
Would __pywin32__ as a Windows dependency for __dulwich__ be ok?
Most people should have it installed by default.

    win32api.FindExecutable('git')

Works for all executable file extensions.
And the performance:

    %timeit call(['git.exe'])
    10 loops, best of 3: 58.4 ms per loop

    %timeit call([win32api.FindExecutable('git.exe')[1]])
    10 loops, best of 3: 57.1 ms per loop

Virtually the same...

jelmer:
Perhaps it could optionally call out to that, if pywin32 is available? I'd rather avoid another dependency.

userzimmermann:
I've added the speedup via `try: import win32api ... except: ...`
I hope the whole solution is now an acceptable compromise, with some __pywin32__ info in the __README__...

And yes, if no __pywin32__ is installed then __25%__ overhead for the __.exe__ looks quite much, but that's just the worst case for the dummy __git__ call without a real action. The longer a real __git__ command takes the lower the percentage gets...
And __pywin32__ is kind of a standard package on Windows...

userzimmermann:
Thanks for accepting my stuff :)

BTW: I opened a topic on comp.lang.python discussing a possible PEP for adding implicit .bat and .cmd support to Popen:

https://groups.google.com/forum/#!topic/comp.lang.python/RL3PmyUMsLk

PR #284, opened 2015-05-07
Revert byte paths

garyvdm:
I now think PR #271 was a mistake.

When I started looking a the test results on Windows Python 3 (https://ci.appveyor.com/project/garyvdm/dulwich/build/1.0.104/job/5j9b65x7phrex8tn) , I picked up an issues caused by that change:

As discussed in that PR, `open(bytes)` works fine, but if you do `os.stat(bytes)` or a bunch of other methods in `os`, you get:

    DeprecationWarning: The Windows bytes API has been deprecated, use Unicode filenames instead

That's bad. Don't want to be using deprecated features.

This reverts that change.  So now for paths must be `basestring` on python 2 or `str` but not `bytes` on python3.  The tests in `test_repository` will test with a unicode path so that we don't regress on #263. 

This also has 2 minor fixes for windows related to path handling.

I think we still need to relook at our handling of git tree paths to and from file system path in `index.build_index_from_tree`, `index.get_unstaged_changes`, and `Repo.stage`. I'm working on this.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -683,6 +683,7 @@ class SubprocessGitClient(TraditionalGitClient):
         if self.git_command is None:
             git_command = find_git_command()
         argv = git_command + [service, path]
+        argv = ['git', service.decode('ascii'), path]
         p = SubprocessWrapper(
             subprocess.Popen(argv, bufsize=0, stdin=subprocess.PIPE,
                              stdout=subprocess.PIPE,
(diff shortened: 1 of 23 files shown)


jelmer:
oddly enough, this appears to break one of the pypy tests?

jelmer:
This matches what I thought I knew about bytestrings for paths on Windows. Thanks for sending this in before it ended up in a release!

garyvdm:
The pypy error seems like an existing bug, that is only getting picked now because I've changed the test_repository tests to use a path with non ascii unicode characters. 

I've pushed a fix.

Issue #203, opened 2014-06-21
Unicode file name do not checkout correctly on windows


garyvdm:
Steps to reproduce:

```
dulwich clone https://github.com/garyvdm/git_unicode_files.git
dir git_unicode_files
```

expected: 1 file named `À`   (which is `u'\u00c0'`)
actual: the file is named `Ã€`  (which is `u'\u00c3\u20ac'`)

the file name is what you get if you do `u'\u00c0'.encode('utf8').decode('mbcs')`. mbcs it the default filesystem charter encoding used on windows.

The git client handles this correctly. I'll take a look at their source code in the future to try figure out how they handle this.

garyvdm:
This is what msysgit does: https://github.com/msysgit/git/commit/19d1e75d58d772329372d453ead964c813bbc6b6

jelmer:
Has this been resolved?

jelmer:
@garyvdm Has this been resolved?

garyvdm:
No, not yet.

guyskk:
I also encountered this problem.
see https://github.com/FriendCode/gittle/issues/72

UnicodeDecodeError When filename is"article/python2编码问题.md" or has unicode char

```
dulwich/index.py(423) build_index_from_tree()
-> full_path = os.path.join(prefix, entry.path)
(Pdb) pp prefix
u'E:/work/py/kkblog/article_repo/\u54c8\u54c8\\guyskk\\webhooks_test'
(Pdb) pp entry.path
'article/python2\xe7\xbc\x96\xe7\xa0\x81\xe9\x97\xae\xe9\xa2\x98.md'
(Pdb) os.path.join(prefix, entry.path)
*** UnicodeDecodeError: 'ascii' codec can't decode byte 0xe7 in position 16: ordinal not in range(128)
```

My script:

```python
# coding:utf-8

def pull_or_clone(dest, repo_url):

    from giturlparse import parse
    from gittle import Gittle
    import os
    p = parse(repo_url)
    user_repo_path = os.path.join(dest, p.owner, p.repo)
    if os.path.exists(user_repo_path):
        repo = Gittle(user_repo_path, origin_uri=repo_url)
        repo.pull()
    else:
        repo = Gittle.clone(repo_url, user_repo_path)

if __name__ == '__main__':
    dest = u"E:/work/py/kkblog/article_repo/哈哈"
    repo_url = u"https://github.com/guyskk/webhooks_test.git"
[...]

PR #285, opened 2015-05-07
Improve handling of unicode checkout paths

garyvdm:
We currently have 2 issues with the way we handle unicode checkout paths:

* Issue #203. On windows, the file system api is unicode, and so we can't just store the bytes of the path and not care about the encoding, as we do on unix. 

* We use to just store the bytes from the tree path as the filesystem path. But we no longer do this (due to a change from the python3 porting.)  This means we used to be able to checkout the following repo with delwich 0.10.1, but can no longer do so:  `git@github.com:garyvdm/git_unicode_files2.git`

So this PR fixes this by making methods that deal with the transition between checkout files and tree entries work in the following way:

* By default, on unix systems, the path in bytes form is simply copied, (Back to how it was before.)
* By default on windows system, the tree path encoding is assumed to be utf-8, and the decoded unicode path is used to write to the filesystem, and an encoded path is used to write to a git tree.
* On all systems, the encoding of tree paths can be specified by a new `tree_encoding` argument. This could be use to translate a tree with path encoding different to that of the encoding of the filesystem.

Note that this PR includes the commits from PR #284.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -25,6 +25,7 @@ import platform
 import stat
 import struct
 import sys
+import locale
 
 from dulwich.file import GitFile
 from dulwich.objects import (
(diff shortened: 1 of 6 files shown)


dulwich/index.py
```diff
@@ -463,30 +464,38 @@ def validate_path(path, element_validator=validate_path_element_default):
...
-        in a working dir. Suiteable only for fresh clones.
+        in a working dir. Suitable only for fresh clones.
     """
 
     index = Index(index_path)
 
+    if tree_encoding is None and sys.platform == 'win32':
+        tree_encoding = 'utf-8'
+    if not tree_encoding:
+        root_path = root_path.encode(sys.getfilesystemencoding())
```
jelmer: root_path should already have the filesystem encoding, passed in by the user.

garyvdm: I feel that it should be a `string` on python3, and a `basestring` on python2, like with `Repo(path)` and others.

Please note that there are some cases, where later in the code, `root_path` needs to be `str`/`basestring`, and *not* bytes.

dulwich/index.py
```diff
@@ -463,30 +464,38 @@ def validate_path(path, element_validator=validate_path_element_default):
...
+
     for entry in object_store.iter_tree_contents(tree_id):
-        if not validate_path(entry.path):
+        if not validate_path(entry.path, validate_path_element):
             continue
-        full_path = os.path.join(prefix, entry.path.decode(sys.getfilesystemencoding()))
+        fs_path = tree_to_fs_path(entry.path, tree_encoding)
+        full_path = os.path.join(root_path, fs_path)
```
garyvdm: `root_path` needs to be `str`/`basestring`, and *not* bytes here when `tree_encoding` is not none, either specified, or on windows, because then fs_path will also be `str`/`basestring`, and `os.path.join` requires the arguments to be the same type.

dulwich/index.py
```diff
@@ -502,37 +511,88 @@ def build_index_from_tree(prefix, index_path, object_store, tree_id,
...
     return blob
 
 
-def get_unstaged_changes(index, path):
+def get_unstaged_changes(index, root_path, tree_encoding=None):
     """Walk through an index and check for differences against working tree.
 
     :param index: index to check
-    :param path: path in which to find files
+    :param prefix: path in which to find files
```
garyvdm: Need to update doc here.

jelmer:
Hi Gary,

Thanks again for working on this. It's a tricky area...

On Thu, May 07, 2015 at 02:00:38PM -0700, Gary van der Merwe wrote:
> We currently have 2 issues with the way we handle unicode checkout paths:
> 
> * Issue #203. On windows, the file system api is unicode, and so we can't just store the bytes of the path and not care about the encoding, as we do on unix. 
> 
> * We use to just store the bytes from the tree path as the filesystem path. But we no longer do this (due to a change from the python3 porting.)  This means we used to be able to checkout the following repo with delwich 0.10.1, but can no longer do so.
> 
> So this PR fixes this by making methods that deal with the transition between checkout files and tree entries work in the following way:
Please split this PR up into multiple different ones. As it is
currently, it's hard to review.

> * By default, on unix systems, the path in bytes form is simply copied, (Back to how it was before.)
+1

> * By default on windows system, the tree path encoding is assumed to be utf-8, and the decoded unicode path is used to write to the filesystem, and an encoded path is used to write to a git tree.
I'd really rather not hardcode a specific encoding anywhere. If
encoding needs to happen, the user should have to specify an encoding.

Perhaps fix the behaviour for bytes-oriented filesystem APIs first,
then unicode oriented ones as a separate PR?

Cheers,

Jelmer

garyvdm:
Superseded by #290

PR #290, opened 2015-05-11
Improve handling of checkout paths

garyvdm:
This supersedes #285.

This is split into 2 main commits  + 1 test adding commit to hopefully make it easier to review. It also excludes changes that were made in #285, that I will resubmit later. 

This includes the following changes (copied from commit messages):

* Add `index.fs_to_tree_path` and `index.tree_to_fs_path` to translate between git
  tree paths, and checkout file system paths. These methods just translate the
  path separators for now. Use these methods where needed.
* Rename path arguments and variables to make it clear if they refer to a git
  tree paths, or checkout file system paths, or a checkout root.
* Add test to ensure we don't fiddle with checkout path encoding.
* `index.build_index_from_tree` will not decode tree paths, but will write to the
  file system using the bytes paths.
* `Repo.stage` will accept either a bytes or string path. Only if the path is a
  string with the path be encoded with `sys.getfilesystemencoding()`, otherwise
  the bytes will be used directly to write to the git tree.

This excludes:

* A fix for #203.
* A way to translate tree encodings.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -482,11 +482,13 @@ def build_index_from_tree(root_path, index_path, object_store, tree_id,
     """
 
     index = Index(index_path)
+    if not isinstance(root_path, bytes):
+        root_path = root_path.encode(sys.getfilesystemencoding())
 
     for entry in object_store.iter_tree_contents(tree_id):
         if not validate_path(entry.path, validate_path_element):
             continue
-        fs_path = tree_to_fs_path(entry.path.decode(sys.getfilesystemencoding()))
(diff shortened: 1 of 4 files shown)


garyvdm:
Superseded by #290

garyvdm:
Sorry, please hold off will review this. There is quite a bit of windows breakage with this still (I got mixed up with the wrong tests results.)

garyvdm:
Ok, windows issue has  been resolved. Ready for review.

Windows build result: https://ci.appveyor.com/project/garyvdm/dulwich/build/1.0.156

PR #292, opened 2015-05-12
Symlinks stuff

garyvdm:
I recently discovered that `os.symlink` & `os.readlink` are available on windows as of python 3.2.

So where we have tests that require symlink support, we should check if symlinks are available with ` hasattr(os, 'symlink')` rather than checking `sys.platform` or `os.name`. This enables running these tests on Windows with python >= 3.2.

This introduces a new test failure on Windows python 3, but this is fixed by #290. (https://ci.appveyor.com/project/garyvdm/dulwich/build/1.0.164/job/2u7ra0kyj3onass1#L1131)

This also fixes a test windows failure due to a difference `stat_val.st_size` between windows and unix (on unix 1 byte, on windows 0 bytes.)

This also splits up `BuildIndexTests.test_nonempty` so that the symlink specific code is in a separate test, so that the majority of `BuildIndexTests.test_nonempty` is also run on windows python2

### DIFF
--- a/dulwich/tests/test_index.py
+++ b/dulwich/tests/test_index.py
@@ -50,7 +50,10 @@ from dulwich.objects import (
     Tree,
     )
 from dulwich.repo import Repo
-from dulwich.tests import TestCase
+from dulwich.tests import (
+    TestCase,
+    skipIf,
+)
 
 class IndexTestCase(TestCase):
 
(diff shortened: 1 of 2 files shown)


jelmer:
Please use hasattr rather than getattr, as the latter swallows exceptions.

garyvdm:
Aghh, my rebasing made some comments disappeared. Going to reiterate so that it 's documented for those without notification emails.

Regarding the size for symlinks: msysgit sets the size to 0, so I've change `BuildIndexTests.test_symlink' to check for 0 on windows, and 1 elsewhere.

Switched to using `getattr` rather than `hasattr` for checking for symlink availability.

garyvdm:
> @jelmer merged commit 181426e into jelmer:master 16 hours ago

Thanks!

PR #293, opened 2015-05-14
Checkout paths2

garyvdm:
Another alternative to #290.

I realised I can reorder the commits, so that there are less changes between the commits, which I hope will make it easier to review.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -482,12 +482,13 @@ def build_index_from_tree(prefix, index_path, object_store, tree_id,
     """
 
     index = Index(index_path)
+    if not isinstance(prefix, bytes):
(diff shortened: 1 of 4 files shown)


garyvdm:
I've pushed a rebased version, which eat the comments made on commit. To see those comments, go here: https://github.com/garyvdm/dulwich/commit/b36a25308eff913fb415040301086f3ec327dca8


> jelmer commented on dulwich/index.py in b36a253 5 hours ago
> These two functions have the exact same docstring.. This one is wrong :)

Fixed

jelmer:
Merged. I'd still like to do some more iterations on these two functions, so I've made them private for the moment (to prevent anybody from relying on their current signature).

garyvdm:
Ok. Thanks

PR #300, opened 2015-05-25
Decode path when reading symlink on windows.

garyvdm:
This fixes a new failure on windows: https://ci.appveyor.com/project/garyvdm/dulwich/build/1.0.187/job/7mj30hg52ylb2x71#L1193

I though I handled this previously, but missed it some how.

`os.readlink` has differing behaviours on windows vs unix. For the path argument, and the return value, it must be/is 

| OS | cpython | pypy |
| --- | --- | --- |
| unix | either. Return value is same as argument | bytes |
| windows | unicode string | Not sure. Have not tested. |

So on windows, We have to decode the path argument, and encode the return value.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -516,7 +516,12 @@ def blob_from_path_and_stat(fs_path, st):
         with open(fs_path, 'rb') as f:
             blob.data = f.read()
     else:
-        blob.data = os.readlink(fs_path)
+        if sys.platform == 'win32':
+            fs_encoding = sys.getfilesystemencoding()
+            blob.data = (os.readlink(fs_path.decode(fs_encoding))
+                         .encode(fs_encoding))
+        else:
+            blob.data = os.readlink(fs_path)
     return blob
 
 


dulwich/index.py
```diff
@@ -516,7 +516,12 @@ def blob_from_path_and_stat(fs_path, st):
         with open(fs_path, 'rb') as f:
             blob.data = f.read()
     else:
-        blob.data = os.readlink(fs_path)
+        if sys.platform == 'win32':
+            fs_encoding = sys.getfilesystemencoding()
+            blob.data = (os.readlink(fs_path.decode(fs_encoding))
+                         .encode(fs_encoding))
```
jelmer: This is using the file system encoding as the tree encoding?

garyvdm: Yes :-( . This is the same for other types of files, because methods such as `open` will implicitly decode the path when passed bytes, using `sys.getfilesystemencoding()`, which will be `'mbsc'`, as the encoding. This is the cause of #203. 

This patch  will make symlinks behave the same as other files, which means the will work for ascii only, or mbsc encoded paths, but  will break for any other encoding (as per #203)

I make sure this is fixed when I get round to fixing #203.

jelmer: On Mon, May 25, 2015 at 05:30:36AM -0700, Gary van der Merwe wrote:
> > @@ -516,7 +516,12 @@ def blob_from_path_and_stat(fs_path, st):
> >          with open(fs_path, 'rb') as f:
> >              blob.data = f.read()
> >      else:
> > -        blob.data = os.readlink(fs_path)
> > +        if sys.platform == 'win32':
> > +            fs_encoding = sys.getfilesystemencoding()
> > +            blob.data = (os.readlink(fs_path.decode(fs_encoding))
> > +                         .encode(fs_encoding))
> 
> Yes :-( . This is the same for other types of files, because methods such as `open` will implicitly decode the path when passed bytes, using `sys.getfilesystemencoding()`, which will be `'mbsc'`, as the encoding. This is the cause of #203. 
Can you require the caller pass in the tree_encoding and then pass in sys.getfilesystemencoding() ? At least that way the intent is clear.

Cheers,

Jelmer

garyvdm: Maybe I should just work at fixing #203. What are you time lines for the next release?

jelmer: On Mon, May 25, 2015 at 10:26:58AM -0700, Gary van der Merwe wrote:
> > @@ -516,7 +516,12 @@ def blob_from_path_and_stat(fs_path, st):
> >          with open(fs_path, 'rb') as f:
> >              blob.data = f.read()
> >      else:
> > -        blob.data = os.readlink(fs_path)
> > +        if sys.platform == 'win32':
> > +            fs_encoding = sys.getfilesystemencoding()
> > +            blob.data = (os.readlink(fs_path.decode(fs_encoding))
> > +                         .encode(fs_encoding))
> 
> Maybe I should just work at fixing #203. What are you time lines for the next release?

I'd like to do a new release as soon as we're comfortable enough putting the
Python3 support out there - so probably after the porcelain has been fixed,
and possibly dulwich.web.

Cheers,

Jelmer

jelmer:
ping?

jelmer:
This has been open for >1y, so I'm going to close it for now. Feel free to reopen if you come back to it.

PR #311, opened 2015-06-01
Consent to GPLv2/Apachev2 relicense on behalf of Googlers

michaeledgar:
Google opensource licensing has approved the relicensing of all Googlers' contributions to dulwich.

PR #312, opened 2015-06-01
Consent to apache2 relicensing.

durin42:
Done as a manual step for me - my Google work is approved (see pull request #311), and if I have any work that predates my time at Google (I think I do?) I approve for that too.

Thanks!


jelmer:
Thanks, Augie!

PR #313, opened 2015-06-02
fix off-by-one depth bug in serving shallow clones

michaeledgar:


michaeledgar:
Hmm, I was nervous about Windows, but the failures on windows seem unrelated. The appveyor history looks like that build is consistently failing a different test. I'm happy to fix anything up while I'm here, though!

jelmer:
Thanks!

PR #315, opened 2015-06-02
"Safe" removing of temp/lock files to prevent "access denied"

sebres:
"Safe" removing of temp/lock files to prevent "access denied" affected through rude virus scanner etc.

Closes related hg-git issue https://github.com/schacon/hg-git/issues/297


jelmer:
On Tue, Jun 02, 2015 at 11:30:23AM -0700, Serg G. Brester wrote:
> "Safe" removing of temp/lock files to prevent "access denied" affected through rude virus scanner etc.
> 
> Closes related hg-git issue https://github.com/schacon/hg-git/issues/297

Why is this specifically an issue with dulwich, and not with every single
Python application that calls out to os.remove() when such virus scanners are present?

Jelmer

sebres:
<blockquote>Why is this specifically an issue with dulwich</blockquote>
Very interesting question, shortly - I don't know, but counter question "Why you need a function `fancy_rename` at all (specifically with dulwich)?"

durin42:
ISTR we have some weird shenanigans to try and work around virus scanners
in hg, so this doesn't entirely surprise me. See
https://selenic.com/hg/file/724d7982b790/mercurial/win32.py#l439 for some
details.

jelmer:
On Tue, Jun 02, 2015 at 11:42:37AM -0700, Serg G. Brester wrote:
> <blockquote>Why is this specifically an issue with dulwich</blockquote>
> Very interesting question, shortly - I don't know, but counter question "Why you need a function `fancy_rename` at all (specifically with dulwich)?"

See https://github.com/jelmer/dulwich/commit/c402eb265eabbd3ed8c7e0f56bdbaeef4a15c9fc

fancy_rename was introduced for Windows as well. These kinds of hacks shouldn't
be used on platforms that don't need them.; especially a change like this one,
which has the potential of making calls to Dulwich very slow for trivial operations
(if permissions are inappropriate, IIUC the first write access may take up to 5 seconds to timeout).

Having said that, I'll update dulwich to only use fancy_rename on Windows.

With regards to this particular change, is there any way we can avoid the
timeout? The code in hg that Augie linked to seems more reasonable.

Jelmer

sebres:
<blockquote>(if permissions are inappropriate, IIUC the first write access may take up to 5 seconds to timeout).</blockquote>
Missing permissions for file, that was created just before? Are you sure?
<blockquote>Having said that, I'll update dulwich to only use fancy_rename on Windows</blockquote> No problem, but I can imagine similar situation with linux also, for instant deleting immediately after file creation (ex. something like fulltext indexing, etc.). 
So why not keep it, If doesn't really disturb.

jelmer:
Something like a virus scanner could be mucking about with these files, so I wouldn't be surprised if it e.g. restricted access while it checks the file.

I'm not sure why we would need fancy_rename() on linux, as renames are atomic there.

jelmer:
Closing this stale pull request. Feel free to reopen with the raised issues addressed if you're still interested in this.

PR #316, opened 2015-06-03
fix encoding name check in test_refs when LANG=C

michaeledgar:
The value of sys.getfilesystemencoding() yields 'ANSI_X3.4-1968' and not
'ascii' (at least on linux systems).

(I'm not sure if this is true for all UNIX)


jelmer:
On Wed, Jun 03, 2015 at 11:33:24AM -0700, Michael Edgar wrote:
> The value of sys.getfilesystemencoding() yields 'ANSI_X3.4-1968' and not
> 'ascii' (at least on linux systems).
> 
> (I'm not sure if this is true for all UNIX)

It differs per system :) E.g. on Debian build slaves it is explicitly
set to 'ascii'.

Thanks for the pull request - I've instead just made it attempt a
conversion and skip if that fails.

michaeledgar:
That approach looks a lot better, dropping this :-P

PR #333, opened 2015-06-15
remove myself from relicensing-apachev2.txt

danc86:
I consent to the proposed re-licensing. (My contribution to Dulwich is
not copyrightable anyway.)


jelmer:
Thanks!

PR #345, opened 2015-06-19
Remove deep recursion from PackData

wgrant:
PackData's resolve_object and _follow_chain methods recurse down delta chains, getting very deep and potentially exceeding the interpreter's recursion limit on heavily packed repos. I've rewritten them to iterate instead of recurse.

The only functional change is that _follow_chain is now breadth-first rather than depth-first.


jelmer:
Thanks, merged.

PR #348, opened 2015-06-24
remove Siddharth Agarwal from relicensing text file

sid0:
This code is copyright my employer, Facebook, which already licensed my
contribution as GPLv2 or later + Apache v2 at the time it was submitted. See
https://github.com/jelmer/dulwich/pull/159.


jelmer:
merged, thanks.

PR #352, opened 2015-07-17
Approve my contributions being relicensed

rockstar:
Removing my name from the list of people needed to be tracked down in order to relicense dulwich.


jelmer:
Thank you!

PR #361, opened 2015-07-30
Windows git.bat support broken

userzimmermann:
The support for running **git.bat** (and **git.cmd**) on Windows from #283 is broken, because in `SubprocessGitClient._connect()` the `argv` list for `Popen` gets overwritten with a fixed `'git'` string as first item.


jelmer:
Thanks

PR #377, opened 2015-09-13
Use custom SSH command quoting logic

inducer:


inducer:
You'll notice that this puts quotes around everything except the command, all the time. At least Bitbucket seems to expect this, and they will send you away with 

```
conq: invalid repository syntax.
```

(whatever that means) if you don't have them.

inducer:
Regarding the quoting not being secure: Agreed. Would replacing `\` by `\\` make you happier?

> Is it not possible to just decode the bytes and invoke subprocess.list2cmdline() before passing them to paramiko?

No because:

* We'd have to guess the encoding.
* `list2cmdline` will only put quotes around arguments that have spaces in them, whereas Bitbucket appears to want quotes all the time--because apparently that's what git does, and who would accept anything else? [/sarcasm]

jelmer:
Shell escaping is hard to get right; I don't really want to reimplement shlex.py inside of dulwich. On the other hand, I don't think a perfect solution should hold a fix for this bug hostage. Perhaps an alternative would be to raise an exception if one of the arguments contains a backslash? 

I regret merging in the paramiko code without tests; it's been a source of trouble for people. Dulwich tries to support quite a few platforms (both OSes and Python versions), and making sure those all keep working with paramiko while code changes is hard. In addition, in my experience paramiko doesn't have a particularly stable API; personally I've come across at least two silent changes in their contract (https://github.com/paramiko/paramiko/issues/343, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=794146). I think what I'll do for 0.11.2 or 0.12 is to move the paramiko SSH implementation to dulwich/contrib, with a deprecation warning for those attempting to import it from the old location in dulwich/client. That way its status is clear; we can move it back once there are appropriate tests.

What do you think?

jelmer:
(obviously that should be an assertion for either a backslash or a quote)

inducer:
I understand how paramiko seems like an can of worms from your perspective, especially given today's volume of bug traffic. (Sorry!)

I'd like to politely request that you don't hold a gun up to that code's head and say "or else". :) If you point me towards the existing tests for the `SubprocessSSHVendor`, I'd be happy to write equivalent tests for paramiko.

inducer:
shlex.quote is not long:

```
_find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search

def quote(s):
    """Return a shell-escaped version of the string *s*."""
    if not s:
        return "''"
    if _find_unsafe(s) is None:
        return s 

    # use single quotes, and put single quotes into double quotes
    # the string $'b is then quoted as '$'"'"'b'
    return "'" + s.replace("'", "'\"'\"'") + "'"
```

How about we copy this (adapted for byte strings)?

jelmer:
I've moved the ParamikoSSHVendor to dulwich.contrib.paramiko_vendor. I won't remove it, but I do want to make people think twice before using it; I think this communicates that bettter. We can always move it back to the main package if its quality improves. There is a compatibility wrapper around that warns about importing from the new location.

As an added bonus, this means that we don't attempt to import paramiko every time dulwich.client is imported.

jelmer:
I'm not happy about copying that code, but I guess it's not too bad. Send me a pull request? Please fetch before changing, as the code has moved.

inducer:
There you go. :) Thanks for your patience (and help!) today.

PR #388, opened 2015-11-01
Fix local cloning under Python 3

myint:
This gets local cloning working. It supports Python 3 while leaving the Python 2 effectively untouched.

```
$ python examples/clone.py . /tmp/clone
```

The above used to result in a `TypeError`. This did not show up in the tests since the tests don't use the default arguments.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -48,7 +48,6 @@ Differences in behaviour are considered bugs.
 
 __docformat__ = 'restructuredText'
 
-import codecs
 from collections import namedtuple
 from contextlib import (
     closing,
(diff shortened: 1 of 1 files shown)


myint:
Note that the Travis CI errors are unrelated to this pull request.

dulwich/porcelain.py
```diff
@@ -97,6 +98,14 @@
 GitStatus = namedtuple('GitStatus', 'staged unstaged untracked')
 
 
+if sys.version_info[0] < 3:
+    STDOUT = sys.stdout
```
jelmer: Please use a more descriptive name; these are not the same as stdout/stderr.

dulwich/porcelain.py
```diff
@@ -97,6 +98,14 @@
...
+    STDERR = sys.stderr
+else:
+    STDOUT = codecs.getwriter('utf-8')(sys.stdout.buffer)
```
jelmer: The system encoding is not necessarily utf-8. Is there any reason why just sys.stdout.buffer wouldn't work?

myint:
I've made the relevant changes.

jelmer:
Thanks, merged with two changes:

clarified variables names to not be capitalized (as they're not constants) and include the word 'bytes'.
I've only merged the change for clone() as that's the only one we've verified is correct. For some of the other functions the outstream is not actually a bytestring (e.g. print_commit). This needs adjusting/renaming.

PR #389, opened 2015-11-04
subprocess's mswindows attribute has gone private in 3.5

stefanor:
See: https://bugs.python.org/issue10838
Elsewhere, we look at `sys.platform` directly, so do it here too.

As seen in https://launchpad.net/ubuntu/+source/dulwich/0.11.2-1ubuntu2


jelmer:
Thanks! Merged.

PR #391, opened 2015-11-30
Enhanced parsing of partial/ambiguous refs

crbunney:
Whilst looking into wildcards in refspecs I noticed that https://git-scm.com/docs/gitrevisions describes an order of precedence for matching ambiguous references, so I updated `parse_ref` to match.

The precedence is described as follows:

> 1. If $GIT_DIR/<refname> exists, that is what you mean;
> 1. otherwise, refs/<refname> if it exists;
> 1. otherwise, refs/tags/<refname> if it exists;
> 1. otherwise, refs/heads/<refname> if it exists;
> 1. otherwise, refs/remotes/<refname> if it exists;
> 1. otherwise, refs/remotes/<refname>/HEAD if it exists.

### DIFF
--- a/dulwich/objectspec.py
+++ b/dulwich/objectspec.py
@@ -46,7 +46,15 @@ def parse_ref(container, refspec):
     :raise KeyError: If the ref can not be found
     """
     refspec = to_bytes(refspec)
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks, merged!

Issue #249, opened 2015-01-22
Clone from private repository


chamilad:
I couldn't find any resources on how to clone from a private repository by providing the username and the password. The following code gives an error. It would be a great help if you can add a section on the tutorial on how to do this. 

```
repo = Repo.init("local_repo_path", mkdir=True)
client, remote_path = get_transport_and_path("repo_url", auth=True, username="username", password="password")
```

Error

```
Traceback (most recent call last):
  File "dulwichtest.py", line 54, in <module>
    auth_clone_from_repo("https://chamilad@bitbucket.org/chamilad/testpr.git", "authlocal", "chamilad", "XegKhwMGMlmX")
  File "dulwichtest.py", line 45, in auth_clone_from_repo
    client, remote_path = get_transport_and_path(repo_url, auth=True, username=username, password=password)
  File "/home/chamilad/virtualenv/cartridgeagent/local/lib/python2.7/site-packages/dulwich/client.py", line 1110, in get_transport_and_path
    return get_transport_and_path_from_url(location, **kwargs)
  File "/home/chamilad/virtualenv/cartridgeagent/local/lib/python2.7/site-packages/dulwich/client.py", line 1091, in get_transport_and_path_from_url
    **kwargs), parsed.path
  File "/home/chamilad/virtualenv/cartridgeagent/local/lib/python2.7/site-packages/dulwich/client.py", line 935, in __init__
    GitClient.__init__(self, *args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'username'
```

Thanks!

jelmer:
On Thu, Jan 22, 2015 at 09:53:37AM -0800, Chamila wrote:
> I couldn't find any resources on how to clone from a private repository by providing the username and the password. The following code gives an error. It would be a great help if you can add a section on the tutorial on how to do this. 

URLs allow specification of usernames and passwords, e.g.:

http://username:password@host/path/to/repo

Cheers,

Jelmer

chamilad:
Hi Jelmer,

Thanks for the prompt response. However specifying the credentials in the URL results in the following error. The `:` is parsed as the port specifier.

```
repo_url = "https://username:password@bitbucket.org/chamilad/testpr.git"
repo = Repo.init(local_repo_path, mkdir=True)
client, remote_path = get_transport_and_path(repo_url)
remote_refs = client.fetch(remote_path, repo)
```

```
Traceback (most recent call last):
  
[...]

PR #393, opened 2015-12-01
Allow use of a custom urllib2 opener for authentication purposes over…

jsbain:
… http.  This solves #249.  See code included in that thread for an example opener.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -61,6 +61,7 @@ from dulwich.archive import (
     tar_stream,
     )
 from dulwich.client import (
+    HttpGitClient,
     get_transport_and_path,
     )
 from dulwich.errors import (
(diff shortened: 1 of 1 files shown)


jelmer:
Thanks for the patch.

I have some reservations about this - I'd rather not add transport-specific parameters to every porcelain function. #249 was specifically about encoding credentials in the URL. Wouldn't it already be possible to set a global opener?

jsbain:
how would one set a global opener?  

Looking again, perhaps it makes more sense to handle the encoded uri in get_transport_and_path?

jelmer:
Yes, the encoded URL should be handled in get_transport_and_path(). Other transport-specific options should probably be passed either using a generic object that can be used for all transports or be set using global variables.

You would set a global opener e.g by monkey patching dulwich.client.default_urllib2_opener

PR #396, opened 2015-12-03
Add missing DECREF on local variable

lelit:
The local variable `errors_module` wasn't decref-ed.


jelmer:
Thanks, merged.

Note that this is a cosmetic change. It won't make an actual difference since we're holding on to a reference of PyExc_ApplyDeltaError, which would also keep a reference to errors_module open.

PR #404, opened 2016-01-05
Parse user:passwords for HTTPGitClient

jsbain:
This solves #249, allowing get_transport_and_path to use http basic auth user:pass style urls, or to directly specify user/pass in a HTTPGitClient.

This works in porcelain as well (push, etc).

One possible issue: when using porcelain to push, using user:pass urls, the password will be displayed in the success /failure messages.  probably not a major issue if using the default outstream and errstream, but if logging is implemented, passwords would be displayed in the clear.  The solution might be to use client.base_url for httpgitclients, rather than remote_location.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -949,13 +949,18 @@ def default_urllib2_opener(config):
 
 class HttpGitClient(GitClient):
 
-    def __init__(self, base_url, dumb=None, opener=None, config=None, **kwargs):
+    def __init__(self, base_url, dumb=None, opener=None, config=None, 
+                 user='',passwd='', **kwargs):
         self.base_url = base_url.rstrip("/") + "/"
         self.dumb = dumb
         if opener is None:
             self.opener = default_urllib2_opener(config)
         else:
             self.opener = opener
+        if user:
+            pass_man=urllib2.HTTPPasswordMgrWithDefaultRealm()
+            pass_man.add_password(None,base_url,user,passwd)
+            self.opener.add_handler(urllib2.HTTPBasicAuthHandler(pass_man))
         GitClient.__init__(self, **kwargs)
 
     def __repr__(self):
@@ -1111,6 +1116,7 @@ def get_transport_and_path_from_url(url, config=None, **kwargs):
     :return: Tuple with client instance and relative path.
     """
     parsed = urlparse.urlparse(url)
+    auth, host = urllib2.splituser(parsed.netloc)
     if parsed.scheme == 'git':
         return (TCPGitClient(parsed.hostname, port=parsed.port, **kwargs),
                 parsed.path)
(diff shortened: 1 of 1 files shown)


jelmer:
On Tue, Jan 05, 2016 at 03:59:08AM -0800, jsbain wrote:
> This solves #249, allowing get_transport_and_path to use http basic auth user:pass style urls, or to directly specify user/pass in a HTTPGitClient.
Thanks! Any chance you can add a test or two for this to dulwich/tests/test_client.py, just making sure that
the username and password in the URL end up in HTTPGitClient?

> This works in porcelain as well (push, etc).
> 
> One possible issue: when using porcelain to push, using user:pass urls, the password will be displayed in the success /failure messages.  probably not a major issue if using the default outstream and errstream, but if logging is implemented, passwords would be displayed in the clear.  The solution might be to use client.base_url for httpgitclients, rather than remote_location.

Adding GitClient.get_url() seems reasonable to me; we can then use that in error messages.

Jelmer

mjmaenpaa:
So what's the status of this pull request, because I really need this functionality.

If only tests are needed, then I can write them. Adding GitClient.get_url is propably little more complicated if we want it to return meaningful urls with all clients. However I can do basic version which can be updated later to include more meaningful urls.

jelmer:
I haven't heard from jsbain since January so any help getting this implemented would be great. I don't think we need a get_url() method just yet, but we probably want to make base_url private.

mjmaenpaa:
It seems that I forgot to make pull request even though i wrote tests two weeks ago.

I wrote few tests for this that make sure that username and password end up in HttpGitClient when using get_transport_and_path. Additionally they make sure that init actually ends up calling HTTPPasswordMgrWithDefaultRealm.add_password.

I can make new pull request with just my tests or new pull request that includes both my tests and jsbain's changes rebased against master.

jelmer:
On Wed, Jul 27, 2016 at 01:57:24PM -0700, mjmaenpaa wrote:
> It seems that I forgot to make pull request even though i wrote tests two weeks ago.
> 
> I wrote few tests for this that make sure that username and password end up in HttpGitClient when using get_transport_and_path. Additionally they make sure that init actually ends up calling HTTPPasswordMgrWithDefaultRealm.add_password.
> 
> I can make new pull request with just my tests or new pull request that includes both my tests and jsbain's changes rebased against master.
A pull request with both would be great. I've just merged your other
one.

jelmer:
Superseded by #445

jelmer:
On Wed, Jul 27, 2016 at 01:57:24PM -0700, mjmaenpaa wrote:
> It seems that I forgot to make pull request even though i wrote tests two weeks ago.
> 
> I wrote few tests for this that make sure that username and password end up in HttpGitClient when using get_transport_and_path. Additionally they make sure that init actually ends up calling HTTPPasswordMgrWithDefaultRealm.add_password.
> 
> I can make new pull request with just my tests or new pull request that includes both my tests and jsbain's changes rebased against master.

Thanks.

FWIW this is still on my todo list. I'm hoping to get to it sometime this week.

PR #409, opened 2016-02-07
Fix a couple of compiler warnings

jonashaag:
btw, I think you should start using unsigned types for counters etc (`size_t` instead of `int`)

### DIFF
--- a/dulwich/_pack.c
+++ b/dulwich/_pack.c
@@ -47,10 +47,10 @@ static int py_is_sha(PyObject *sha)
 }
 
 
-static size_t get_delta_header_size(uint8_t *delta, int *index, int length)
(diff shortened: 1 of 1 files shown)


dulwich/_pack.c
```diff
@@ -110,16 +110,16 @@ static PyObject *py_apply_delta(PyObject *self, PyObject *args)
 	}
 
 	src_buf = (uint8_t *)PyString_AS_STRING(py_src_buf);
-	src_buf_len = PyString_GET_SIZE(py_src_buf);
+	src_buf_len = (size_t)PyString_GET_SIZE(py_src_buf);
```
jonashaag: We can be sure here that `src_buf_len > -1`, thus the conversion is safe.

jelmer:
Merged, thanks.

PR #413, opened 2016-02-17
Support completely empty messages in git objects

olasd:
As it turns out, a git object can be terminated by the last header,
without empty lines. In that case, dulwich currently "eats" the last
header line, which is obviously wrong.

When such a case happens, return an object with a None message instead
of an empty line. We make sure that serialization/deserialization
still works as expected.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -602,17 +602,39 @@ def _parse_message(chunks):
     f = BytesIO(b''.join(chunks))
     k = None
     v = ""
+    eof = False
+
+    # Parse the headers
+    #
+    # Headers can contain newlines. The next line is indented with a space.
+    # We store the latest key as 'k', and the accumulated value as 'v'.
     for l in f:
         if l.startswith(b' '):
+            # Indented continuation of the previous line
             v += l[1:]
         else:
             if k is not None:
+                # We parsed a new header, return its value
                 yield (k, v.rstrip(b'\n'))
             if l == b'\n':
                 # Empty line indicates end of headers
                 break
             (k, v) = l.split(b' ', 1)
-    yield (None, f.read())
+
+    else:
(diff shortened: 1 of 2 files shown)


jelmer:
On Wed, Feb 17, 2016 at 03:16:31AM -0800, Nicolas Dandrimont wrote:
> As it turns out, a git object can be terminated by the last header,
> without empty lines. In that case, dulwich currently "eats" the last
> header line, which is obviously wrong.
> 
> When such a case happens, return an object with a None message instead
> of an empty line. We make sure that serialization/deserialization
> still works as expected.
How is such a commit created with C Git?

olasd:
> How is such a commit created with C Git?

Actually, I have noticed this issue with tags, not with commits (out of ~3.6 million unique tag objects found on github, this was found on a bit more than 8400, so it's not a *big* issue, but it happens). This looks like an artifact of a conversion from another VCS to git.

See `git help mktag`: when the message is empty, the empty line between the headers and the mesasge is not mandatory.

One of the repositories where such tags have been found is https://github.com/antoniofabio/iDMC (just git cat-file one of the tags and you'll see the tag ends with the last header). When one of those tags is parsed with dulwich, the tagger line is ignored.

jelmer:
C Git doesn't appear to support ending the headers without an empty line either:

% git cat-file -p rel-2.0.1 | git mktag
error: char119: trailing garbage in tag header
fatal: invalid tag signature file

There is something else weird about these tags:

% git cat-file -p rel-2.0.1
object 06cbc6b3fe3272f2b7f652c3e20551d8f74776df
type commit
tag rel-2.0.1
tagger antonio.fabio <antonio.fabio> 0 +0000

The tag time and timezone are both set to 0. 

Either way, as these tags are appearing in the wild, we should support this in Dulwich anyway.

PR #417, opened 2016-04-12
Don't expect 'wsgi.input' to have working 'seek' method

jonashaag:
This is a follow-up to 5a69c46 which misses the fact that even though an input stream might have 'seek' and 'tell' methods, they do not necessarily have to be implemented.  See also http://bugs.python.org/issue12877


jelmer:
Merged, thanks!

PR #418, opened 2016-04-12
Fix hang on Gzip decompression

jonashaag:
'LimitedInputFilter' should be applied before the Gzip middleware as 'CONTENT_LENGTH' refers to the size of the *compressed* body.  If applied in wrong order, this may cause 'copyfileobj()' to try to read more data from the socket than is available, resulting in a call to 'recv()' that hangs forever.


jonashaag:
Note: There don't seem to exist any tests for `make_wsgi_chain`, so I wasn't sure how to test this.

jelmer:
Merged, thanks!

PR #419, opened 2016-04-19
fastexport module : Invalid dataref 

dysosmus:
The `fastexport.GitFastExporter` generating invalid exports, like this one: 
```
$ > git fast-import << EOF
blob
mark :1
data 17
{"name": "Felix"}
commit refs/heads/master
mark :2
author Foo Bar <foo.bar@foo.bar> 1461062670 -0200
committer Foo Bar <foo.bar@foo.bar> 1461062670 -0200
data 14
Initial commit
M 644 1 test
EOF
$ > fatal: Invalid dataref: M 644 1 test
```
The line`M 644 1 test`  should be `M 644 :1 test` 

I fixed `fastexport.GitFastExporter._iter_files` method and the corresponding test.


jelmer:
Thanks!

PR #420, opened 2016-04-22
MemoryObjectStore loses objects

dysosmus:
With the current implementation of the `MemoryObjectStore` the store loses the object if you update the object, e.g `store[b'sha'].data = b'new data'`.

This pull request use a customized dict to make sure that don't happen.

``` python
from dulwich.repo import MemoryRepo, Blob, Repo
repository = MemoryRepo() 
blob = Blob.from_string(b'data')
blob_sha = blob.id 
repository.object_store.add_object(blob)
blob.data = b'' 
assert blob.id != blob_sha 
assert repository[blob_sha].id == blob_sha  # that assert fail, the initial blob she should be equal to the stored one.
```

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -685,6 +685,14 @@ class MemoryObjectStore(BaseObjectStore):
             This dict only accept bytestring of 40 as keys and ShaFile
             instances for values.
         """
+        def __init__(self, copy_only=True):
+            """
+                :param copy_only: (bool)
+                        If set at False the store will behave like a normal
+                        dictionary and will not return copies of the stored
+                        object. True by default.
+            """
+            self._copy_only = copy_only
 
         def __getitem__(self, key):
             if not len(key) == 40 or not isinstance(key, bytes):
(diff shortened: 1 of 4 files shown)


jelmer:
Thanks for pointing this out.

There is a question here about whether the caller should allowed to modify the object, rather than requiring that they copy it before modifying (so that read access doesn't involve a copy). I'm undecided on this, as allowing modification works well for the disk based object store.

Why do we need to store the objects serialized? Couldn't we just make a copy in __getitem__ and return that ?

dysosmus:
I don't have any good reasons to store the objects serialized, I just overlooked the `copy` method. I updated the pull request.


On the other point, I think the object store should never lose an object just by side effect.
That change doesn't bring any strong constraint, if you want to delete the previous version and only keep the current is still possible, it will only take one extra line. 
Besides it makes the object store more reliable by default which serves the global goal of a stable git implementation.

Also in the current master the disk store implementation doesn't lose objects even if you update an object, like in the code snippet of my first comment.

jelmer:
I've merged your test, but taken a different approach for the implementation. Adding a safe_store flag unnecessarily complicates the API IMO, and I think the implementation could be simpler. See https://github.com/jelmer/dulwich/commit/e099e140600140a92cc09cd9bb8d12aba0ca6429

Thanks for the report & the test!

PR #423, opened 2016-04-26
Incorrect id value after copying a Tree

dysosmus:
If a `Tree` is updated and copied the `id` isn't recomputed and the `_need_serializing` value is lost.
That can lead to weird bugs.
The pull-request update the `ShaFile.copy` method to be sure that the id is correct before copying.

```python

from dulwich.objects import Tree, Blob

shas = []
tree = Tree()
shas.append(tree.id)
_ = tree.add(b'data', 0o644, Blob().id)
copied = tree.copy()
shas.append(tree.id)
shas.append(copied.id)

assert shas[0] not in shas[1:] #  fail the initial sha is the same after and before the add operation
assert shas[1] == shas[2]
```

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -484,10 +484,15 @@ class ShaFile(object):
     def copy(self):
         """Create a new copy of this SHA1 object from its raw string"""
         obj_class = object_class(self.get_type())
+        # The id need to be retrieved before calling as_raw_string because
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the patch. I've merged this, but modified the fix - the issue is in ShaFile.id rather than ShaFile.copy.

PR #425, opened 2016-04-29
Update porcelain docs example to add and commit correctly

kkellyy:
Fixes the example `porcelain.add()` to successfully stage the file. Also makes example `porcelain.commit()` py3 compatible since the file is open for writing with bytes in `prepare_msg()`.

`make doc` succeeds and html output appears as expected.


jelmer:
Merged, thanks!

PR #431, opened 2016-05-31
File-like objects' encodings can be set, and None.

Julian:
Separately, it seems quite unfortunate that this function's method of communicating errors is by writing to a stream and swallowing the exception. Wouldn't it be useful to have high level git operations but still actually communicate errors in a way that can be handled programmatically?

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -585,7 +585,7 @@ def push(repo, remote_location, refspecs=None,
                     refs[rh] = r.refs[lh]
             return refs
 
-        err_encoding = getattr(errstream, 'encoding', 'utf-8')
+        err_encoding = getattr(errstream, 'encoding', None) or 'utf-8'
         remote_location_bytes = remote_location.encode(err_encoding)
         try:
             client.send_pack(path, update_refs,


jelmer:
Unfortunately this is a property of the way that Git works, and Dulwich has the same interface because it needs to talk to the CGit server interface.

A wrapper around this interface that is more pythonic would be nice, but is tricky to get right.

Julian:
I don't know much about CGit clearly but what I meant in case it was unclear was that ISTM that this is the API that should have the "nice" interface, and you can always build the current one on top of it to support whatever client is expecting a string on a file. Possibly I'm missing context though (I'm sure I am).

Regardless, is the actual patch here acceptable?

Thanks!

jelmer:
As mentioned in my previous comment, it would be nice to have a higher level interface in the form of a wrapper. However, that is tricky to get right. Patches welcome though.

This change itself looks reasonable to me. In which situations is the encoding present but None?

Julian:
The simplest one is if it's just a file.

`open("foo", "w").encoding == None`

jelmer:
Thanks, merged!

Julian:
Thank you!

PR #440, opened 2016-06-30
Add equality comparisons.

Julian:
Not sure this is the right implementation for subclasses but it seems like a start?

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -120,6 +120,14 @@ class BaseObjectStore(object):
         """Iterate over the SHAs that are present in this store."""
         raise NotImplementedError(self.__iter__)
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        return set(self) == set(other)
+
+    def __ne__(self, other):
+        return not self == other
+
     def add_object(self, obj):
         """Add a single object to this object store.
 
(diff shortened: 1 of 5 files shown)


jelmer:
The object store side of this looks reasonable to me. Can you submit that as a separate pull request?

Repositories are trickier - your change doesn't add a Repo.__eq__ method so no two Repo objects are going to be the same.

jelmer:
Sorry, I missed the new method on BaseRepo. Just comparing refs and objects is sufficient as far as the tests are concerned. 

For proper comparison you would probably also want to check e.g. the reflog and the configuration of the repository - possibly the working tree. This is not likely to be very useful I think, and it will be unusably slow on any repository that isn't trivial in size. What do you need repository comparison for, is there perhaps a more efficient alternative?

Julian:
Hrm. Understood. My use case is also for tests, I have my own objects with repo attributes and not being able to compare Repo objects means my own equality checks are awkward. I agree that comparing everything recursively inside the repository doesn't sound very useful.

jelmer:
On Sun, Jul 03, 2016 at 08:22:44AM -0700, Julian Berman wrote:
> Hrm. Understood. My use case is also for tests, I have my own objects with repo attributes and not being able to compare Repo objects means my own equality checks are awkward. I agree that comparing everything recursively inside the repository doesn't sound very useful.

Possibly we could just add a test helper for this purpose?

Jelmer

PR #445, opened 2016-07-28
Username and password support for HttpGitClient

mjmaenpaa:
Added tests for original pull request #404 by @jsbain and rebased jsbain's changes against master. Original pull request fixes issue #249. Additionally added support for only supplying username which allows to use github personal access tokens in url. This should fix issue #428.

I wrote few tests for this that make sure that username and password end up in HttpGitClient when using get_transport_and_path. Additionally tests make sure that init actually ends up calling HTTPPasswordMgrWithDefaultRealm.add_password.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ env:
 matrix:
   include:
     - python: "2.7"
-      env: TEST_REQUIRE="gevent greenlet geventhttpclient fastimport"
+      env: TEST_REQUIRE="gevent greenlet geventhttpclient fastimport mock"
     - python: "pypy"
-      env: TEST_REQUIRE="fastimport"
+      env: TEST_REQUIRE="fastimport mock"
     - python: "3.4"
       env: TEST_REQUIRE="gevent greenlet geventhttpclient fastimport"
     - python: "3.5"
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -998,16 +998,16 @@ def default_urllib2_opener(config):
 class HttpGitClient(GitClient):
 
     def __init__(self, base_url, dumb=None, opener=None, config=None,
-                 user='',passwd='', **kwargs):
+                 username=None, passwd=None, **kwargs):
         self._base_url = base_url.rstrip("/") + "/"
         self.dumb = dumb
         if opener is None:
             self.opener = default_urllib2_opener(config)
         else:
             self.opener = opener
-        if user:
+        if username is not None:
             pass_man=urllib2.HTTPPasswordMgrWithDefaultRealm()
-            pass_man.add_password(None,base_url,user,passwd)
+            pass_man.add_password(None, base_url, username, passwd)
             self.opener.add_handler(urllib2.HTTPBasicAuthHandler(pass_man))
         GitClient.__init__(self, **kwargs)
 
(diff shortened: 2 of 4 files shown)


gadams999:
Any update on when these pulls will be accepted? I have a use case where username:password is needed (AWS Lambda function to get a ref and create tgz from CodeCommit). Using the GitHub version, seeing the same `httplib.InvalidURL: nonnumeric port` error.

Loving that dulwich is completely native as there are not git clients in Lambda.

jelmer:
I was going to make changes to this pull request while merging it, but I haven't found time to do so. I'll comment here instead - if somebody can address these issues, that would be great.

.travis.yml
```diff
@@ -5,9 +5,9 @@ env:
 matrix:
   include:
     - python: "2.7"
-      env: TEST_REQUIRE="gevent greenlet geventhttpclient fastimport"
+      env: TEST_REQUIRE="gevent greenlet geventhttpclient fastimport mock"
```
jelmer: Please don't add extra dependencies. There should be no need for mocking here; you can just create a HttpGitClient and verify that the right attributes are set.

dulwich/client.py
```diff
@@ -997,13 +997,22 @@ def default_urllib2_opener(config):
 
 class HttpGitClient(GitClient):
 
-    def __init__(self, base_url, dumb=None, opener=None, config=None, **kwargs):
+    def __init__(self, base_url, dumb=None, opener=None, config=None,
+                 username=None, passwd=None, **kwargs):
```
jelmer: s/passwd/password/

dulwich/client.py
```diff
@@ -997,13 +997,22 @@ def default_urllib2_opener(config):
...
         if opener is None:
             self.opener = default_urllib2_opener(config)
         else:
             self.opener = opener
+
+        if passwd is None:
+            passwd = ""
+
+        if username:
+            pass_man=urllib2.HTTPPasswordMgrWithDefaultRealm()
```
jelmer: Please use spaces around equals signs, per PEP8

dulwich/client.py
```diff
@@ -997,13 +997,22 @@ def default_urllib2_opener(config):
...
+                 username=None, passwd=None, **kwargs):
         self._base_url = base_url.rstrip("/") + "/"
         self.dumb = dumb
         if opener is None:
             self.opener = default_urllib2_opener(config)
         else:
             self.opener = opener
+
+        if passwd is None:
+            passwd = ""
```
jelmer: No password specified is not the same as no password; you probably to check for 'username is not None and password is not None' on line 1012.

dulwich/client.py
```diff
@@ -1187,8 +1197,9 @@ def get_transport_and_path_from_url(url, config=None, **kwargs):
         return SSHGitClient(parsed.hostname, port=parsed.port,
                             username=parsed.username, **kwargs), path
     elif parsed.scheme in ('http', 'https'):
-        return HttpGitClient(urlparse.urlunparse(parsed), config=config,
-                **kwargs), parsed.path
+        base_url = urlparse.urlunparse(parsed._replace(netloc=host))
```
jelmer: Why not add this logic in HttpGitClient? That would also remove the need to add 'username' and 'password' arguments to it.

jelmer:
Merged with some modifications, thanks :)

Israphel:
is this available on 0.15.0 ?

jelmer:
No, it's not in 0.15.0 yet. It'll be in the next release, which I hope to release in the next couple of days.

jelmer:
0.16.0 is released now with these changes

PR #454, opened 2016-10-05
Support linked working directories

lrineau:
Support for linked working directories:
- Add `commondir()` (equivalent to `GIT_COMMON_DIR`)
- Read the `commondir` file, to set it.

See `git-worktree(1)` and `gitrepository-layout(5)`.


jelmer:
Thanks for the patch.

Can you please add some tests for the new behaviour. Also, you'd probably
want to use .get_named_file() rather than adding a new function.

lrineau:
I have:
- added a new testcase 
- actually fixed a bug, in `DiskRefsContainer.refpath()`: for a linked working tree, the `HEAD` ref is not to be read from the common dir but from the worktree dir,
- cleaned up the useless function.

lrineau:
https://github.com/jelmer/dulwich/pull/454/commits/0927deb7cd2ad24294b89e319ea060ed488acbba added `@skipIf` so that the new tests are skipped if `git_version()` < 2.5, to fix Travis CI tests.

dulwich/refs.py
```diff
@@ -403,8 +403,9 @@ def get_peeled(self, name):
 class DiskRefsContainer(RefsContainer):
     """Refs container that reads refs from disk."""
 
-    def __init__(self, path):
+    def __init__(self, path, **kwargs):
         self.path = path
+        self.worktree_path = kwargs["worktree"]
```
lrineau: Here I have chosen to use a named parameter, maybe I should have added a real second parameter.

jelmer: I think we should actually not make 'HEAD' a part of RefsContainer if it's working tree specific. Can you add a # TODO comment mentioning that.

jelmer: (on a separate note, this change is also an API change and so ideally you should only be adding *optional* arguments).

dulwich/repo.py
```diff
@@ -676,10 +677,17 @@ def __init__(self, root):
...
+        worktree_commondir = os.path.join(self._controldir, COMMONDIR)
+        if(os.path.isfile(worktree_commondir)):
+            with open(worktree_commondir, 'r') as f:
+                self._commondir = os.path.join(self.controldir(),\
+                                                   f.read().rstrip("\n"))
+        else:
+            self._commondir = self._controldir
         self.path = root
-        object_store = DiskObjectStore(os.path.join(self.controldir(),
+        object_store = DiskObjectStore(os.path.join(self.commondir(),
```
jelmer: Please add tests for this in dulwich/tests/test_repo.py. Dulwich should also just be tested against itself, we can't rely on C git being available.

dulwich/repo.py
```diff
@@ -676,10 +677,17 @@ def __init__(self, root):
             raise NotGitRepository(
                 "No git repository was found at %(path)s" % dict(path=root)
             )
+        worktree_commondir = os.path.join(self._controldir, COMMONDIR)
+        if(os.path.isfile(worktree_commondir)):
```
jelmer: Please just try to open this file, rather than probing for it ("ask forgiveness, not permission"). That saves a syscall per repo open.

dulwich/tests/compat/test_repository.py
```diff
@@ -119,3 +127,32 @@ def test_packed_objects(self):
...
+
+    def tearDown(self):
+        self._worktree_repo.close()
+        rmtree_ro(self._worktree_path)
+        self._repo = self._mainworktree_repo
+        super(WorkingTreeTestCase, self).tearDown()
+
+    def test_refs(self):
+        super(WorkingTreeTestCase, self).test_refs()
+        self.assertTrue(self._mainworktree_repo.refs.allkeys()==\
```
jelmer: Please use assertEquals/assertNotEquals.

These methods give useful output when a test fails.

jelmer: Per PEP8, please don't include unnecessary trailing backslashes

dulwich/repo.py
```diff
@@ -676,10 +677,17 @@ def __init__(self, root):
...
+            with open(worktree_commondir, 'r') as f:
+                self._commondir = os.path.join(self.controldir(),\
```
jelmer: Please just use self._get_named_file(), which also works for e.g. subclasses that aren't using local disk.

dulwich/refs.py
jelmer: E.g. have worktree_path as a path that defaults to just 'path'.

dulwich/repo.py
lrineau: I do not undertstand? How can I create a new working tree (à la git-worktree) with dulwich?

dulwich/repo.py
lrineau: Done in 14e7c73ef26378e3f82b2d0d9bd6e6d74a79dd42.

dulwich/tests/compat/test_repository.py
lrineau: Fixed in 7b05f09d856a79ada383b6e2b2c514b996cb55d8.

Sorry, I am rather new in Python programming ([I am rather a C++ guru][CGAL]). Thanks for the advice.

[CGAL]: http://github.com/CGAL/cgal

lrineau:
According to [`gitrepository-layout(5)`][*], the `GIT_COMMON_DIR` supersedes everything but:
- `HEAD`,
- and `index`.

In particular, `info/grafts` must be taken from `GIT_COMMON_DIR` as well. There should be a variant of `get_named_file()`... maybe with an optional parameter. I'll do it...

[*]: https://git-scm.com/docs/gitrepository-layout

lrineau:
> In particular, info/grafts must be taken from GIT_COMMON_DIR as well. There should be a variant of get_named_file()... maybe with an optional parameter. I'll do it...

Done in 101a56088fdf4e3c9a14e34919005f9f463cf7a1.

dulwich/repo.py
lrineau: ... Actually I have understood how the `tests/data/repos/` works. I have added the test in c74aa2452aba3fee822595deff18314e2387d29c.

lrineau:
I think I have now handled all your comments. Let me know.

dulwich/refs.py
```diff
@@ -403,8 +403,9 @@ def get_peeled(self, name):
 class DiskRefsContainer(RefsContainer):
     """Refs container that reads refs from disk."""
 
-    def __init__(self, path):
+    def __init__(self, path, **kwargs):
```
jelmer: There is no need to use **kwargs here. You can just use 'worktree_path=None', then use 'worktree_path or path' below.

dulwich/refs.py
```diff
@@ -450,7 +451,12 @@ def refpath(self, name):
             name = name.decode(sys.getfilesystemencoding())
         if os.path.sep != "/":
             name = name.replace("/", os.path.sep)
-        return os.path.join(self.path, name)
+        #TODO: as the 'HEAD' reference is working tree specific, it
+        # should actually not be a part of RefsContainer
+        if(name == b'HEAD'):
```
jelmer: Please don't use parentheses here, per PEP8 - they're not necessary

dulwich/repo.py
```diff
@@ -189,7 +190,7 @@ def _init_files(self, bare):
         self._put_named_file('config', f.getvalue())
         self._put_named_file(os.path.join('info', 'exclude'), b'')
 
-    def get_named_file(self, path):
+    def get_named_file(self, path, **kwargs):
```
jelmer: **kwargs isn't necessary here (and unused)

dulwich/tests/test_repository.py
```diff
@@ -521,6 +521,23 @@ def check(repo):
             check(nonbare)
             check(bare)
 
+    def test_working_tree(self):
+        temp_dir = tempfile.mkdtemp()
```
jelmer: Please create the new repository from scratch, don't rely on testdata. You can easily do this using Repo.init.

dulwich/repo.py
```diff
@@ -730,21 +748,23 @@ def _put_named_file(self, path, contents):
         with GitFile(os.path.join(self.controldir(), path), 'wb') as f:
             f.write(contents)
 
-    def get_named_file(self, path):
+    def get_named_file(self, path, **kwargs):
```
jelmer: See above - please use named arguments where possible.

dulwich/repo.py
```diff
@@ -730,21 +748,23 @@ def _put_named_file(self, path, contents):
...
         the control dir in a disk-based Repo, the object returned need not be
         pointing to a file in that location.
 
         :param path: The path to the file, relative to the control dir.
+        :param basedir: Optional argument that specifies an alternative to the control dir.
         :return: An open file object, or None if the file does not exist.
         """
         # TODO(dborowitz): sanitize filenames, since this is used directly by
         # the dumb web serving code.
+        basedir=kwargs.get("basedir", self.controldir())
```
jelmer: Use spaces around equals sign per PEP8.

dulwich/tests/compat/test_repository.py
```diff
@@ -119,3 +127,32 @@ def test_packed_objects(self):
     def test_all_objects(self):
         expected_shas = self._get_all_shas()
         self.assertShasMatch(expected_shas, iter(self._repo.object_store))
+
+
+@skipIf(git_version() < (2, 5), 'Git version must be >= 2.5')
+class WorkingTreeTestCase(ObjectStoreTestCase):
```
jelmer: This doesn't have anything to do with ObjectStore, so please derive from CompatTestCase

dulwich/tests/compat/test_repository.py
```diff
@@ -119,3 +127,32 @@ def test_packed_objects(self):
     def test_all_objects(self):
         expected_shas = self._get_all_shas()
         self.assertShasMatch(expected_shas, iter(self._repo.object_store))
+
+
+@skipIf(git_version() < (2, 5), 'Git version must be >= 2.5')
```
jelmer: Just set `min_git_version (see utils.py:208), rather than using skipIf

dulwich/tests/compat/test_repository.py
lrineau: `WorkingTreeTestCase` derives from `ObjectStoreTestCase` to re-test tests from `ObjectStoreTestCase` with a `Repo` that corresponds to a linked working tree. See the output from `python -m unittest -v dulwich.tests.compat_test_suite` (with the `-v` option):
```
test_all_objects (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
test_bare (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
test_head (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
test_loose_objects (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
test_packed_objects (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
test_refs (dulwich.tests.compat.test_repository.WorkingTreeTestCase) ... ok
```

dulwich/tests/test_repository.py
lrineau: I do not understand how to do that. Are there other tests doing something similar, so that I can understand what to do?

lrineau:
I think I have address all your comments but two:
- [This doesn't have anything to do with ObjectStore, so please derive from CompatTestCase][1]
- [Please create the new repository from scratch, don't rely on testdata.][2]

[1]: https://github.com/jelmer/dulwich/pull/454#discussion_r82505413
[2]: https://github.com/jelmer/dulwich/pull/454#discussion_r82505330

And I still have a Travis Ci build error that I do not now how to deal with.

dulwich/repo.py
```diff
@@ -676,18 +677,26 @@ def __init__(self, root):
             raise NotGitRepository(
                 "No git repository was found at %(path)s" % dict(path=root)
             )
+        commondir = self.get_named_file(COMMONDIR)
+        if commondir:
+            for line in commondir:
+                self._commondir = os.path.join(self.controldir(), line.rstrip("\n"))
```
lrineau: @jelmer Would you agree that I add an other argument `mode='rb'` to `get_named_file`?

lrineau:
I have manage to test locally with python3.5. I am rather confident that b88322e188134e382e1996059d5d1f1947c6d290 will fix the Travis Ci tests with python 3.4 and 3.5.

I have added a named parameter `mode` to `get_named_file`.

dulwich/tests/test_repository.py
jelmer: E.g. line 422 of this file.

lrineau: Done in c3c9a2ccf21e2afe0f64275d4da7f4c08e541a83.

jelmer:
Merged, thanks!

Issue #457, opened 2016-10-24
startswith first arg must be bytes or a tuple of bytes, not str in walk.py at line 259


TomasHubelbauer:
Windows 10 Home, Python 3.5.2, latest `dulwich` you get with `pip install`

```python
from dulwich.repo import Repo

r = Repo(".")
w = r.get_walker(paths=[".gitignore"], max_entries=1)
c = next(iter(w)).commit
print(c)
```

The `.gitignore` file exists in the repo.

```
Traceback (most recent call last):                                                                                                                                      
  File "test.py", line 5, in <module>                                                                                                                                   
    c = next(iter(w)).commit                                                                                                                                            
  File "C:\Users\Tom\AppData\Local\Programs\Python\Python35-32\lib\site-packages\dulwich\walk.py", line 323, in _next                                                   
    if self._should_return(entry):                                                                                                                                      
  File "C:\Users\Tom\AppData\Local\Programs\Python\Python35-32\lib\site-packages\dulwich\walk.py", line 309, in _should_return                                          
    if self._change_matches(change):                                                                                                                                    
  File "C:\Users\Tom\AppData\Local\Programs\Python\Python35-32\lib\site-packages\dulwich\walk.py", line 271, in _change_matches                                         
    if self._path_matches(new_path):                                                                                                                                    
  File "C:\Users\Tom\AppData\Local\Programs\Python\Python35-32\lib\site-packages\dulwich\walk.py", line 259, in _path_matches                                           
    if (changed_path.startswith(followed_path) and                                                                                                                      
TypeError: startswith first arg must be bytes or a tuple of bytes, not str 
```

The line numbers in the traceback are off by one or two in most cases because I've manually introduced `print` within the file while tracing this. 

https://github.com/jelmer/dulwich/blob/master/dulwich/walk.py#L257

Here, `changed_path` is bytes while `followed_path` is str.

I don't want to dig much deeper into this (lazy and do not know Python), hope you are able to figure this out with this information. My lazy attempt to overcome this was to add a bytes array into `w = r.get_walker(paths=[".gitignore"], max_entries=1)` instead of a str array:

```python
w = r.get_walker(paths=[".gitignore".encode("utf-8")], max_entries=1)
```

But that's just silly and of course didn't work at all.

jelmer:
Paths in Git repositories are bytestrings, so this behaviour is correct - though the error message should arguably be better.

See also https://github.com/jelmer/dulwich/issues/79 for an earlier discussion about this.

[...]

PR #458, opened 2016-10-24
Hack-fix str versus bytes

TomasHubelbauer:
Disclaimer: I don't know Python. http://stackoverflow.com/a/6273618/2715716

Related to #457


jelmer:
It's the job of the caller to encode paths - Git is encoding agnostic (upstream recommends using utf8, but other encodings can and are used). Arguably the error message should be clearer on this.

Issue #459, opened 2016-10-26
Wrong error message when socket connection is not establishing


ardumont:
Hello, 

for supposedly internal reasons, establishing the socket connection does not work when trying to fetch pack from a remote repository.
The thing is that then the error message is not about that.
It raises an UnboundLocalError and the initial error message is not provided.

Here is an extract of the stack trace:

```
Oct 26 14:09:31 uffizi python3[16259]: [2016-10-26 14:09:31,108: ERROR/MainProcess] Task swh.loader.git.tasks.ReaderGitRepository[a9df82ad-802f-4271-8c49-73d8ed4e4746] raised unexpected: UnboundLocalError("local variable 'err' referenced before assignment",)
Oct 26 14:09:31 uffizi python3[16259]: Traceback (most recent call last):
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/celery/app/trace.py", line 240, in trace_task
Oct 26 14:09:31 uffizi python3[16259]: R = retval = fun(*args, **kwargs)
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/celery/app/trace.py", line 437, in __protected_call__
Oct 26 14:09:31 uffizi python3[16259]: return self.run(*args, **kwargs)
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/swh/loader/git/tasks.py", line 53, in run
Oct 26 14:09:31 uffizi python3[16259]: return loader.load(repo_url)
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/swh/loader/git/reader.py", line 157, in load
Oct 26 14:09:31 uffizi python3[16259]: self.fetch_data()
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/swh/loader/git/updater.py", line 303, in fetch_data
Oct 26 14:09:31 uffizi python3[16259]: do_progress)
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/swh/loader/git/updater.py", line 232, in fetch_pack_from_origin
Oct 26 14:09:31 uffizi python3[16259]: progress=do_activity)
Oct 26 14:09:31 uffizi python3[16259]: File "/usr/lib/python3/dist-packages/dulwich/client.py", line 555, in fetch_pack
[...]

PR #460, opened 2016-10-26
Fix losing error message when socket connection fails

ardumont:
I don't know if the err variable here was intentional or not.
But this does not behave properly.
This breaks with UnboundLocalError, thus losing a precious information about the initial problem.

- Tests are still ok 
- runtime gives the right error message with this patch applied.

Related #459

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -672,7 +672,7 @@ class TCPGitClient(TraditionalGitClient):
             try:
                 s.connect(sockaddr)
                 break
-            except socket.error as err:
+            except socket.error:
                 if s is not None:
                     s.close()
                 s = None


ardumont:
> - Tests are still ok


```
python -m unittest  dulwich.tests.test_suite
.s......................................s..................................................................................................................................................s.............................sss............................................................................................................................................................................................................................................................................................................................................................s.............................................................................................................................................................................................................................................................................................................................s................................................s..............................................................ss.sss.ss..........................................sssssssssssssssssssssssssssssssssssssssssssssss
Ran 1096 tests in 60.749s

OK (skipped=63)
```

> runtime gives the right error message with this patch applied.

I've updated the issue with a sample extract of the unmasked error once the patch is applied.

Cheers,

jelmer:
This behaviour is actually intentional - it's supposed to overwrite the 'err' variable. See the associated bug report.

PR #467, opened 2016-12-15
Fix porcelain.reset to not ignore committish argument.

sonologic:
Spotted a bug, wrote test-case for it, fixed the bug.

This is just a prelude to fixing a more elaborate issue, which is that porcelain.reset() will leave files in the working directory that should be removed when doing a hard reset.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -20,6 +20,9 @@
 
  BUG FIXES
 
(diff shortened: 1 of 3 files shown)


jelmer:
Thanks, Koen!

Issue #479, opened 2017-01-10
repo.reset_index rewrite file stats even if no changes


khertan:
Does file stats like modification time should not be modified if no modification are made in the tree ?
As else doing a porcelain.pull erase all file modification time tracked in the git repository.

jelmer:
Merged, thanks.

PR #480, opened 2017-01-10
Avoid useless write and file stats change  #479

khertan:
Avoid writing file when no change happen and so writing file without modification will change file stats  #479

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -434,6 +434,11 @@ def build_file_from_blob(blob, mode, target_path, honor_filemode=True):
             else:
                 raise
     else:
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the pull request. Could you please also add a test for this new behaviour?

khertan:
Test case added, got some difficulties to test it, as my filesystem cache the metadatas. So using os.system('sync') call is a bit horrible and hackish, as it ll not works on Windows. But as os.sync isn t available before py3.3, i didn t see multi platform way to do it.

Any other idea ?

jelmer:
Merged, thanks!

Issue #478, opened 2017-01-10
test suite fails on windows due to TypeError in write_cache_time


koenfugro:
Lots of failures with stack-traces ending in:

File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\repo.py", line 934, in reset_index
validate_path_element=validate_path_element)
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 514, in build_index_from_tree
index.write()
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 220, in write
write_index_dict(f, self._byname)
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 175, in write_index_dict
write_index(f, entries_list)
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 165, in write_index
write_cache_entry(f, x)
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 125, in write_cache_entry
write_cache_time(f, ctime)
File "C:\Python27\lib\site-packages\dulwich-0.16.2-py2.7.egg\dulwich\index.py", line 93, in write_cache_time
raise TypeError(t)
TypeError: 1484050597

This is due to a 'formatting change' in commit f1ab124de0539ce07c42b1e2f09e03e571c62c26:

@@ -505,8 +505,7 @@ def build_index_from_tree(root_path, index_path, object_store, tree_id,
         # Add file to index
         st = os.lstat(full_path)
         if not honor_filemode:
-            st_tuple = (entry.mode, st.st_ino, st.st_dev, st.st_nlink, st.st_uid, st.st_gid, st.st_size, st.st_atime, st.st_mtime, st.st_ctime)
-            st = st.__class__(st_tuple)
+            st = st.__class__((entry.mode, ) + st[1:])
         index[entry.path] = index_entry_from_stat(st, entry.sha, 0)
 
     index.write()

Accessing the stat_result object as a tuple, on windows, will represent the time as a long, while if you access it as st.st_atime (or mtime, ctime) you get the actual float instead of the long representation..

jelmer:
On Tue, Jan 10, 2017 at 05:22:55AM -0800, koenfugro wrote:
> Lots of failures with stack-traces ending in:
> 
[...]

PR #481, opened 2017-01-11
Revert formatting change in index.py that breaks windows compatibility.

koenfugro:
Fix for issue https://github.com/jelmer/dulwich/issues/478.


jelmer:
Thanks, merged.

Issue #482, opened 2017-01-11
Concurrent access to a repository on Windows sometimes fails


wme-at-contact-de:
I have been testing Dulwich as a Git-Server on Windows, using multiple clients cloning and pushing the same repository concurrently. Most of the time, this works fine.
However, sometimes cloning fails with a traceback like this:
[traceback.txt](https://github.com/jelmer/dulwich/files/698672/traceback.txt)

The setup is Apache 2.4.6 with mod_wsgi on Windows 7.
I'm using a minimal WSGI application created with "make_wsgi_chain" (from dulwich/web.py).

Any idea what could be the problem, or whether this is a bug?

jelmer:
As far as I can tell, the order in which Dulwich writes files is correct.

This might be caused by a lack of filesystem syncs.

PR #484, opened 2017-01-12
Fixing #482

wme-at-contact-de:
I believe I have found the problem with #482: Pack files were sometimes loaded into the object store when they were not yet completely written by a different thread. It seems like there is a very small time window where this could happen.

With this change, my automated tests run without errors (multiple clients repeatedly cloning and pushing to the same server repository, running on Windows with Apache and mod_wsgi).


jelmer:
Merged, thanks!

Issue #486, opened 2017-01-17
Tests emit errors under windows


koenfugro:
While tests pass, there's a number of errors emitted (see below). This is basically due to the expectation in test_index.py that 'sync' is an os command that can be started via os.system. It is not on windows, so basically the sync() calls are a no-op on that platform. I would suggest to make sync() a real no-op (lambda: None) on win32. The errors are harmless, but they go against my obsession for an error- and warning-free world!

```
D:\Sources\dulwich>py -2 -m unittest dulwich.tests.test_suite
.s................................................................................................................................ss.s.s.....................s.s
.................................................................ssssss.....'sync' is not recognized as an internal or external command,
operable program or batch file.
'sync' is not recognized as an internal or external command,
operable program or batch file.
'sync' is not recognized as an internal or external command,
operable program or batch file.
'sync' is not recognized as an internal or external command,
operable program or batch file.
.s.......................................................................................................................................s..s.s.................
................................................................................................................................................................
.........................s......................................................................................................................................
...............................s.................................sss.s..........................................................................................
........................s.....................................x.......ss.s........................ssssssssssssssssssssssssssss.....sssssssssssssssssssssssssssss
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss
----------------------------------------------------------------------
Ran 1110 tests in 84.942s

OK (skipped=158, expected failures=1)

D:\Sources\dulwich>

```

jelmer:
Merged your pull request, thanks!

PR #487, opened 2017-01-17
Squelch errors in test_index under windows.

koenfugro:
Fix for issue https://github.com/jelmer/dulwich/issues/486

### DIFF
--- a/dulwich/tests/test_index.py
+++ b/dulwich/tests/test_index.py
@@ -368,8 +368,8 @@ class BuildIndexTests(TestCase):
             self.assertEqual(['d'],
                 sorted(os.listdir(os.path.join(repo.path, 'c'))))
 
+    @skipIf(not getattr(os, 'sync', None), 'Requires sync support')
     def test_norewrite(self):
-        sync = getattr(os, 'sync', lambda: os.system('sync'))
         repo_dir = tempfile.mkdtemp()
         self.addCleanup(shutil.rmtree, repo_dir)
         with Repo.init(repo_dir) as repo:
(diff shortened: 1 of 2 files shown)


dulwich/tests/test_index.py
```diff
@@ -369,7 +369,10 @@ def test_nonempty(self):
                 sorted(os.listdir(os.path.join(repo.path, 'c'))))
 
     def test_norewrite(self):
-        sync = getattr(os, 'sync', lambda: os.system('sync'))
+        if sys.platform != 'win32':
```
jelmer: Ignoring these is going to lead to spurious errors. Perhaps we can just stick to calling os.sync when it is available and skipping this test otherwise (e.g. using @skipIf)? os.sync is available in python3 on windows too AFAIK.

koenfugro: Hmm, i don't think so, according to https://docs.python.org/3/library/os.html:

```
os.sync()
Force write of everything to disk.

Availability: Unix.

New in version 3.3.

```

I'll go with the approach of skipping the test on windows .. I've done some cursory research (googling it) and it seems sync-ing system-wide isn't something one easily does on windows.

jelmer: Seems reasonable to me :)

koenfugro: Hmm, skipping the test if os.sync is not present will also have it skipped when using python2. I can check if sync is a system command of course, but that's going to be a pretty complicated check for finding out whether to skip the test..

Just checking if os.sync exists is most straight-forward it seems, but feels like throwing out the baby with the bathwater..

jelmer: Either seems reasonable. It's fine to have the test python3-only, I guess. Just so long as we don't run the test with a dummy os.sync, as that'll lead to spurious failures.

koenfugro:
Ok, force pushed the ifskip-solution.

jelmer:
Merged, thanks!

PR #502, opened 2017-02-21
SHAFile: Do not trim trailing newlines from extra headers

olasd:
Hi,

Software Heritage has been throwing a lot of data to dulwich. Around 50 million repositories, 700 million commits. Of course, we have stumbled upon corner cases (e.g. #413). Here's another one, affecting a handful of commits in the wild:

Some commits contain extra headers with trailing newlines. We need to keep those
intact if we want to be able to round-trip data back to the original.

The attached test is one such commit found in the wild (twitter/secureheaders@eff3da9721aae64c25d31b890a2817c21907af96).

Cheers,
Nicolas

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -609,6 +609,12 @@ def _parse_message(chunks):
     v = ""
     eof = False
 
+    def _strip_last_newline(value):
+        """Strip the last newline from value"""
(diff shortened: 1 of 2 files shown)


jelmer:
Merged, thanks!

Issue #509, opened 2017-03-15
ValueError in _parse_message() when using PyPy


haypo:
Hi,

I'm working on the http://github.com/python/performance benchmark suite which has a "dulwich_log" benchmark. When I tried to upgrade Dulwich from 0.16.2 to 0.17.1, I got the following error using PyPy 5.4.0 on Fedora 25:

```
INFO:root:Running `/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/bin/python /home/haypo/prog/python/performance/performance/benchmarks/bm_dulwich_log.py --debug-single-value --output /tmp/tmpZjCAtn`
Traceback (most recent call last):
  File "/home/haypo/prog/python/performance/performance/benchmarks/bm_dulwich_log.py", line 27, in <module>
    runner.bench_func('dulwich_log', iter_all_commits, repo)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 636, in bench_func
    return self._main(name, time_func, inner_loops, metadata)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 547, in _main
    bench = self._worker(name, time_func, inner_loops, metadata)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 510, in _worker
    inner_loops)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 468, in _worker_run_bench_mem
    inner_loops)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 448, in _worker_run_bench
    loops=loops, nvalue=args.values)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 372, in _run_bench
    raw_value = time_func(loops)
  File "/home/haypo/prog/python/performance/venv/pypy2.7-b5a8c1089518/site-packages/perf/_runner.py", line 616, in time_func
    local_func(*local_args)
  File "/home/haypo/prog/python/performance/performance/benchmarks/bm_dulwich_log.py", line 14, in iter_all_commits
    for entry in repo.get_walker(head):
[...]

PR #510, opened 2017-03-16
Replace PyString_AsString with PyString_AS_STRING

haypo:
Fix issue #509: Replace PyString_AsString with PyString_AS_STRING to
workaround a PyPy bug in cpyext (PyPy older than 5.6).


jelmer:
Merged, thanks!

PR #519, opened 2017-04-25
Fix TypeError: mix str and non-str arguments

dzhuang:
``path`` is supposed to be a bytestring according to [here](https://github.com/jelmer/dulwich/blob/master/dulwich/client.py#L1168) and [there](https://github.com/jelmer/dulwich/blob/master/dulwich/client.py#L1181), while ``self._base_url`` is a ``str``, that will result in ``TypeError: Cannot mix str and non-str arguments`` when calling ``urlparse.urljoin``.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1099,6 +1099,8 @@ class HttpGitClient(GitClient):
             type(self).__name__, self._base_url, self.dumb)
 
     def _get_url(self, path):
+        if not isinstance(path, str):
+            path = path.decode(sys.getfilesystemencoding())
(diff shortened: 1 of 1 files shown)


jelmer:
Why the close?

jelmer:
(Sorry, I haven't had time to look at this PR yet)

dzhuang:
Sorry, I am not sure whether it is a documentation issue, because the tests were working well without this PR.

dzhuang:
Sorry, I pushed a squashed commit and the PR cannot be repoened, I will create another PR.

PR #525, opened 2017-07-04
Wrong argument order in ignore.py

segevfiner:


jelmer:
Merged (with additional test change). Thanks!

PR #527, opened 2017-07-04
Git ignore *.pyd

segevfiner:
Compiled extension modules on Windows. Created by `setup.py build_ext --inplace`/`setup.py develop`/`pip install -e`.


jelmer:
Merged, thanks!

Issue #526, opened 2017-07-04
.gitignore consistency with Git


segevfiner:
Getting gitignore behavior just right seems to be quite hard. (I think even libgit2 still has some issues with it 😛) This are some things that I encountered which are different in the current implementation in master:
- [x] A pattern including a trailing slash should match only directories. That slash should not be considered in deciding whether a pattern is a filename glob or a path glob. &ndash; PR #528
- [x] I think that if you ignore a file, re-include it (A pattern with a '!') and than ignore it again. It should be ignored. (The code currently does `return`). &ndash; PR #529
- [x] A pattern without a trailing slash should match both files and directories. &ndash; PR #530 
- [x] The behavior in regards to ignoring directories is quite finicky. If a directory is ignored,  Git doesn't recurse into it, meaning that all the files in it are ignored. Also, "negated" patterns don't affect anything in such directories. There is no such code ATM in Dulwich but this is important to remember.
- [x] There probably needs to be a way to properly handle .gitignore files from subdirectories. Their patterns should only apply below them and path patterns from parent directories should be anchored to the directory where they are defined. It's unclear how to currently achieve this if at all possible with the current implementation. We probably need to accept the repository relative path to the directory or .gitignore file, and either alter the regex (carefully!) or check the path as an extra check before checking for patterns in the IgnoreFilter (I think that's what the real Git does).
- [x] Git matches patterns case insensitively on case insensitive file systems. i.e. on Windows.
       Git enables case insensitivity based on the `core.ignorecase` config.
- [x] Trailing spaces are ignored unless they are quoted with backslash ("\\").
- [x] Handle .gitignore files with \r\n line endings &ndash; PR https://github.com/jelmer/dulwich/pull/535
- [x] We should respect a trailing slash in an absolute path passed to `IgnoredFilterManager.is_ignored`.

Reference: https://git-scm.com/docs/gitignore

jelmer:
I believe most of these are implemented - assuming you pass in relative paths to the directory the gitignore file was found in, and add a trailing slash to the path if your checking for a directory. I just followed the syntax described in the URL you mention.

Do you have specific cases that work differently with the Dulwich parser vs C git?

The documentation suggests that you can't redo an ignore that was previously excluded.

Segev Finer <notifications@github.com> schreef op 4 juli 2017 18:55:38 GMT+01:00:

segevfiner:
Note that I edited the issue.

I probably have used this module incorrectly. I will try later to use it
[...]

PR #528, opened 2017-07-04
Correctly handle include, exclude than include in .gitignore

segevfiner:
For example, for a .gitignore file like:
```
test
!test
test
```

"test" should be ignored.

Related issue: #526

## Reproduction
```sh
git init temp && cd temp
touch test
cat > .gitignore <<EOF
test
!test
test
EOF
git check-ignore -v test
```

Output:
```
.gitignore:3:test       test
```
(The file is ignored)

Dulwich (before fix):
```py
...
>>> a.is_ignored('test')
False
```


jelmer:
Merged, thanks!

PR #529, opened 2017-07-04
A pattern with only a trailing slash should be treated as a glob

segevfiner:
A pattern like "bar/" should still ignore "foo/bar/", but "/bar/" should
not.

Related issue: #526

## Reproduction
```sh
git init temp && cd temp
mkdir -p test/foo
cat > .gitignore <<EOF
foo/
EOF
git check-ignore -v test/foo
```

Output:
```
.gitignore:1:foo/       test/foo
```
(The file is ignored)

Dulwich (before fix):
```py
...
>>> a.is_ignored('test/foo/')
# None
```

### DIFF
--- a/dulwich/ignore.py
+++ b/dulwich/ignore.py
@@ -36,7 +36,7 @@ def translate(pat):
 
     res = b'(?ms)'
 
-    if b'/' not in pat:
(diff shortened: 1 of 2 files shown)


jelmer:
Merged, thanks!

PR #530, opened 2017-07-05
gitignore patterns without a trailing slash should match directories too

segevfiner:
gitignore patterns with a trailing slash should match both directories and files, while patterns with a trailing slash should match only directories.

I added a '/?' to any pattern that doesn't end in a slash. Please review if this is the right way to handle this.

Related issue: #526

## Reproduction
A pattern without a trailing slash should match both files and directories:
```sh
git init temp && cd temp
mkdir -p foo/bar
cat > .gitignore <<EOF
bar
EOF
git check-ignore -v foo/bar
```

Output:
```
.gitignore:1:bar.       foo/bar/
```

Dulwich (Before fix):
```py
... 
>>> a.is_ignored('foo/bar/')
# None
```

### DIFF
--- a/dulwich/ignore.py
+++ b/dulwich/ignore.py
@@ -81,6 +81,10 @@ def translate(pat):
                 res = res + b'[' + stuff + b']'
         else:
             res = res + re.escape(c)
+
+    if not res.endswith(b'/'):
+        res = res + b'/?'
+
     return res + b'\Z'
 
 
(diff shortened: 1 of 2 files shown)


jelmer:
That seems reasonable to me, I can't think of any counterexamples. We can always revisit this if we find any.

segevfiner:
@jelmer I played with this in https://github.com/jelmer/dulwich/issues/526#issuecomment-313249344 and it seems to work correctly. For example: It ignores the "\_\_pycache__" directories that are ignored with a pattern without a trailing slash in CPython's .gitignore file.

jelmer:
Great, thanks! I'll close this pull request as this change has been merged.

Issue #543, opened 2017-08-04
diff_tree output differs from git diff


Trane9991:
Hello, 
I'm trying to get and parse diff of 2 commits, and I noticed the difference between diff from git and diff from dulwich.

Diff from dulwich:
```
diff --git a/lala /dev/null
deleted mode 100644
index d800886..0000000
--- a/lala
+++ /dev/null
@@ -1,1 +1,0 @@
-123
\ No newline at end of file
```

Diff from git:
```
diff --git a/lala b/lala
deleted file mode 100644
index d800886..0000000
--- a/lala
+++ /dev/null
@@ -1 +0,0 @@
-123
\ No newline at end of file
```

Then, I'm trying to find added and deleted files between this commits with https://github.com/matiasb/python-unidiff.

unidiff output from dulwich diff:
```
Summary
-------
lala: +0 additions, -1 deletions

1 modified file(s), 0 added file(s), 0 removed file(s)
Total: 0 addition(s), 1 deletion(s)
```

unidiff output from git diff:
```
Summary
-------
lala: +0 additions, -1 deletions

0 modified file(s), 0 added file(s), 1 removed file(s)
Total: 0 addition(s), 1 deletion(s)
```

Removed file not being detected.

Trane9991:
Looks like problem in this line:
`@@ -1,1 +1,0 @@`   it should be `@@ -1,1 +0,0 @@`

PR #544, opened 2017-08-04
Fix unified_diff

Trane9991:
This closes issue #543 

replaced custom unified_diff implementation with the call to difflib.unified_diff


jelmer:
Thanks for the bug report and the patch. Using difflib.unified_diff introduces another difference with the existing git output - it removes the "\ No newline at end of file" lines. Can you patch the existing implementation instead?

Issue #554, opened 2017-08-26
Avoid leaking pre-existing pack instance when adding known pack


earlchew:
```
        """Add a newly appeared pack to the cache by path.
        """
        self._pack_cache[base_name] = pack
```

jelmer:
Merged, thanks!

PR #555, opened 2017-08-26
Issue 554

earlchew:
https://github.com/jelmer/dulwich/issues/554

Included is a second commit that provides a slight simplification of the repack() implementation.


jelmer:
Merged, thanks!

Issue #556, opened 2017-08-26
When committing a pack, file should be flushed prior to fsync() of fd


earlchew:
```
        def commit():
            os.fsync(fd)
            f.close()
```

jelmer:
On Sat, Aug 26, 2017 at 07:58:51AM -0700, earlchew wrote:
> ```
>         def commit():
>             os.fsync(fd)
>             f.close()
> ```

There are probably a few places like this where Dulwich relies on
ext3 data=ordered behaviour. We need a more holistic approach to this:

When writing refs, packs, objects and the index, flush + fsync
before closing: perhaps just do this in GitFile.close ?

earlchew:
I think you are proposing that `add_pack()` just use `class GitFile` directly (rather than resorting to `tempfile.mkstemp()`), and that the flush-fsync-close semantics be applied to `class GitFile`.

Is that what you are suggesting?

jelmer:
On Tue, Aug 29, 2017 at 01:14:06AM +0000, earlchew wrote:
> I think you are proposing that `add_pack()` just use `class GitFile` directly (rather than resorting to `tempfile.mkstemp()`), and that the flush-fsync-close semantics be applied to `class GitFile`.

> Is that what you are suggesting?
Yep, I am indeed suggesting the flush-fsync-close semantics are added to GitFile.

We can't use GitFile directly in add_pack(), because it requires the
name of the file be known upfront - and the name of a pack file isn't
known until the file is completely written (as it includes a SHA of
all of the objects inside).

earlchew:
> because it requires the name of the file be known upfront 

Well, that's true, but I think the name is not actually required until the file is closed, so if a `rename()` method were to be supported, the correct name could be imposed at any time prior to that.

Additionally, the current temporary flie semantics could be supported by allowing for an additional argument to the `GitFile` constructor:

```
def GitFile(...., fd=None):
    ...
    if fd is None:
        fd = os.open(...)
    self._file = os.fdopen(fd, ...)
```

jelmer:
On Tue, Aug 29, 2017 at 03:21:29PM +0000, earlchew wrote:
> > because it requires the name of the file be known upfront 
> 
> Well, that's true, but I think the name is not actually required until the file is closed, so if a `rename()` method were to be supported, the correct name could be imposed at any time prior to that.
> 
> Additionally, the current temporary flie semantics could be supported by allowing for an additional argument to the `GitFile` constructor:
> 
> ```
[...]

PR #557, opened 2017-08-26
[object_store] Flush before fsync when adding pack

earlchew:
https://github.com/jelmer/dulwich/issues/556

Signed-off-by: Earl Chew <earl_chew@yahoo.com>


jelmer:
I've merged this, but as mentioned in the bug: this is just one instance, I think we need a more holistic approach.

Issue #567, opened 2017-10-30
Commit with overflow date field does not raise concerns


ardumont:
Hello,

## TL;DR

As a user, i'd like to be notified when a commit field date (author/committer) can cause overflow error.

## Details

I have a repository on disk which holds a well formatted commit except
that the author/committer date can cause overflow error.
But I don't see checks on those fields even though some checks exists for other objects's date field (tag for example).

Here is a partial example commit with such properties:

```
commit 24b03c53f91ec7719928abfeee77d61ad995f5ed
tree 7b6189adcaa2f420f4aa44e1550c3a7de659ad2f
parent 416c46f96c648c0effb9f6adccae3e10d2d7b9f9
author Igor Sysoev <igor@sysoev.ru> 18446743887488505614 +42707004
committer Kirill A. Korinskiy <catap@catap.ru> 18446743887488505614 +42707004

    nginx 0.7.63
    ...
```

Loading it through dulwich, everything seems fine since the object is found:
```
>>> from dulwich.repo import Repo
>>> r = Repo('./mainline.git')
>>> c = r[b'24b03c53f91ec7719928abfeee77d61ad995f5ed']
>>> c
<Commit b'24b03c53f91ec7719928abfeee77d61ad995f5ed'>
>>> c._author_time
18446743887488505614
>>> c.check()
>>>
```

`git fsck` (git version 1:2.11.0-3+deb9u2) reveals that error:

```
git fsck 2>&1 | grep 24b03c53f91ec7719928abfeee77d61ad995f5ed
error in commit 24b03c53f91ec7719928abfeee77d61ad995f5ed: badDateOverflow: invalid author/committer line - date causes integer overflow
```

But, there exists checks on date for other objects (tag for
one). For example, taking another repository with a wrong release date:
```
$ git show --format=raw ae51106031a0bb39a8def57a8592f70116487eab
# git show does not display the date here
#...
tag HYLAFAX-4_4_4-rc1
Tagger: Aidan Van Dyk <aidan@ifax.com>

commit 222d9a22c53ab359cc1b1bf94d01c34369fa2d5c
tree 802f3ede8ded6240e1c9d15ee4a1a29165d1fb8d
parent 815a7290a929509e1a4ffbbc493d25457f10def6
author Aidan Van Dyk <aidan@ifax.com> 1198083348 +0000
committer Aidan Van Dyk <aidan@ifax.com> 1198083348 +0000

    HylaFAX 4.4.4-rc1

[...]

PR #568, opened 2017-10-30
Add some more checks on commit/tag

ardumont:
Here is my take on #567.

Cheers,

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -63,6 +63,9 @@ _TAGGER_HEADER = b'tagger'
 S_IFGITLINK = 0o160000
 
 
+MAX_TIME = 9223372036854775807  # (2**63) - 1 - signed long int max
+
+
 def S_ISGITLINK(m):
     """Check if a mode indicates a submodule.
 
@@ -188,6 +191,20 @@ def check_identity(identity, error_msg):
             or identity.find(b'>', email_end + 1) >= 0
             or not identity.endswith(b'>')):
         raise ObjectFormatException(error_msg)
+
+
+def check_time(time_seconds):
+    """Check if the specified time is not prone to overflow error.
+
+    This will raise an exception if the time is not valid.
+
+    :param time_info: author/committer/tagger info
+
+    """
+    # Prevent overflow error
+    if time_seconds > MAX_TIME:
+        raise ObjectFormatException(
+            'Date field should not exceed %s' % MAX_TIME)
 
 
 def git_line(*items):
@@ -694,6 +711,9 @@ class Tag(ShaFile):
         if getattr(self, "_tagger", None):
             check_identity(self._tagger, "invalid tagger")
 
+        self._check_has_member("_tag_time", "missing tag time")
+        check_time(self._tag_time)
+
         last = None
         for field, _ in _parse_message(self._chunked_text):
             if field == _OBJECT_HEADER and last is not None:
@@ -742,23 +762,10 @@ class Tag(ShaFile):
             elif field == _TAG_HEADER:
                 self._name = value
             elif field == _TAGGER_HEADER:
-                try:
-                    sep = value.index(b'> ')
-                except ValueError:
-                    self._tagger = value
-                    self._tag_time = None
-                    self._tag_timezone = None
-                    self._tag_timezone_neg_utc = False
-                else:
-                    self._tagger = value[0:sep+1]
-                    try:
-                        (timetext, timezonetext) = (
-                                value[sep+2:].rsplit(b' ', 1))
-                        self._tag_time = int(timetext)
-                        self._tag_timezone, self._tag_timezone_neg_utc = (
-                                parse_timezone(timezonetext))
-                    except ValueError as e:
-                        raise ObjectFormatException(e)
+                (self._tagger,
+                 self._tag_time,
+                 (self._tag_timezone,
+                  self._tag_timezone_neg_utc)) = parse_time_entry(value)
             elif field is None:
                 self._message = value
             else:
@@ -1084,6 +1091,29 @@ def format_timezone(offset, unnecessary_negative_timezone=False):
             (sign, offset / 3600, (offset / 60) % 60)).encode('ascii')
 
 
+def parse_time_entry(value):
+    """Parse time entry behavior
+
+    :param value: Bytes representing a git commit/tag line
+    :raise: ObjectFormatException in case of parsing error (malformed
+            field date)
+    :return: Tuple of (author, time, (timezone, timezone_neg_utc))
+    """
+    try:
+        sep = value.index(b'> ')
+    except ValueError:
+        return (value, None, (None, False))
+    try:
+        person = value[0:sep+1]
+        rest = value[sep+2:]
+        timetext, timezonetext = rest.rsplit(b' ', 1)
+        time = int(timetext)
+        timezone, timezone_neg_utc = parse_timezone(timezonetext)
+    except ValueError as e:
+        raise ObjectFormatException(e)
+    return person, time, (timezone, timezone_neg_utc)
+
+
 def parse_commit(chunks):
     """Parse a commit object from chunks.
 
@@ -1108,14 +1138,9 @@ def parse_commit(chunks):
         elif field == _PARENT_HEADER:
             parents.append(value)
         elif field == _AUTHOR_HEADER:
-            author, timetext, timezonetext = value.rsplit(b' ', 2)
-            author_time = int(timetext)
-            author_info = (author, author_time, parse_timezone(timezonetext))
+            author_info = parse_time_entry(value)
         elif field == _COMMITTER_HEADER:
-            committer, timetext, timezonetext = value.rsplit(b' ', 2)
-            commit_time = int(timetext)
-            commit_info = (
-                    committer, commit_time, parse_timezone(timezonetext))
+            commit_info = parse_time_entry(value)
         elif field == _ENCODING_HEADER:
             encoding = value
         elif field == _MERGETAG_HEADER:
@@ -1177,7 +1202,8 @@ class Commit(ShaFile):
         self._check_has_member("_tree", "missing tree")
         self._check_has_member("_author", "missing author")
         self._check_has_member("_committer", "missing committer")
-        # times are currently checked when set
+        self._check_has_member("_author_time", "missing author time")
+        self._check_has_member("_commit_time", "missing commit time")
 
         for parent in self._parents:
             check_hexsha(parent, "invalid parent sha")
@@ -1185,6 +1211,9 @@ class Commit(ShaFile):
 
         check_identity(self._author, "invalid author")
         check_identity(self._committer, "invalid committer")
+
+        check_time(self._author_time)
+        check_time(self._commit_time)
 
         last = None
         for field, _ in _parse_message(self._chunked_text):
--- a/dulwich/tests/test_objects.py
+++ b/dulwich/tests/test_objects.py
@@ -56,6 +56,7 @@ from dulwich.objects import (
     _parse_tree_py,
     sorted_tree_items,
     _sorted_tree_items_py,
+    MAX_TIME
     )
 from dulwich.tests import (
     TestCase,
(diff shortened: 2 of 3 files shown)


dulwich/objects.py
```diff
@@ -1084,6 +1072,30 @@ def format_timezone(offset, unnecessary_negative_timezone=False):
...
+    except ValueError:
+        return (value, None, (None, False))
+    try:
+        person = value[0:sep+1]
+        rest = value[sep+2:]
+        timetext, timezonetext = rest.rsplit(b' ', 1)
+        time = int(timetext)
+        timezone, timezone_neg_utc = parse_timezone(timezonetext)
+        # checking for overflow error
+        datetime.datetime.fromtimestamp(time)
```
jelmer: This is an expensive way of checking the time, and depends on the implementation of datetime. Could you just check if the time was higher than the supported CGit value in .check() ?

dulwich/tests/test_objects.py
```diff
@@ -639,6 +639,34 @@ def test_check_order(self):
...
+            Commit,
+            self.make_commit_text(
+                tree=b'7b6189adcaa2f420f4aa44e1550c3a7de659ad2f',
+                parents=[b'416c46f96c648c0effb9f6adccae3e10d2d7b9f9'],
+                author=default_committer,
+                committer=committer_with_wrong_time,
+                message=b'an awesome explanation ../b\n'))
+
+    def test_check_commit_with_overflow_date(self):
+        # committer Kirill A. Korinskiy <catap@catap.ru> 18446743887488505614 +42707004
```
jelmer: This line is too long.

dulwich/objects.py
ardumont: Sure, will look into that.
Thanks for the heads up!

dulwich/tests/test_objects.py
ardumont: Ah yes, even though i read it [in the doc](https://github.com/jelmer/dulwich/blob/master/CONTRIBUTING.md#coding-style), i completely forgot to check for pep8 violations. Sorry about that.
Will fix.

dulwich/objects.py
ardumont: I need some help here, please. Here are my findings.

According to git fsck's code, [here is
the check on parsing the date and failing if error](https://github.com/git/git/blob/master/fsck.c#L695-L696)..

The definition of the function is in [date.c](https://github.com/git/git/blob/master/date.c#L1206-L1221).
This checks for a TIME_MAX value.

[Through the hide-and-seek game of header inclusion, we can then find the definition of that value](https://github.com/git/git/blob/master/git-compat-util.h#L328).
(from: date.c -> cache.h -> git-compat-util.h)

Its value is `UINTMAX_MAX` (unsigned integer max value).
So, is 2^32 sounds good enough?

Note: I interpreted the CGIT value you mentioned as the main git repository.
I'm not so sure i interpreted right though :)

dulwich/objects.py
```diff
@@ -1072,6 +1072,9 @@ def format_timezone(offset, unnecessary_negative_timezone=False):
             (sign, offset / 3600, (offset / 60) % 60)).encode('ascii')
 
 
+MAX_TIME = 4294967296  # 2**32
```
jelmer: Is this the same limit that C Git uses?

dulwich/objects.py
```diff
@@ -1084,6 +1072,35 @@ def format_timezone(offset, unnecessary_negative_timezone=False):
...
+    try:
+        person = value[0:sep+1]
+        rest = value[sep+2:]
+        timetext, timezonetext = rest.rsplit(b' ', 1)
+        time = int(timetext)
+        timezone, timezone_neg_utc = parse_timezone(timezonetext)
+    except ValueError as e:
+        raise ObjectFormatException(e)
+    # Prevent overflow error
+    if time > MAX_TIME:
```
jelmer: Please have this fail on .check(), not at parse time - similar to C Git. In other words, reading this incorrect times will work but .check will fail.

jelmer:
On Tue, Oct 31, 2017 at 10:58:54AM -0700, Antoine R. Dumont wrote:
> ardumont commented on this pull request.
> 
> 
> 
> > +    :param value: Bytes representing a git commit/tag line
> +    :raise: ObjectFormatException in case of parsing error.
> +    :return: Tuple of (author, time, (timezone, timezone_neg_utc))
> +    """
> +    try:
> +        sep = value.index(b'> ')
> +    except ValueError:
> +        return (value, None, (None, False))
> +    try:
> +        person = value[0:sep+1]
> +        rest = value[sep+2:]
> +        timetext, timezonetext = rest.rsplit(b' ', 1)
> +        time = int(timetext)
> +        timezone, timezone_neg_utc = parse_timezone(timezonetext)
> +        # checking for overflow error
> +        datetime.datetime.fromtimestamp(time)
> 
> I need some help here, please. Here are my findings.
> 
> According to git fsck's code, [here is
> the check on parsing the date and failing if error](https://github.com/git/git/blob/master/fsck.c#L695-L696)..
> 
> The definition of the function is in [date.c](https://github.com/git/git/blob/master/date.c#L1206-L1221).
> This checks for a TIME_MAX value.
> 
> [Through the hide-and-seek game of header inclusion, we can then find the definition of that value](https://github.com/git/git/blob/master/git-compat-util.h#L328).
> (from: date.c -> cache.h -> git-compat-util.h)
> 
> Its value is `UINTMAX_MAX` (unsigned integer max value).
> So, is 2^32 sounds good enough?
On my system, UINTMAX_MAX is 18446744073709551615 (from stdint.h), or
2**64. That's the same as sys.maxint*2 (since int is signed) in
Python; could we use that?

2**32 is only 4294967296.

> Note: I interpreted the CGIT value you mentioned as the main git repository.
> I'm not so sure i interpreted right though :)
Yep, that's indeed what I meant :)

-- 

dulwich/objects.py
ardumont: I'm not certain of it.
Thus my question here https://github.com/jelmer/dulwich/pull/568#discussion_r148078638.

dulwich/objects.py
ardumont: Right!

ardumont:
Note, I have pushed multiple commits and all.
I'll rebase (and possibly squash commits :) when we find a reasonable implementation.

ardumont:
> On my system, UINTMAX_MAX is 18446744073709551615 (from stdint.h), or
2**64. That's the same as sys.maxint*2 (since int is signed) in
Python; could we use that?

Well, that's where the trouble lies. This is too high (well for
python3 on my box at least).

I guess the question might be, where do we draw the line between technicals (c/python runtimes) and
reality (we are reading stuff from the past)?

2^32 boundary sets us in the future already (but not that much, year
2106).

2^64 is not even possible for the python runtime to work with (well as
you mentioned earlier depending on the date library's implementation
used, i settled on datetime since it's in native python).

Working on exponents, here is my experiment to see when we can reach
(~ year 6325):

```
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
[GCC 6.3.0 20170118] on linux
>>> import datetime
>>> for e in range(32, 64): print('2^%s' % e); d= datetime.datetime.fromtimestamp(2**e); print('date: %s' % d);
...
2^32
date: 2106-02-07 07:28:16
2^33
date: 2242-03-16 13:56:32
2^34
date: 2514-05-30 03:53:04
2^35
date: 3058-10-26 05:46:08
2^36
date: 4147-08-20 09:32:16
2^37
date: 6325-04-08 17:04:32
2^38
ValueError: year is out of range
```

As expected, with that limit (my initial max time btw), the tests created (from the real sample i have mentioned in #567) won't fail.
But since 2^38 sets us in year 6325 already... 2^64's year is clearly way out there in terms of current reality :)

```
$ python3 -m unittest dulwich/tests/test_objects.py
....................F...................................F..........................s..s.s...
FAIL: test_check_commit_with_overflow_date (dulwich.tests.test_objects.CommitParseTests)
Date with overflow should raise an ObjectFormatException when checked
  File "/home/tony/repo/public/dulwich/dulwich/tests/test_objects.py", line 672, in test_check_commit_with_overflow_date
    commit.check()
AssertionError: ObjectFormatException not raised

FAIL: test_check_tag_with_overflow_time (dulwich.tests.test_objects.TagParseTests)
Date with overflow should raise an ObjectFormatException when checked
  File "/home/tony/repo/public/dulwich/dulwich/tests/test_objects.py", line 1051, in test_check_tag_with_overflow_time
    tag.check()
AssertionError: ObjectFormatException not raised

Ran 92 tests in 0.026s

FAILED (failures=2, skipped=3)
python3 -m unittest dulwich/tests/test_objects.py  0.17s user 0.01s system 97% cpu 0.185 total
```

I'm looking forward to hear your thoughts about it!

> Yep, that's indeed what I meant :)

Cool.
Sometimes, some creepy feeling crawls and makes me doubt my initial understanding.

Cheers,

jelmer:
On Wed, Nov 01, 2017 at 09:45:39AM +0000, Antoine R. Dumont wrote:
> > On my system, UINTMAX_MAX is 18446744073709551615 (from stdint.h), or
> 2**64. That's the same as sys.maxint*2 (since int is signed) in
> Python; could we use that?
> 
> Well, that's where the trouble lies. This is too high (well for
> python3 on my box at least).
I think we mostly need to check that the number fits into an int (rather than
making sure that the resulting value is too large).

> I guess the question might be, where do we draw the line between technicals (c/python runtimes) and
> reality (we are reading stuff from the past)?
The answer to that is very simple: we need to be consistent with C Git;
Dulwich should accept and reject the same values that C git does.

> 2^64 is not even possible for the python runtime to work with (well as
> you mentioned earlier depending on the date library's implementation
> used, i settled on datetime since it's in native python).
It's fine to allow values that are bigger than what fits into a datetime
object. Users of datetime can always catch the OverflowException from datetime.

Cheers,

Jelmer

ardumont:
Ack on your latest consistent response.
I'll adapt accordingly to that limit then (2^64).

In this current implementation though, that somehow defeats the point of me initiating the discussion and proposing a fix. 

Well, not totally defeating it :) but for my case it does ;)

...
(thinking)
...

And now, i remember that this implemented check is not totally matching the current C git check.
That is but the first one.
[There is a secondary check done which i did not grasp at the time and which slipped my mind](https://github.com/git/git/blob/master/date.c#L1214-L1220).

I sense it will answer the part missing for my case.
So i will continue digging this and get back to you on this :)

Thanks for your patience.

Cheers,

ardumont:
> There is a secondary check done which i did not grasp at the time and which slipped my mind.

## TL;DR

`(2^63)-1` is the limit which matches `git fsck`.

## For the curious, here is my rabbit hole path :)

According to the comment linked, they implement a range check to be within the `time_t` structure (which is platform-dependent).

I tried to read the `time.h` header but i'm not so sure of my reading.
I believed that it confirms the 2^64 bits max (well preprocessor
conditional, aliasing, including and all that sorts of things does not
quite help)...

Anyway, I settled [for the documentation](https://www.gnu.org/software/libc/manual/html_node/Simple-Calendar-Time.html), which confirms the platform dependency and hint at a possible size.

Quoting:
```
In the GNU C Library, time_t is equivalent to long int. In other
systems, time_t might be either an integer or floating-point type.
```

According to my current understanding, `long int`.
This would either be a max of 2^32 or 2^64... 

I sense a loop in the Force!
/o\

Anyway, checking again some header file, `/usr/include/limits.h`:
```
/* Minimum and maximum values a `signed long int' can hold.  */

/* Maximum value an `unsigned long int' can hold.  (Minimum is 0.)  */
#  if __WORDSIZE == 64
#   define ULONG_MAX	18446744073709551615UL
#  else
#   define ULONG_MAX	4294967295UL
#  endif
```

Settling on 2^64 as mentioned earlier did not match the reality of my experience with
`git fsck` though (#267 example again which are the tests i programmed as well).

So, I took the problem the other way around.  
With the current dulwich code, i crafted a dummy repository and iterated
over a range of 32-64 exponents (to create commit date time).

Then `git fsck` it when done.  Turns out the commits which are
detected on overflow errors are from 2^63 and beyond.

And then i realized that in `limits.h`, that's the signed long max limit just prior unsigned long ones:
(9223372036854775807L is `(2^63)-1`):

```
/* Minimum and maximum values a `signed long int' can hold.  */
#  if __WORDSIZE == 64
#   define LONG_MAX	9223372036854775807L
#  else
#   define LONG_MAX	2147483647L
#  endif
#  define LONG_MIN	(-LONG_MAX - 1L)
```
So there goes the story \m/.

For reproductibility:
```
$ git init repository
$ cd repository
$ touch README.md ; git add . ; git commit -m 'initial commit'
$ python3
>>> for i in range(32, 65): r.do_commit(b'basic commit - time in seconds at 2^%s' % str(i).encode(), commit_timestamp=(2**i))
$ git fsck
error in commit b425efbaa606f2faa7fb125902aec0e5d8964c8c: badDateOverflow: invalid author/committer line - date causesinteger overflow
error in commit e82b7260e85dec96d3baf4a7c41fe6ad169d68c8: badDateOverflow: invalid author/committer line - date causesinteger overflow
Checking object directories: 100% (256/256), done.
$ git show --format=raw b425efbaa606f2faa7fb125902aec0e5d8964c8c
commit b425efbaa606f2faa7fb125902aec0e5d8964c8c
tree f93e3a1a1525fb5b91020da86e44810c87a2d7bc
parent 325d8cfced8681094edbc91ec46399901d53ab06
author Antoine R. Dumont (@ardumont) <antoine.romain.dumont@gmail.com> 9223372036854775808 +0000
committer Antoine R. Dumont (@ardumont) <antoine.romain.dumont@gmail.com> 9223372036854775808 +0000

    basic commit - time in seconds at 2^63
$ git show --format=raw e82b7260e85dec96d3baf4a7c41fe6ad169d68c8
commit e82b7260e85dec96d3baf4a7c41fe6ad169d68c8
tree f93e3a1a1525fb5b91020da86e44810c87a2d7bc
parent b425efbaa606f2faa7fb125902aec0e5d8964c8c
author Antoine R. Dumont (@ardumont) <antoine.romain.dumont@gmail.com> 18446744073709551616 +0000
committer Antoine R. Dumont (@ardumont) <antoine.romain.dumont@gmail.com> 18446744073709551616 +0000

    basic commit - time in seconds at 2^64
```

Note:
I forgot how painful reading c header could be (i have a headache now ;)

Cheers,

ardumont:
I've rebased to work around `the won't trigger the ci build due to conflict` policy.

Which was the good call since some errors occurred for some python3 versions (about bytes and join syntax).

ardumont:
Hello,

Should i squash commits now?
I have left them for now to demonstrate the iterative steps (if that could help understand).
Not sure you need it though :)

Note that i don't get the errors in the appveyor ci either...

Cheers,

jelmer:
Looks good! It'd be great if you could squash the commits, and I'll merge.

ardumont:
> Looks good! It'd be great if you could squash the commits, and I'll merge.

Sure.
Done (+ rebased to latest master)
Thanks.

Cheers,

jelmer:
Merged, thank you!

Issue #575, opened 2017-11-07
Porcelain add() does not work relative paths, or a directory on Windows


MattSable:
porcelain.add() only works if the filepaths are absolute paths. You cannot specify a directory either, you must pass a list of files. I would expect

`porcelain.add(repo, repo.path)`

should stage all files in the repo. Instead, nothing gets staged.

jelmer:
On Tue, Nov 07, 2017 at 06:59:51PM +0000, Matt wrote:
> porcelain.add() only works if the filepaths are absolute paths. You cannot specify a directory either, you must pass a list of files. I would expect
> 
> `porcelain.add(repo, repo.path)`
> 
> should stage all files in the repo. Instead, nothing gets staged.
This is very likely Windows-specific. Unfortunately I don't have access to
Windows, but pull requests to fix this on Windows would be great.

derSeddy:
This has nothing to do with windows.   
A quick look into the code shows that directories are simply not supported and that file paths are assumed to be relative to the current python working dir.
This code will fail if working dir is different than repo dir:
```
repo = porcelain.open_repo(repo_dir)
status = porcelain.status(repo)
porcelain.add(repo, status.unstaged + status.untracked)
```

Issue #559, opened 2017-08-26
move_in_pack() can fail on Windows if the target already exists


earlchew:
```
======================================================================
ERROR: test_repack_existing (dulwich.tests.test_object_store.DiskObjectStoreTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\projects\dulwich\dulwich\tests\test_object_store.py", line 311, in test_repack_existing
    self.assertEqual(2, self.store.repack())
  File "c:\projects\dulwich\dulwich\object_store.py", line 367, in repack
    consolidated = self.add_objects(objects)
  File "c:\projects\dulwich\dulwich\object_store.py", line 441, in add_objects
    return commit()
  File "c:\projects\dulwich\dulwich\object_store.py", line 701, in commit
    return self.move_in_pack(path)
  File "c:\projects\dulwich\dulwich\object_store.py", line 681, in move_in_pack
    os.rename(path, basename + ".pack")
WindowsError: [Error 183] Cannot create a file when that file already exists
```

earlchew:
This is a known issue: https://stackoverflow.com/questions/8107352

jelmer:
There's a workaround in dulwich.file._fancy_rename to work around this.

One option would be to just avoid writing the file if it already exists - it's supposed to have the same contents as the new file.

PR #579, opened 2017-11-10
Improve handling of comment characters in Git configuration files

dandersson:
Comment characters (`#` , `;`) that are within balanced quotes should not discard the rest of the line:

    [branch "my#branch"]

is a valid section for a branch named `my#branch` ([`#` is a valid character in a branch name](https://git-scm.com/docs/git-check-ref-format)). Previously this normalized to:

    [branch "my

which raised an exception for finding an invalid section header.

Adhering strictly to the exact allowed comment format might require more care, but this should at least fix some problems I have had with real-life repositories.

(I tried to keep with the surrounding code style even though [the contribution guide](https://github.com/jelmer/dulwich/blob/master/CONTRIBUTING.md) preferred `'` as string delimiter.)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,10 @@
   * Make `dulwich.archive` set the gzip header file modification time so that
     archives created from the same Git tree are always identical.
     (#577, Jonas Haag)
+
+  * Allow comment characters (#, ;) within configuration file strings
+    (Daniel Andersson, #579)
+
 0.18.6	2017-11-11
 
  BUG FIXES
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -262,8 +262,16 @@ def _check_section_name(name):
 
 
 def _strip_comments(line):
-    line = line.split(b"#")[0]
-    line = line.split(b";")[0]
+    comment_bytes = {ord(b"#"), ord(b";")}
+    quote = ord(b'"')
+    string_open = False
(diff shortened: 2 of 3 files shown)


dandersson:
I see that Python 3 builds are failing, though I cannot obviously see why. I am at work with Windows+Python 2.7 -- will try to check the issue at home during the weekend.

jelmer:
Thanks!

dandersson:
The issue with my previous attempt tested only on Python 2.7 was that the input line came as a `bytes` object, and iterating over a `bytes` object in Python 3 yields `int`s, not new `bytes` objects of length 1. Normalizing the input line to a `bytearray` for iteration and comparing ASCII code points instead of "characters" should work in both versions.

---

CI builds currently fail, but:

* Travis fails seem to be due to code style issues from 64e084e77. I will make a separate merge request to fix these, unless someone else is quicker.
* AppVeyor fails for

    * `dulwich.tests.test_object_store.DiskObjectStoreTests.test_repack_existing`, which seems to be #559.
    * `dulwich.tests.test_porcelain.AddTests.test_add_not_in_repo`, which at least sounds related to #575.

jelmer:
Yep, those two bugs were filed specifically for those issues on Windows. I don't have access to a Windows machine so any help getting them resolved would be great :)

Issue #582, opened 2017-11-13
Empty/corrupted file in repository breaks reading object api


ardumont:
Hello,

In the same logic as #567, `git fsck` raises errors when corrupted file exists in the repository.
But dulwich is not prepared for it and breaks on it with a TypeError.

Are you interested to check against this in the same manner as before, that is try and raise an ObjectFormatException when the case raises?

### Reproductibility

Given a repository with corrupted data:
```
cd ..
$ cd sortix-gitorious-wiki.git
$ git fsck
error: object file ./objects/99/ea0c61b44c16ddd9956f8bb58749d707d14b47 is empty
error: object file ./objects/99/ea0c61b44c16ddd9956f8bb58749d707d14b47 is empty
fatal: loose object 99ea0c61b44c16ddd9956f8bb58749d707d14b47 (stored in ./objects/99/ea0c61b44c16ddd9956f8bb58749d707d14b47) is corrupt
```

For that same repository, checking with dulwich the object `99ea0c61b44c16ddd9956f8bb58749d707d14b47`:

```sh
$ python3
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
[GCC 6.3.0 20170118] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dulwich.repo import Repo
>>> r = Repo('.')
>>> r[b'99ea0c61b44c16ddd9956f8bb58749d707d14b47']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3/dist-packages/dulwich/repo.py", line 472, in __getitem__
    return self.object_store[name]
  File "/usr/lib/python3/dist-packages/dulwich/object_store.py", line 118, in __getitem__
    type_num, uncomp = self.get_raw(sha)
  File "/usr/lib/python3/dist-packages/dulwich/object_store.py", line 415, in get_raw
    ret = self._get_loose_object(hexsha)
  File "/usr/lib/python3/dist-packages/dulwich/object_store.py", line 565, in _get_loose_object
    return ShaFile.from_path(path)
  File "/usr/lib/python3/dist-packages/dulwich/objects.py", line 389, in from_path
    return cls.from_file(f)
  File "/usr/lib/python3/dist-packages/dulwich/objects.py", line 395, in from_file
    obj = cls._parse_file(f)
  File "/usr/lib/python3/dist-packages/dulwich/objects.py", line 365, in _parse_file
    if cls._is_legacy_object(map):
[...]

PR #583, opened 2017-11-13
WIP - Raise exception when disk object corruption detected

ardumont:
I still have implementation questions pending in the related issue.

I prefer opening the PR now to show the current state so that we can
discuss freely the implementation.

Related #582

(rebased as i was not up-to-date with upstream/master)

### DIFF
--- a/dulwich/errors.py
+++ b/dulwich/errors.py
@@ -164,6 +164,13 @@ class ObjectFormatException(FileFormatException):
     """Indicates an error parsing an object."""
 
 
+class EmptyFileException(FileFormatException):
+    """Indicates an empty file instead of the object's disk
+       representation.
+
+    """
+
+
 class NoIndexPresent(Exception):
     """No index is present."""
 
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -25,6 +25,7 @@
 from io import BytesIO
 import errno
 from itertools import chain
+import logging
 import os
 import stat
 import sys
@@ -456,6 +457,7 @@ class DiskObjectStore(PackBasedObjectStore):
         self._pack_cache_time = 0
         self._pack_cache = {}
         self._alternates = None
+        self.log = logging.getLogger('dulwich.object_store.DiskObjectStore')
 
     def __repr__(self):
         return "<%s(%r)>" % (self.__class__.__name__, self.path)
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -208,7 +208,7 @@ def check_time(time_seconds):
 
 
 def git_line(*items):
-    """Formats items into a space sepreated line."""
+    """Formats items into a space separated line."""
     return b' '.join(items) + b'\n'
 
 
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -38,6 +38,7 @@ from dulwich.errors import (
     NotTagError,
     NotTreeError,
     ObjectFormatException,
+    EmptyFileException,
     )
 from dulwich.file import GitFile
 
@@ -362,6 +363,9 @@ class ShaFile(object):
     @classmethod
     def _parse_file(cls, f):
         map = f.read()
+        if not map:
+            raise EmptyFileException('Corrupted empty file detected')
+
         if cls._is_legacy_object(map):
             obj = cls._parse_legacy_object_header(map, f)
             obj._parse_legacy_object(map)
--- a/dulwich/tests/test_object_store.py
+++ b/dulwich/tests/test_object_store.py
@@ -33,6 +33,7 @@ from dulwich.index import (
     )
 from dulwich.errors import (
     NotTreeError,
+    EmptyFileException,
     )
 from dulwich.objects import (
     sha_to_hex,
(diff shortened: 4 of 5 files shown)


jelmer:
Having check report these and not raise an exception seems reasonable to me. Similarly, when iterating over all objects in the repo, it seems reasonable to warn and skip problematic objects.

However, I don't think we should silently ignore corrupted files, that just results in more confusion. If the user requests an object that is unparseable, that should just raise an exception like it did before (Perhaps a slightly different exception than ValueError though).

That would also be more in line with C git.

ardumont:
I did not reply as i was trying to improve the current patch and go in the direction you proposed.
I'm not feeling that i did good though.

> Having check report these and not raise an exception seems reasonable to me.

Ok. 

I don't see where i can do that check in the objects though (since those are not path format).
What am i missing?

My understanding must be too much tied with the corruption case i have in #582 (empty file).
The only corruption i see is that of the empty file (which seems to be only possible in the loose objects pool).

>  Similarly, when iterating over **all** objects in the repo, it seems reasonable to warn and skip problematic objects.

Ok. I don't see where i can treat this error for all objects now.
I'm plugged on the loose objects since i only see path objects manipulated there.

> However, I don't think we should silently ignore corrupted files, that just results in more confusion. 

I agree with that.

It's just that i did not see clearly where i could have put this log behavior :)
So, i finally just added log to the DiskObjectStore and log from there.
I'm not sure that that's what you are suggesting though.

> If the user requests an object that is unparseable, that should just raise an exception like it did before (Perhaps a slightly different exception than ValueError though).

I did go with a new CorruptedFormatException inheriting from dulwich.errors.FileFormatException.

Looking forward to hear your thoughts on this.

Thanks for your help.

Cheers,

dulwich/errors.py
```diff
@@ -164,6 +164,10 @@ class ObjectFormatException(FileFormatException):
     """Indicates an error parsing an object."""
 
 
+class CorruptedFormatException(FileFormatException):
+    """Indicates a corruption in the object's disk representation."""
```
jelmer: Any FileFormatException is a sign of an invalid disk representation; perhaps you mean something like EmpyFileException?

dulwich/object_store.py
```diff
@@ -557,14 +560,19 @@ def _iter_loose_objects(self):
...
+                    continue
+                yield fpath
 
     def _get_loose_object(self, sha):
         path = self._get_shafile_path(sha)
         try:
-            if os.lstat(path).st_size == 0:
-                return None
             return ShaFile.from_path(path)
+        except CorruptedFormatException:
```
jelmer: You'd probably want to let this bubble up to the caller. If they explicitly asked for this object (rather than e.g. as part of iterating all the objects in the repo), they'd want to know about corruption.

dulwich/tests/test_object_store.py
```diff
@@ -346,13 +346,19 @@ def test_alternates(self):
     def test_corrupted_loose_object_not_returned(self):
         """Check corrupted sha1 file are not returned"""
         self.store.add_object(testobject)
-        self.assertEqual((Blob.type_num, b'yummy data'),
-                         self.store.get_raw(testobject.id))
+        assert ((Blob.type_num, b'yummy data') ==
```
jelmer: Please use standard unittest mehotds; e.g. self.assertXXX rather than assert.

dulwich/objects.py
```diff
@@ -383,8 +384,23 @@ def _serialize(self):
         raise NotImplementedError(self._serialize)
 
     @classmethod
+    def _check_is_corrupted(cls, path):
```
jelmer: for performance reasons, please raise exceptions about formatting issues where they are noticed, rather than checking for all corruption up front. E.g. in ShaFile._parse_file, raise an exception if the file was empty.

dulwich/tests/test_object_store.py
```diff
@@ -343,6 +343,23 @@ def test_alternates(self):
...
         self.assertEqual(b2, store[b2.id])
 
+    def test_corrupted_loose_object_not_returned(self):
+        """Check corrupted sha1 file are not returned"""
+        self.store.add_object(testobject)
+        assert ((Blob.type_num, b'yummy data') ==
+                self.store.get_raw(testobject.id))
+
+        assert self.store.contains_loose(testobject.id) is True
+        assert self.store._get_loose_object(testobject.id) is not None
```
ardumont: > Please use standard unittest mehotds; e.g. self.assertXXX rather than assert.

(I don't see that comment anymore but still, explaining my thought here.)

Those are state assertions needed for the setup phase of the test.
If i'm wrong here, i want a different error message to raise than the standard `self.assertXXX`'s failure message below.

dulwich/object_store.py
```diff
@@ -557,12 +560,19 @@ def _iter_loose_objects(self):
...
+                yield fpath
 
     def _get_loose_object(self, sha):
         path = self._get_shafile_path(sha)
         try:
             return ShaFile.from_path(path)
+        except CorruptedFormatException:
+            self.log.warn('Object with id %s is corrupted! Skipping.' % (
+                sha.decode('ascii'), ))
+            return None
```
ardumont: > You'd probably want to let this bubble up to the caller. If they explicitly asked for this object (rather than e.g. as part of iterating all the objects in the repo), they'd want to know about corruption.

(Same here, i don't see that comment anymore.)

It's the same idea you already explained to me, but now i think, i fully understand.

ardumont:
(I'll rebase and squash commits when we deem this ready)

dulwich/tests/test_object_store.py
jelmer: I think you do want the same kinds of assertions, e.g. self.assertEqual, self.assertIs. The downsides of assert are that it doesn't return the values on either side of the assert statement, and it gets ignored if you run Python with optimizations enabled.

jelmer:
Thanks, looks good to me apart from the asserts.

dulwich/tests/test_object_store.py
ardumont: Ok. Will adapt accordingly then.

ardumont:
Rebased and squashed commits.

I kept the one fix docstring commit aside.
This is not the same perimeter so i kept them separated (for ease of potential future revert).

Tell me if you prefer i squash everything or if you prefer another PR (i was lazy on that and thought it was a bit much...).

Cheers,

PR #589, opened 2017-11-29
By default file.encoding=None, so .encoding() fails. 

darkdragon-001:
See https://bugs.python.org/issue32163.


jelmer:
Merged, thanks!

PR #600, opened 2018-01-06
Make config keys (sections and names) case insensitive

### LINKED ISSUE #599: Gitignore ignores core.excludesfile setting
This is a fairly minor one, I worked around it by changing where I store my global ignores. I had git's core.excludesfile set to `~/.gitignore_global` but the gitignore handling code was hitting `~/.config/git/ignore` (which didn't exist). I have moved my gitignore to this path and everything works fine. 

Having had a poke into the code it seems like it's looking for `${XDG_CONFIG_HOME:=$HOME/.config}/git/ignore` in the running environment, rather than getting the value of `git config --get core.excludesfile`. 

With a little digging it seems like XDG_CONFIG_HOME is fairly standard on a number of linux distributions (at least debian and arch) but is certainly not a thing on OSX or Windows, and doesn't look to be standard on at least one current Ubuntu LTS release.

alistair-broomhead:
This is done to match the behaviour of git itself.

The main change here is to add the lower_key function and the CaseInsensitiveDict class.

lower_key coerces keys that are used in DictConfig to make sure all keys consist only of lower-cased strings.

CaseInsensitiveDict has altered __getitem__ and __setitem__ methods to call lower_key on the key for both get and set operations.

CaseInsensitiveDict.make takes care of the creation logic found in ConfigDict.__init__ as well as converting any passed-in dicts to CaseInsensitiveDict.

CaseInsensitiveDict.get and CaseInsensitiveDict.setdefault allow the default value returned to be a CaseInsensitiveDict without repeating it everywhere.

ConfigDict.check_section_and_name just moves some logic that was previously repeated.

Fixes #599

### DIFF
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -30,6 +30,7 @@ import errno
 import os
 
 from collections import (
+    Iterable,
     OrderedDict,
     MutableMapping,
     )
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the fix. I'm currently traveling but will try to review your merge proposal some time in the next week.

On 7 January 2018 00:54:20 GMT+08:00, Alistair Broomhead <notifications@github.com> wrote:
>This is done to match the behaviour of git itself.
>
>The main change here is to add the lower_key function and the
>CaseInsensitiveDict class.
>
>lower_key coerces keys that are used in DictConfig to make sure all
>keys consist only of lower-cased strings.
>
>CaseInsensitiveDict has altered __getitem__ and __setitem__ methods to
>call lower_key on the key for both get and set operations.
>
>CaseInsensitiveDict.make takes care of the creation logic found in
>ConfigDict.__init__ as well as converting any passed-in dicts to
>CaseInsensitiveDict.
>
>CaseInsensitiveDict.get and CaseInsensitiveDict.setdefault allow the
>default value returned to be a CaseInsensitiveDict without repeating it
>everywhere.
>
>ConfigDict.check_section_and_name just moves some logic that was
>previously repeated.
>
>Fixes #599
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/jelmer/dulwich/pull/600
>
>-- Commit Summary --
>
>  * Make config keys (sections and names) case insensitive
>
>-- File Changes --
>
>    M dulwich/config.py (91)
>    M dulwich/tests/test_config.py (7)
>
>-- Patch Links --
>
>https://github.com/jelmer/dulwich/pull/600.patch
>https://github.com/jelmer/dulwich/pull/600.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/jelmer/dulwich/pull/600

-- 
https://www.jelmer.uk/

alistair-broomhead:
Awesome, thanks for the speedy response!

PR #601, opened 2018-01-06
Prevent clean files from showing up as untracked

### LINKED ISSUE #598: Git status shows all files as untracked
This is observed using a fresh pipenv with only dulwich==0.18.6 installed using python3.6 from pyenv on OSX

This is obviously not consistent with the way that git itself describes the repository state. Is this by design or a known issue? Are there any known work-arounds? If not I'm happy to give any spare cycles to have a dig into improving things, even if that's just by contributing docs.

```
bash-3.2$ > ls -l && \
git status && \
python3 -c 'import dulwich.porcelain; print(dulwich.porcelain.status("."))'
total 16
-rw-r--r--  1 al  staff   166  4 Jan 17:50 Pipfile
-rw-r--r--  1 al  staff  1264  4 Jan 17:50 Pipfile.lock
-rw-r--r--  1 al  staff     0  4 Jan 17:46 bar
-rw-r--r--  1 al  staff     0  4 Jan 17:46 foo
On branch master
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	new file:   bar

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	Pipfile
	Pipfile.lock

GitStatus(staged={'add': [b'bar'], 'delete': [], 'modify': []}, unstaged=[], untracked=['bar', 'foo', 'Pipfile', 'Pipfile.lock'])
```

alistair-broomhead:
Looking at the code in porcelain.path_to_tree_path, this looks like there was an attempt at this in 85b5383fb816fece4a36faa538f6477703595139, however relpath is non-destructive, and has no effect unless the result is assigned to something. If this were assigned when path is relative, then relpath would assume path was at root, and so we must check that the path is absolute before calling relpath.

Fixes #598

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -168,11 +168,12 @@ def open_repo_closing(path_or_repo):
 def path_to_tree_path(repopath, path):
     """Convert a path to a path usable in e.g. an index.
 
-    :param repo: Repository
+    :param repopath: Repository
     :param path: A path
     :return: A path formatted for use in e.g. an index
     """
-    os.path.relpath(path, repopath)
+    if os.path.isabs(path):
+        path = os.path.relpath(path, repopath)
     if os.path.sep != '/':
         path = path.replace(os.path.sep, '/')
     return path.encode(sys.getfilesystemencoding())
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -796,17 +796,27 @@ class StatusTests(PorcelainTestCase):
             results.staged)
         self.assertEqual([], results.unstaged)
 
+    def commit_file(self, name, content='original', author=b'', committer=b''):
+        fullpath = os.path.join(self.repo.path, name)
+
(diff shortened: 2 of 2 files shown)


alistair-broomhead:
Rebased on top of current master to check for conflicts (there were none)

dulwich/porcelain.py
```diff
@@ -168,11 +168,12 @@ def open_repo_closing(path_or_repo):
 def path_to_tree_path(repopath, path):
     """Convert a path to a path usable in e.g. an index.
 
-    :param repo: Repository
+    :param repopath: Repository
     :param path: A path
     :return: A path formatted for use in e.g. an index
     """
-    os.path.relpath(path, repopath)
+    if os.path.isabs(path):
```
jelmer: "path" will be relative to os.getcwd(), so I think we should unconditionally call relpath since path_to_tree_path should return the relative path to the repository.

jelmer:
See also #601

jelmer:
This has already been fixed.

Issue #586, opened 2017-11-17
"Broken pipe" pushing packs over 128 kB to GitHub


dandersson:
* Happens with both Python 2.7.14 and Python 3.6.3 (Debian Sid)
* `dulwich` 0.18.6 (with C extensions)

Using `dulwich`, I started noticing issues pushing files above a certain size to a local GitLab instance using Python 2.7.13 on Windows 7, where I initially suspected the corporate proxy was interfering since I could not seem to push packs larger than ~256 kB. However, pushing to GitHub from my home computer (Debian Sid) seems to indicate similar problems, now for a max size of ~128 kB. In all cases, issuing `git push` with the standard Git binary in the repository works without issue.

In practice I run:

```python
dulwich.porcelain.push('/path/to/demo-github', 'https://dandersson:api-token@github.com/dandersson/demo-github.git', refspecs=b'proxy-test')
```

(see <https://github.com/dandersson/dulwich-test-23> for the full test script).

When pushing a commit in the `proxy-test` branch with a single binary file of size 1 MB to GitHub, I get:

    DEBUG:__main__:CLI args: Namespace(branch='proxy-test', config=<_io.TextIOWrapper name='.credentials' mode='r' encoding='UTF-8'>, debug=True, repository='/home/daniel/src/demo-github')
    DEBUG:__main__:Local branches:  {b'git-logs', b'proxy-test', b'master'}
    DEBUG:__main__:Remote branches: {b'origin/git-logs', b'origin/HEAD', b'origin/master', b'origin/proxy-test'}
    DEBUG:__main__:{b'HEAD': b'89ccc9fe76e83dbe423745f77e47abe7d1b3b9f5', b'refs/heads/git-logs': b'afe0b4e2c6e01162eee2f72ee5f5f030d47c53df', b'refs/heads/master': b'89ccc9fe76e83dbe423745f77e47abe7d1b3b9f5', b'refs/heads/proxy-test': b'271900c378411bef9779f01f4a2976c0ec2625f2'}
    Traceback (most recent call last):
      File "/usr/lib/python3.6/urllib/request.py", line 1318, in do_open
        encode_chunked=req.has_header('Transfer-encoding'))
      File "/usr/lib/python3.6/http/client.py", line 1239, in request
        self._send_request(method, url, body, headers, encode_chunked)
      File "/usr/lib/python3.6/http/client.py", line 1285, in _send_request
        self.endheaders(body, encode_chunked=encode_chunked)
[...]

PR #611, opened 2018-01-26
`urllib3` migration

dandersson:
See issue #586. This is an attempt to fix issues with transferring "large" (O(100 kB)) pack sizes in certain environments.

API changes (i.e. changes to "non-underscored" things):

* `dulwich.client.default_urllib2_opener` has been replaced by `dulwich.client.default_urllib3_manager` with otherwise similar input arguments, returning a `urllib3.PoolManager` or `urllib3.ProxyManager`, depending on whether a proxy configuration is given.
* The `opener` argument to `dulwich.client.HttpGitClient` that took a `urllib2` opener instance has been replaced by a `pool_manager` argument that takes a `urllib3` pool manager instance.

I guess this should warrant a major version bump.

---

Initially tested with `urllib3` version 1.21, but anything later in the 1.x series should work.

I added a `requirements.txt` as a common convention for local setup with a known working version. I pinned `urllib3` to its latest available version at the time.

---

I have only tested in Windows (through MSYS2) for Python 2 and Python 3, so there might be surprises lurking, but I wanted to get the pull request up early for others to review and try. I have not yet tested proxy configurations apart from making unit tests run.

---

I have not implemented certificate verification for `urllib3`, which makes the current code issue a warning when performing HTTPS requests: https://urllib3.readthedocs.io/en/latest/user-guide.html#ssl . From what I can tell it should just be to change the dependency from `urllib3` to `urllib3[secure]` and add parameters when initializing the `PoolManager`/`ProxyManager`. I will take a look at this the coming week.


jelmer:
Thanks, I'll take a look at this over the next couple of days.

The certificate verification issue needs to be resolved before this can be merged, otherwise that's a major regression.

dandersson:
> The certificate verification issue needs to be resolved before this can be merged, otherwise that's a major regression.

Agreed. It has been implemented now.

For Python versions older than 2.7.9, there is an [extra workaround needed to enable SSL verification due to changes made in the standard library](https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification-in-python-2). In the current commit I include this fix since 2.7 is a supported version for Dulwich, but if you feel that "2.7 support" means supporting the latest patch version of the 2.7 branch, I can remove it.

I believe that the `urllib2` solution does not verify HTTPS requests for versions older than 2.7.9 either for the very same reasons, so it would not be a regression against `master` to remove the fix.

dandersson:
Travis CI segfaults for Python 3.7-dev. It seems to be repeatable, but I am not sure why.

I see that 3.7-dev runs were [recently disabled in `master` due to `gevent`](https://github.com/dulwich/dulwich/commit/fa5d8319465ec19d905be999fe42e40fe18fdb13), but my branch includes that commit and still runs it in some respect.

[From the log](https://travis-ci.org/dulwich/dulwich/jobs/334702267):

    $ python -m coverage run -p --source=dulwich -m unittest dulwich.tests.test_suite
    /home/travis/.travis/job_stages: line 57:  3392 Segmentation fault      (core dumped) python -m coverage run -p --source=dulwich -m unittest dulwich.tests.test_suite

dandersson:
I have tried running against GitHub now as well, and I can successfully push a 256 kB file using Dulwich built from this branch, while the current Dulwich `master` fails as described in #586 (tested on Windows, Python 3).

I have also tried it with a repository that has the `http.proxy` Git configuration option set, and proxy settings were applied correctly.

jelmer:
Great, thanks! I'll try to test and merge this during the next couple of days.

PR #614, opened 2018-02-17
Make porcelain.fetch() actually do the fetch into the local checkout

minddrive:
The previous version of porcelain.fetch() simply returned the remotes, but didn't update the on-disk information.  This should correct it.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -116,7 +116,10 @@ from dulwich.protocol import (
     Protocol,
     ZERO_SHA,
     )
-from dulwich.refs import ANNOTATED_TAG_SUFFIX
+from dulwich.refs import (
+    ANNOTATED_TAG_SUFFIX,
+    strip_peeled_refs,
+)
 from dulwich.repo import (BaseRepo, Repo)
 from dulwich.server import (
     FileSystemBackend,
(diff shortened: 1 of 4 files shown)


dulwich/porcelain.py
```diff
@@ -1045,11 +1048,12 @@ def branch_list(repo):
         return r.refs.keys(base=b"refs/heads/")
 
 
-def fetch(repo, remote_location, outstream=sys.stdout,
+def fetch(repo, remote_name, remote_location, outstream=sys.stdout,
```
jelmer: This is an API change, could you perhaps make remote_name an optional argument that defaults to None (in which case the original behaviour is preserved)?

dulwich/refs.py
```diff
@@ -882,3 +882,9 @@ def write_info_refs(refs, store):
 
 def is_local_branch(x):
     return x.startswith(b'refs/heads/')
+
+
+def strip_peeled_refs(refs):
+    """Remove all peeled refs"""
+    return {ref: sha for (ref, sha) in refs.items()
+            if not ref.endswith(b"^{}")}
```
jelmer: Please use ANNOTATED_TAG_SUFFIX rather than b"^{}"

dulwich/refs.py
```diff
@@ -882,3 +882,9 @@ def write_info_refs(refs, store):
 
 def is_local_branch(x):
     return x.startswith(b'refs/heads/')
+
+
+def strip_peeled_refs(refs):
```
jelmer: Please add a basic test for strip_peeled_refs.

dulwich/refs.py
minddrive: Didn't realize this existed, changing...

dulwich/porcelain.py
minddrive: No problem; my upcoming change will actually set it to "b'origin'", since that avoids an unnecessary conditional in the code, if that's okay?

dulwich/refs.py
minddrive: Should I create a new test class for this test, or is there an existing one where this would be appropriate?

minddrive: Test added.

Issue #364, opened 2015-08-18
tests for paramiko SSH implementation


jelmer:
There are currently no tests for the Paramiko SSH support in Dulwich, leading to bugs like #363. We should add some.

jelmer:
The Paramiko SSH Vendor has been moved to dulwich/contrib, making this issue somewhat less important.

jelmer:
Replying to the comments by @friz-zy in https://github.com/dulwich/dulwich/pull/615:

There is currently no end-to-end test for SSH with a real SSH server, since this is tricky to do (e.g. involves depending on openssh installation).

Ideally, tests for paramiko would be end to end tests that simply used the paramiko SSH server implementation; depending on external network connectivity is a problem.

PR #615, opened 2018-02-23
Add ability to use password and keyfile ssh options

Friz-zy:
Update for `SSHVendor`, `SubprocessSSHVendor`, `SSHGitClient` and `ParamikoSSHVendor` classes:

`SubprocessSSHVendor` now use `putty.exe` on `win32` platform and `sshpass` or `ssh` on other depend on the options.

`ParamikoSSHVendor` class currently most useful and can use all `paramiko.client.connect` options, but this options should be given on `__init__` call or with `run_command`. Also `SSHGitClient` can't pass this options except new `password` or `key_filename`

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1070,15 +1070,18 @@ default_local_git_client_cls = LocalGitClient
 class SSHVendor(object):
     """A client side SSH implementation."""
 
-    def connect_ssh(self, host, command, username=None, port=None):
+    def connect_ssh(self, host, command, username=None, port=None,
+                    password=None, key_filename=None):
         # This function was deprecated in 0.9.1
         import warnings
         warnings.warn(
             "SSHVendor.connect_ssh has been renamed to SSHVendor.run_command",
             DeprecationWarning)
-        return self.run_command(host, command, username=username, port=port)
+        return self.run_command(host, command, username=username, port=port,
+                                password=password, key_filename=key_filename)
 
-    def run_command(self, host, command, username=None, port=None):
+    def run_command(self, host, command, username=None, port=None,
+                    password=None, key_filename=None):
         """Connect to an SSH server.
 
         Run a command remotely and return a file-like object for interaction
@@ -1088,6 +1091,8 @@ class SSHVendor(object):
         :param command: Command to run (as argv array)
         :param username: Optional ame of user to log in as
         :param port: Optional SSH port to use
+        :param password: Optional ssh password for login
+        :param key_filename: Optional path to private keyfile
         """
         raise NotImplementedError(self.run_command)
 
@@ -1102,16 +1107,49 @@ class StrangeHostname(Exception):
 class SubprocessSSHVendor(SSHVendor):
     """SSH vendor that shells out to the local 'ssh' command."""
 
-    def run_command(self, host, command, username=None, port=None):
-        # FIXME: This has no way to deal with passwords..
-        args = ['ssh', '-x']
-        if port is not None:
-            args.extend(['-p', str(port)])
-        if username is not None:
+    def run_command(self, host, command, username=None, port=None,
+                    password=None, key_filename=None):
+        if password and key_filename:
+            raise NotImplementedError(
+                "You can't set passphrase for ssh key with SubprocessSSHVendor, " \
+                "use ParamikoSSHVendor instead"
+            )
+
+        if sys.platform == 'win32':
+            args = ['putty.exe', '-ssh']
+            if password:
+                args.extend(['-pw', str(password)])
+        else:
+            if password:
+                try:
+                    subprocess.call(["sshpass"])
+                except OSError as e:
+                    import warnings
+                    warnings.warn(
+                        "You need sshpass for using password option, " \
+                        "you can use ssh-add or private key without password " \
+                        "or ParamikoSSHVendor"
+                    )
+                    raise e
+                args = ['sshpass', '-p', str(password)]
+            else:
+                args = ['ssh', '-x']
+
+        if port:
+            if sys.platform == 'win32':
+                args.extend(['-P', str(port)])
+            else:
+                args.extend(['-p', str(port)])
+
+        if key_filename:
+            args.extend(['-i', str(key_filename)])
+
+        if username:
             host = '%s@%s' % (username, host)
         if host.startswith('-'):
             raise StrangeHostname(hostname=host)
         args.append(host)
+
         proc = subprocess.Popen(args + [command], bufsize=0,
                                 stdin=subprocess.PIPE,
                                 stdout=subprocess.PIPE)
@@ -1134,10 +1172,12 @@ get_ssh_vendor = SubprocessSSHVendor
 class SSHGitClient(TraditionalGitClient):
 
     def __init__(self, host, port=None, username=None, vendor=None,
-                 config=None, **kwargs):
+                 config=None, password=None, key_filename=None, **kwargs):
         self.host = host
         self.port = port
         self.username = username
+        self.password = password
+        self.key_filename = key_filename
         super(SSHGitClient, self).__init__(**kwargs)
         self.alternative_paths = {}
         if vendor is not None:
@@ -1175,7 +1215,8 @@ class SSHGitClient(TraditionalGitClient):
         argv = (self._get_cmd_path(cmd).decode(self._remote_path_encoding) +
                 " '" + path + "'")
         con = self.ssh_vendor.run_command(
-            self.host, argv, port=self.port, username=self.username)
+            self.host, argv, port=self.port, username=self.username,
+            password=self.password, key_filename=self.key_filename)
         return (Protocol(con.read, con.write, con.close,
                          report_activity=self._report_activity),
                 con.can_read)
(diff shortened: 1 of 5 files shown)


Friz-zy:
```
$ make style
flake8 --exclude=build,.git,build-pypy,.tox
./dulwich/client.py:1114:80: E501 line too long (83 > 79 characters)
./dulwich/client.py:1114:83: E502 the backslash is redundant between brackets
./dulwich/client.py:1129:72: E502 the backslash is redundant between brackets
./dulwich/client.py:1130:80: E501 line too long (80 > 79 characters)
./dulwich/client.py:1130:80: E502 the backslash is redundant between brackets
./dulwich/contrib/paramiko_vendor.py:134:13: E265 block comment should start with '# '
./dulwich/contrib/paramiko_vendor.py:135:13: E265 block comment should start with '# '
./dulwich/contrib/paramiko_vendor.py:136:13: E265 block comment should start with '# '
./dulwich/tests/test_client.py:635:1: W191 indentation contains tabs
./dulwich/tests/test_client.py:635:1: E101 indentation contains mixed spaces and tabs
./dulwich/tests/test_client.py:635:7: E128 continuation line under-indented for visual indent
./dulwich/tests/test_client.py:636:1: E101 indentation contains mixed spaces and tabs
make: *** [style] Error 1
The command "make style" exited with 2.
```

I fixed `./dulwich/contrib/paramiko_vendor.py:136` and `./dulwich/tests/test_client.py:635` but maybe you can suggest me a good way to leave long messages?

dulwich/client.py
```diff
@@ -1102,16 +1107,50 @@ def __init__(self, hostname):
...
+            )
+
+        if sys.platform == 'win32':
+            args = ['putty.exe', '-ssh']
+            if password:
+                args.extend(['-pw', str(password)])
+        else:
+            if password:
+                try:
+                    subprocess.call(["sshpass"])
```
jelmer: I'm uncomfortable with supporting sshpass like this; I'm sure there are some legitimate uses for it, but on e.g. a multi-user system this will cause the password to show up in the process list.

dulwich/client.py
```diff
@@ -1102,16 +1107,50 @@ def __init__(self, hostname):
...
+                        "or private key without password "
+                        "or ParamikoSSHVendor"
+                    )
+                    raise e
+                args = ['sshpass', '-p', str(password)]
+            else:
+                args = ['ssh', '-x']
+
+        if port:
+            if sys.platform == 'win32':
```
jelmer: This is assuming that the only ssh binary on Windows is putty, which isn't necessarily true. Putty is also available on *nix. Would it perhaps be possible to create a PuttySSHVendor class?

Friz-zy: Problem that there is no good multi platform way for obtain the same with just only subprocess. If I right remember, for sending password directly into stdin of ssh we should use fork or threads or asinc solution and most of all default solutions do it. We can use `pexpect` for it, but again, on windows this solution require a [special way](https://pexpect.readthedocs.io/en/stable/overview.html#windows) and it possibly would have a conflict with current `dulwitch` wrapper

In my opinion `paramiko` variant would be the best option for security and amount of features in compare with subprocess variant. But it you don't want to support this, and user already have `sshpass` (or `putty` in case ow windows) and this user don't want to use the key auth or preconfigured ssh config file instead - security is not a problem of `dulwitch` itself...

dulwich/client.py
Friz-zy: Problem that there is no good multi platform way for obtain the same with just only subprocess. If I right remember, for sending password directly into stdin of ssh we should use fork or threads or asinc solution and most of all default solutions do it. We can use `pexpect` for it, but again, on windows this solution require a [special way](https://pexpect.readthedocs.io/en/stable/overview.html#windows) and it possibly would have a conflict with current `dulwitch` wrapper

In my opinion `paramiko` variant would be the best option for security and amount of features in compare with subprocess variant. But it you don't want to support this, and user already have `sshpass` (or `putty` in case ow windows) and this user don't want to use the key auth or preconfigured ssh config file instead - security is not a problem of `dulwitch` itself...

dulwich/client.py
Friz-zy: We can, but then we should add the same check of platform into `get_transport_and_path` for using it by default on windows or leave it for user choice. But this would add more complexity only

Again, in my opinion `paramiko` client would be a better multiplatform solution for all cases...

dulwich/client.py
jelmer: see https://github.com/dulwich/dulwich/issues/364. The paramiko module doesn't have any tests and my experience in another project is that its API isn't super stable, which is why I moved it to contrib.

If somebody were to contribute end-to-end tests for paramiko, i.e. tests that setup a SSH server and verify that running dulwich using paramiko works over it, I'd be happy to move it back to dulwich proper (and e.g. making it the default on Windows).

Friz-zy: Okay :)
Is there some ready-to-go tests for the subprocess based class that we can use for it?
There is a [paramiko test for client](https://github.com/paramiko/paramiko/blob/master/tests/test_client.py) under LGPL that we can use, also there are a lot of some mock ssh servers and even docker image with git server and ssh.
But imo we can do a easier\better way to test ssh channel: we can add a read only ssh key into this repo (it called 'deployment key') and test all operation over real existed data :)
We can use it for both native ssh client and paramiko.

We can lock paramiko dependency into one latest specific version and check it properly as other solution

dulwich/client.py
jelmer: I think this would be simpler; initially, we could just check the value of sys.platform to determine the default value for get_ssh_vendor, but we could move to something more complex in the future.

Friz-zy:
@jelmer okay, so for this PR I'll move `putty` into separate class and add `unimplemented error` into main ssh class in case of `password` variable instead of using `sshpass`

What about test cases? I think that that the easier way is test the real git repo with real key.
I can add my own repo but it would be more logic to use `dulwich` own repo instead

jelmer:
Thanks, that would be great. For the putty class, just mocking out subprocess.call() and verifying that putty gets called with the right arguments should be sufficient. 

As for adding tests for the paramiko side of things, let's do that in a separate pull request and keep the discussion related to that in #364. That way, this PR can land while we sort out the paramiko tests.

dulwich/client.py
jelmer: Replying on #364, to keep these two conversations separate.

Friz-zy:
@jelmer I think that this is done, I'll send paramiko tests in separate PR for 364

dulwich/contrib/paramiko_vendor.py
```diff
@@ -111,22 +111,57 @@ def close(self):
...
-    def run_command(self, host, command, username=None, port=None,
-                    progress_stderr=None):
-        # Paramiko needs an explicit port. None is not valid
-        if port is None:
-            port = 22
+    # http://docs.paramiko.org/en/2.4/api/client.html
+
+    def __init__(self, **kwargs):
+        self.ssh_kwargs = {
+            'port': 22,
```
jelmer: Please don't set defaults here; rather just don't specify them, and paramiko can take care of the defaults.

dulwich/client.py
```diff
@@ -1102,16 +1107,67 @@ def __init__(self, hostname):
...
         args.append(host)
+
+        proc = subprocess.Popen(args + [command], bufsize=0,
+                                stdin=subprocess.PIPE,
+                                stdout=subprocess.PIPE)
+        return SubprocessWrapper(proc)
+
+
+class PuttySSHVendor(SSHVendor):
+    """SSH vendor that shells out to the local 'putty' command."""
```
jelmer: Please add a note that specifying the password will cause it to leak to other users on the local machine.

dulwich/tests/test_client.py
```diff
@@ -23,6 +23,7 @@
 import sys
 import shutil
 import tempfile
+import mock
```
jelmer: Please just monkey patch, so we can avoid a dependency on mock just for these two tests.

Friz-zy:
This one normal? )

Friz-zy:
@jelmer I added warning into code instead of adding into docstrings...

jelmer:
@Friz-zy AFAICT you only added a warning for the regular SSH Vendor, not for the Putty one.

dulwich/client.py
```diff
@@ -1102,16 +1107,72 @@ def __init__(self, hostname):
...
+
+        if sys.platform == 'win32':
+            args = ['putty.exe', '-ssh']
+        else:
+            args = ['putty', '-ssh']
+
+        if password:
+            import warnings
+            warnings.warn(
+                "Putty show you password in clear text "
```
Friz-zy: No, I added it into `putty` class too with default `user` level :)

jelmer: Ah, sorry - I missed that. Will be merging shortly :)

Friz-zy: @jelmer No problem :)
Btw we can move it into docs instead of warning if you think that this would be better
In case if you plan new release, I planned send another 2 PRs soon: one for paramiko tests and possibly one for porcelain

jelmer: A warning is fine. I'm still not a fan of having this functionality because it leads to users exposing their password without the ability to do anything about it. I'm hoping we can remove it if the paramiko vendor ends up in a better state.

Friz-zy: Yep, I agree

jelmer:
Sorry, I had to roll this back. Please fix the tests.

jelmer:
Nevermind, this is simple to fix. I'll push a fix.

PR #617, opened 2018-03-29
Add a new SSHVendor to allow plink use for SSH connections.

atbradley:
The PLinkSSHVendor uses the [command-line tool `plink`](https://www.ssh.com/ssh/putty/putty-manuals/0.68/Chapter7.html) instead of `putty` to make SSH connections.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,6 +5,9 @@
   * Add 'dulwich.mailmap' file for reading mailmap files.
     (Jelmer Vernooĳ)
 
+  * New 'client.PLinkSSHVendor' for creating connections using PuTTY's plink.exe.
+    (Adam Bradley)
+
  API CHANGES
 
   * Index.iterblobs has been renamed to Index.iterobjects.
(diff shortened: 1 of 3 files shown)


jelmer:
Is there any reason why you'd want to use putty rather than plink in a SSHVendor?

atbradley:
Not that I can think of. The PuttySSHVendor was already there; I just adapted it to work with plink.

jelmer:
Adding in @Friz-zy, who implemented the putty SSH Vendor and has more background, to comment. Is there any reason why we'd want to invoke putty rather than plink?

Friz-zy:
Hi guys :)
I'm not keen on windows but as I understand, plink is a console alternative of putty and it's already existed in putty distribution
I don't see a big problem of using it )

jelmer:
On Fri, Mar 30, 2018 at 06:19:29AM -0700, Filipp Frizzy wrote:
> Hi guys :)
> I'm not keen on windows but as I understand, plink is a console alternative of putty and it's already existed in putty distribution
> I don't see a big problem of using it )
Let's switch to plink altogether in that case.

-- 

atbradley:
I've submitted pull request #619, which simply replaces `putty` with `plink`.

PR #619, opened 2018-04-02
Replace the PuttySSHVendor with a PLinkSSHVendor.

atbradley:
See also pull request #617.


jelmer:
Thanks!

Can you please also update the tests to use the new PLink SSH vendor?

jelmer:
Nevermind, it looks like this is just a matter of fixing an import.

PR #634, opened 2018-06-14
Fix ls_tree function

r0mainK:
Argument `recursive` was not being used in the function, which led to the tree of sub directories in a given repo to be listed, even if not specified.


jelmer:
Could you add a test to dulwich/tests/test_porcelain.py that uses recursive?

r0mainK:
Done

jelmer:
Merged, thanks!

PR #635, opened 2018-06-14
Fix path_to_tree_path

r0mainK:
A couple entangled issues here:

1. In`check_ignore`: 
- `path_to_tree_path` was not being used appropriately, basically `path = os.path.relpath(path, r.path)` was done twice
- paths in index are always bytes, but paths passed to `ignore_manager` should always be strings: had to add some encoding/decoding

2. In `path_to_tree_path`:
- not checking if the tree_path did indeed start by repo_path 
- types again, after all calls to `path_to_tree_path` the returned path is used to check if it's in the index: might as well always return bytes. also small bug possible when replacing the separator, if the input path was in fact bytes.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -174,16 +174,25 @@ def open_repo_closing(path_or_repo):
 
 
 def path_to_tree_path(repopath, path):
-    """Convert a path to a path usable in e.g. an index.
+    """Convert a path to a path usable in an index, e.g. bytes and relative to
+    the repository root.
 
-    :param repo: Repository
-    :param path: A path
+    :param repopath: Repository path, absolute or relative to the cwd
+    :param path: A path, absolute or relative to the cwd
     :return: A path formatted for use in e.g. an index
     """
-    os.path.relpath(path, repopath)
-    if os.path.sep != '/':
-        path = path.replace(os.path.sep, '/')
-    return path.encode(sys.getfilesystemencoding())
+    if not isinstance(path, bytes):
+        path = path.encode(sys.getfilesystemencoding())
+    if not isinstance(repopath, bytes):
+        repopath = repopath.encode(sys.getfilesystemencoding())
+    if os.path.isabs(path) and not os.path.isabs(repopath):
+        repopath = os.path.abspath(repopath)
+    elif os.path.isabs(repopath) and not os.path.isabs(path):
+        path = os.path.abspath(path)
+    treepath = os.path.relpath(path, repopath)
+    if treepath.startswith(b'..'):
+        raise ValueError('Path not in repo')
+    return treepath
 
 
 def archive(repo, committish=None, outstream=default_bytes_out_stream,
@@ -1188,10 +1197,10 @@ def check_ignore(repo, paths, no_index=False):
         index = r.open_index()
         ignore_manager = IgnoreFilterManager.from_repo(r)
         for path in paths:
-            if os.path.isabs(path):
-                path = os.path.relpath(path, r.path)
             if not no_index and path_to_tree_path(r.path, path) in index:
                 continue
+            if os.path.isabs(path):
+                path = os.path.relpath(path, r.path)
             if ignore_manager.is_ignored(path):
                 yield path
 
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -827,7 +827,7 @@ class StatusTests(PorcelainTestCase):
             results.staged)
         self.assertEqual([], results.unstaged)
 
-    def test_status(self):
+    def test_status_base(self):
         """Integration test for `status` functionality."""
 
         # Commit a dummy file then modify it
@@ -858,6 +858,41 @@ class StatusTests(PorcelainTestCase):
         self.assertEqual(results.staged['add'][0],
                          filename_add.encode('ascii'))
         self.assertEqual(results.unstaged, [b'foo'])
+
+    def test_status_all(self):
+        del_path = os.path.join(self.repo.path, 'foo')
+        mod_path = os.path.join(self.repo.path, 'bar')
+        add_path = os.path.join(self.repo.path, 'baz')
+        us_path = os.path.join(self.repo.path, 'blye')
+        ut_path = os.path.join(self.repo.path, 'blyat')
+        with open(del_path, 'w') as f:
+            f.write('origstuff')
+        with open(mod_path, 'w') as f:
+            f.write('origstuff')
+        with open(us_path, 'w') as f:
+            f.write('origstuff')
+        porcelain.add(repo=self.repo.path, paths=[del_path, mod_path, us_path])
+        porcelain.commit(repo=self.repo.path, message=b'test status',
+                         author=b'author <email>',
+                         committer=b'committer <email>')
+        porcelain.remove(self.repo.path, [del_path])
+        with open(add_path, 'w') as f:
+            f.write('origstuff')
+        with open(mod_path, 'w') as f:
+            f.write('more_origstuff')
+        with open(us_path, 'w') as f:
+            f.write('more_origstuff')
+        porcelain.add(repo=self.repo.path, paths=[add_path, mod_path])
+        with open(us_path, 'w') as f:
+            f.write('\norigstuff')
+        with open(ut_path, 'w') as f:
+            f.write('origstuff')
+        results = porcelain.status(self.repo.path)
+        self.assertDictEqual(
+            {'add': [b'baz'], 'delete': [b'foo'], 'modify': [b'bar']},
+            results.staged)
+        self.assertListEqual(results.unstaged, [b'blye'])
+        self.assertListEqual(results.untracked, ['blyat'])
 
     def test_get_tree_changes_add(self):
         """Unit test for get_tree_changes add."""
@@ -1283,27 +1318,48 @@ class CheckIgnoreTests(PorcelainTestCase):
 
     def test_check_ignored(self):
         with open(os.path.join(self.repo.path, '.gitignore'), 'w') as f:
-            f.write("foo")
-        with open(os.path.join(self.repo.path, 'foo'), 'w') as f:
-            f.write("BAR")
-        with open(os.path.join(self.repo.path, 'bar'), 'w') as f:
-            f.write("BAR")
+            f.write('foo')
+        foo_path = os.path.join(self.repo.path, 'foo')
+        with open(foo_path, 'w') as f:
+            f.write('BAR')
+        bar_path = os.path.join(self.repo.path, 'bar')
+        with open(bar_path, 'w') as f:
+            f.write('BAR')
         self.assertEqual(
             ['foo'],
-            list(porcelain.check_ignore(self.repo, ['foo'])))
-        self.assertEqual([], list(porcelain.check_ignore(self.repo, ['bar'])))
+            list(porcelain.check_ignore(self.repo, [foo_path])))
+        self.assertEqual(
+            [], list(porcelain.check_ignore(self.repo, [bar_path])))
 
-    def test_check_added(self):
-        with open(os.path.join(self.repo.path, 'foo'), 'w') as f:
-            f.write("BAR")
+    def test_check_added_abs(self):
+        path = os.path.join(self.repo.path, 'foo')
+        with open(path, 'w') as f:
+            f.write('BAR')
         self.repo.stage(['foo'])
         with open(os.path.join(self.repo.path, '.gitignore'), 'w') as f:
-            f.write("foo\n")
+            f.write('foo\n')
         self.assertEqual(
-            [], list(porcelain.check_ignore(self.repo, ['foo'])))
+            [], list(porcelain.check_ignore(self.repo, [path])))
         self.assertEqual(
             ['foo'],
-            list(porcelain.check_ignore(self.repo, ['foo'], no_index=True)))
+            list(porcelain.check_ignore(self.repo, [path], no_index=True)))
+
+    def test_check_added_rel(self):
+        with open(os.path.join(self.repo.path, 'foo'), 'w') as f:
+            f.write('BAR')
+        self.repo.stage(['foo'])
+        with open(os.path.join(self.repo.path, '.gitignore'), 'w') as f:
+            f.write('foo\n')
+        cwd = os.getcwd()
+        os.mkdir(os.path.join(self.repo.path, 'bar'))
+        os.chdir(os.path.join(self.repo.path, 'bar'))
+        try:
+            self.assertEqual(
+                list(porcelain.check_ignore(self.repo, ['../foo'])), [])
+            self.assertEqual(['../foo'], list(
+               porcelain.check_ignore(self.repo, ['../foo'], no_index=True)))
+        finally:
+            os.chdir(cwd)
 
 
 class UpdateHeadTests(PorcelainTestCase):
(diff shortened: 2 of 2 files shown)


dulwich/porcelain.py
```diff
@@ -170,17 +170,21 @@ def open_repo_closing(path_or_repo):
     return closing(Repo(path_or_repo))
 
 
-def path_to_tree_path(repopath, path):
+def path_to_tree_path(repo_path, tree_path):
```
jelmer: See comments in other PR, please don't rename arguments.

dulwich/porcelain.py
```diff
@@ -170,17 +170,21 @@ def open_repo_closing(path_or_repo):
...
-    :param path: A path
-    :return: A path formatted for use in e.g. an index
+    :param repo_path: Absolute path to the repository
+    :param tree_path: Possibly non relative path to a file in the repository
+    :return: A path formatted for use in e.g. an index (bytes)
     """
-    os.path.relpath(path, repopath)
+    if tree_path.startswith(repo_path):
+        tree_path = os.path.relpath(tree_path, repo_path)
+    if isinstance(tree_path, str):
```
jelmer: This is invalid on Python2.

dulwich/porcelain.py
```diff
@@ -1189,9 +1193,12 @@ def check_ignore(repo, paths, no_index=False):
         ignore_manager = IgnoreFilterManager.from_repo(r)
         for path in paths:
             if os.path.isabs(path):
-                path = os.path.relpath(path, r.path)
```
jelmer: This is intentional. paths specified here are relative to the current path, not to the repository root (similar to the behaviour of "git check-ignore").

dulwich/porcelain.py
r0mainK: which part ? `startswith` is defined, as is `isinstance`, no ?

dulwich/porcelain.py
r0mainK: hmm let me get this straight. 
- if at the start both `path` and `r.path` are already relative to the current path then you do not need this
- if at the start you know for sure that `r.path` is relative to the current path, and `path` might be absolute, then the right code should be: `path = os.path.relpath(path, os.getcwd())` no ?

dulwich/porcelain.py
jelmer: the isinstance bit; if something is a bytestring on python2 and it has non-ascii characters, then encoding it will lead to mayhem since it tries to convert from ascii:

>>> "ĳ".encode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 0: ordinal not in range(128)

r0mainK:
So I guess I have two follow up question after seeing the comments, before I add rewrite and add tests: 

- should `path_to_tree_path` assume that `path` and `repopath` are both relative paths (to the same dir) or not ? 
- regarding types, should there be assumptions on `path` and `repopath`'s types, and is there a specific type one should expect as output ?

My personal opinion is that `path` and `repopath` should be assumed to be of the same type, either bytes or str, and that the output should always be bytes, given the usage of this function is exclusively to check if paths are in the `Index`. Furthermore, we should not have to check both paths are relative to the same path, the job should have been done previously.

dulwich/porcelain.py
jelmer: Either path or r.path can be absolute paths, so we do need the call to os.path.relpath. 

os.path.relpath will determine the absolute paths for path and r.path relative to os.getcwd() *and* then determine the relative path between them. E.g.:

    % cd /tmp
    /tmp% python
    Python 2.7.15 (default, May  1 2018, 05:55:50) 
    [GCC 7.3.0] on linux2
    >>> import os
    >>> os.path.relpath("foo", "/home/blah/bar")
    '../../../tmp/foo'
    >>> os.path.relpath("/tmp/foo", "/home/blah/bar")
    '../../../tmp/foo'

"foo" gets interpreted as "/tmp/foo" and then the relative path from /home/blah/bar is determined.

jelmer:
tree/index paths are supposed to be bytestrings (since they're git paths)

repopath and paths passed into the various porcelain functions are filesystem paths; they can be either bytes or strings, depending on the OS/Python version you're on. The porcelain functions should be able to deal with all this, hence the path_to_tree_path helper.

dulwich/porcelain.py
r0mainK: ah right, then we can just do the opposite: `not isinstance(f(path), bytes)`. just checked it seems to work in python2

r0mainK:
@jelmer  ok thks I understand better now. So in the end no need to change the logic of `check_ignore`, also since `path_to_tree_path` is only a helper function that makes sure the input path is bytes with the correct separator I just removed the line `os.path.relpath(...)` that did nothing and took out the repopath from the input, since not used. Also added a test.

dulwich/porcelain.py
```diff
@@ -170,17 +170,18 @@ def open_repo_closing(path_or_repo):
...
 
-    :param repo: Repository
-    :param path: A path
-    :return: A path formatted for use in e.g. an index
+    :param path: Path relative to repopath of a file in the repository
+    :return: A path formatted for use in e.g. an index (bytes)
     """
-    os.path.relpath(path, repopath)
+    if not isinstance(path, bytes):
+        path = path.encode(DEFAULT_ENCODING)
```
jelmer: You'll want to decode from sys.getfilesystemencoding() if original path was a filesystem path.

E.g. the filesystem path can be iso8859, while the git encoding is 'utf-8'.

dulwich/porcelain.py
```diff
@@ -387,7 +388,7 @@ def remove(repo=".", paths=None, cached=False):
         index = r.open_index()
         for p in paths:
             full_path = os.path.abspath(p).encode(sys.getfilesystemencoding())
-            tree_path = path_to_tree_path(r.path, p)
+            tree_path = path_to_tree_path(p)
```
jelmer: tree_path needs to be relative to the repository path; where is this enforced now? os.getcwd() is not necessarily r.path.

r0mainK: Well it was never enforced here apparently, as you can see from the diff path_to_tree_path did not enforce it before since the os.path.relpath did not change the value of path

r0mainK: anyway since path is either going to be a relative path to cwd or an absolute path we could just systematically check in `path_to_tree_path` with `isabs` and apply `relpath` if False

jelmer: Agreed. IMO that was the bug, not that r.path was passed into path_to_tree_path in the first place.

jelmer:
A related bug is #598

r0mainK:
@jelmer hey, sorry for being afk last week ^^" anyway I looked into #598 and it was indeed related. With the changes I brought it worked for me. 
So now here is the flow of `path_to_tree_path`:

1. Convert both `path` and `repopath` to bytes, replace separator if needed.
2. Test if `path` is relative to `repopath`. Can't use `os.path.join` because it detects common absolute paths, so I do `(repopath + b'/' + path).replace(b'//', b'/')`. If this exists, then we can return `path` directly.
3. If not, at this point we know both `path` and `repopath` can be absolute or relative to CWD. If one is absolute and the other relative, we transform the relative to absolute, then return `relpath(path, repopath)`

I also added tests for `path_to_tree_path`. Regarding the builds,  two test suites are failing (Python:2.7 and Python: pypy) with **a lot** of errors, but the errors do not seem related to my code, e.g.:

```
ERROR: test_simple (dulwich.tests.test_porcelain.LsFilesTests)
  File "dulwich/tests/test_porcelain.py", line 1261, in test_simple
    porcelain.add(repo=self.repo.path, paths=[fullpath])
  File "dulwich/porcelain.py", line 386, in add
    if ignore_manager.is_ignored(relpath):
  File "dulwich/ignore.py", line 336, in is_ignored
    matches = list(self.find_matching(path))
  File "dulwich/ignore.py", line 308, in find_matching
    raise ValueError('%s is an absolute path' % path)
ValueError: /tmp/tmpbM2QIK/repo/foo is an absolute path
```

In this error `path_to_tree_path` is not called, and I did not touch anything else. By the way, I saw you just did a release but I would rly appreciate if you could do a new one once this gets merged, be it as is or after changes. I've finished a project that relies on dulwich last week, and it fails because of the #598 right now ^^".

dulwich/porcelain.py
```diff
@@ -172,16 +172,29 @@ def open_repo_closing(path_or_repo):
...
 
 def path_to_tree_path(repopath, path):
-    """Convert a path to a path usable in e.g. an index.
+    """Convert a path to a path usable in an index, e.g. bytes and relative to
+    the repository root.
 
-    :param repo: Repository
-    :param path: A path
+    :param repopath: Repository path, absolute or relative to the cwd
+    :param path: A path, absolute or relative to either the cwd or repopath
```
jelmer: Allowing relative paths to either cwd or repopath makes the behaviour of this function unpredictable. This should just be cwd (to be compatible with c git).

dulwich/porcelain.py
```diff
@@ -172,16 +172,29 @@ def open_repo_closing(path_or_repo):
...
+    if not isinstance(repopath, bytes):
+        repopath = repopath.encode(sys.getfilesystemencoding())
     if os.path.sep != '/':
-        path = path.replace(os.path.sep, '/')
-    return path.encode(sys.getfilesystemencoding())
+        path = path.replace(
+            os.path.sep.encode(sys.getfilesystemencoding()), b'/')
+        repopath = repopath.replace(
+            os.path.sep.encode(sys.getfilesystemencoding()), b'/')
+    if os.path.exists((repopath + b'/' + path).replace(b'//', b'/')):
```
jelmer: See above - this makes the behaviour of the function unpredictable, relative paths should only ever be interpreted in one way.

dulwich/tests/test_porcelain.py
```diff
@@ -1375,3 +1376,44 @@ def test_git_dir(self):
...
+            ".", "./bar/baz"))
+        self.assertEqual(b"bar/baz", porcelain.path_to_tree_path(
+            ".", "bar/baz"))
+        cwd = os.getcwd()
+        path = os.path.join(cwd, "bar/baz")
+        self.assertEqual(b"bar/baz", porcelain.path_to_tree_path(
+            ".", path))
+        self.assertEqual(b"bar/baz", porcelain.path_to_tree_path(
+            cwd, "bar/baz"))
+        os.mkdir("foo")
```
jelmer: Please use TestCaseInTempDir if you're going to create files/directories, otherwise we're leaving these all over the source directory.

dulwich/porcelain.py
r0mainK: I agree, this was to stay compatible with check ignore command which seemed, according to tests, to necessitate this. Should i change tests or logic of that function then ?

dulwich/tests/test_porcelain.py
r0mainK: My bad, Will change first thing tomoreiw

dulwich/porcelain.py
jelmer: dulwich.porcelain.check_ignore should also behave similarly to "git check-ignore". It looks like this will indeed require changing the tests for check_ignore.

r0mainK: done

r0mainK:
@jelmer did the changes you asked, but again tests for Python2.7 and PyPy are failing, do you know what this is about ?

jelmer:
On Tue, Jun 26, 2018 at 05:53:50AM -0700, Romain Keramitas wrote:
> @jelmer did the changes you asked, but again tests for Python2.7 and PyPy are failing, do you know what this is about ?
I'll take a look this evening.

r0mainK:
@jelmer ping  - I know you made comments on the other PR that I will address, but since it is a feature for `add` which uses `path_to_tree_path` via `get_untracked_paths`I would like to get this merged before in all cases. It fixes, as far as I know, #598 and again tests suites for `Python2.7` and `Pypy` are currently failing for reasons I do not understand, if you could have a look Id really appreciate.

dulwich/porcelain.py
```diff
@@ -1186,10 +1197,10 @@ def check_ignore(repo, paths, no_index=False):
         index = r.open_index()
         ignore_manager = IgnoreFilterManager.from_repo(r)
         for path in paths:
-            if os.path.isabs(path):
-                path = os.path.relpath(path, r.path)
             if not no_index and path_to_tree_path(r.path, path) in index:
                 continue
+            if os.path.isabs(path):
```
jelmer: I suspect this change is what made the python2.7 tests fail.

dulwich/tests/test_porcelain.py
```diff
@@ -1283,27 +1283,31 @@ class CheckIgnoreTests(PorcelainTestCase):
...
         with open(os.path.join(self.repo.path, '.gitignore'), 'w') as f:
-            f.write("foo\n")
+            f.write('foo\n')
         self.assertEqual(
-            [], list(porcelain.check_ignore(self.repo, ['foo'])))
+            [], list(porcelain.check_ignore(self.repo, [path])))
         self.assertEqual(
             ['foo'],
-            list(porcelain.check_ignore(self.repo, ['foo'], no_index=True)))
+            list(porcelain.check_ignore(self.repo, [path], no_index=True)))
```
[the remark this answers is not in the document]
r0mainK: done

dulwich/tests/test_porcelain.py
```diff
@@ -1375,3 +1379,47 @@ def test_git_dir(self):
         self.assertEqual(
                 [(obj.id, 'invalid name .git')],
                 [(sha, str(e)) for (sha, e) in porcelain.fsck(self.repo)])
+
+
+class HelperTests(PorcelainTestCase):
+
+    def test_path_to_tree_path_base(self):
```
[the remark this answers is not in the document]
r0mainK: Sure, then I'll add a ValueError to be raised when the returned path starts with `..`

dulwich/porcelain.py
r0mainK: thks, gonna check it out

r0mainK:
@jelmer added the 3 tests you wanted, and checked if it was the continue block the problem but it doesnt seem so, I wonder if Travis isnt broken for some reason - like when it seemingly cant find `refs` Oo

r0mainK:
ok so actually error was coming from `test_path_to_tree_path_base` i added, gonna rework it so it deosnt fail

r0mainK:
OK so the error was in the tests I added for `path_to_tree_path`, more precisely when testing the case where the separator is different from `/`. This allowed me to see that it seems the `os.path` library of python 2 and 3 is not compatible with this regard:
- in python 2, I had to change the separator after I had done `relpath`:
```
    if os.path.sep != '/':
        treepath = treepath.replace(
            os.path.sep.encode(sys.getfilesystemencoding()), b'/')
```
-> or the returned tree_path was an absolute path: `b'/Users/romain/dulwich/bar/baz'`
- in python 3, I had to do it before:
```
    if os.path.sep != '/':
        path = path.replace(
            os.path.sep.encode(sys.getfilesystemencoding()), b'/')
        repopath = repopath.replace(
            os.path.sep.encode(sys.getfilesystemencoding()), b'/')
```
-> or relpath failed, raising a ValueError bc `tree_path= b'../@sepUsers@sepromain@sepdulwich@sepbar@sepbaz'`

So I have taken this test and feature out.

jelmer:
Thanks, Merged!

PR #641, opened 2018-07-01
Simple describe implementation

TheLastProject:
This has been used by Pext to generate the version strings shown in the about screen rather successfully for months now, so I feel okay enough with this to submit it upstream. It doesn't support all git describe does (no flags are supported), but it's a simple basic implementation that would fulfil the most common use cases.

screenshot_20180701_201333

### DIFF
--- a/bin/dulwich
+++ b/bin/dulwich
@@ -614,6 +614,14 @@ class cmd_ls_files(Command):
             print(name)
 
 
+class cmd_describe(Command):
+
+    def run(self, args):
+        parser = optparse.OptionParser()
+        options, args = parser.parse_args(args)
+        print(porcelain.describe('.'))
+
+
 class cmd_help(Command):
 
     def run(self, args):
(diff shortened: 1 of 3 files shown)


jelmer:
Thanks! Any chance you could add a basic unit test for dulwich.porcelain.describe in dulwich/tests/test_porcelain.py ?

TheLastProject:
I'm not sure how to best unit test this, as the behaviour of describe isn't that clearly documented. Would it be acceptable to call describe from the normal git client and ensure dulwich generates the same output? If not, what would be an acceptable way to write a test for this?

jelmer:
On Mon, Jul 02, 2018 at 01:06:30PM -0700, Sylvia van Os wrote:
> I'm not sure how to best unit test this, as the behaviour of describe isn't that clearly documented. Would it be acceptable to call describe from the normal git client and ensure dulwich generates the same output? If not, what would be an acceptable way to write a test for this?
Probably what you'd want to do is come up with a few situations in
which you know what this implementation of describe should be doing.

For each of these situations you can then add a unit test that
prepares a repository with the required state (see the other tests in
dulwich/tests/test_porcelain.py for examples), and then verifies that
describe() returns the right value.

-- 

jelmer:
Merged, thanks!

TheLastProject:
Thanks for the last bit with the decoding. Always trips me up :)

PR #643, opened 2018-07-22
Pass the <errstream> parameter through from porcelain.clone() to porcelain.fetch().

gruns:
Without such, the `errstream` parameter only applies to

```python
errstream.write(b'Checking out ' + head.id + b'\n')
```

All other output, like that generated by `fetch()` (called by `clone()`), is
written, unwanted, to `porcelain.default_bytes_err_stream`.


jelmer:
Merged, thanks!

PR #646, opened 2018-08-06
Test on Python 3.7 and 3.8-dev

polyzen:


jelmer:
Merged, thanks!

-- 

PR #653, opened 2018-08-19
Add fallback for unavailable stdout/stderr

### LINKED ISSUE #652: Dulwich throws AttributeException when used in pythonw
I got the following report from an user: https://github.com/Pext/Pext/issues/125

On Pext's side, installing a module calls the install_module function. The only moment in that function when 1 downward arrow with an exception is printed is here:

https://github.com/Pext/Pext/blob/095f5a5a9121141ccd16704ee486e3c943315d3f/pext/__main__.py#L2835-L2839

Clearly, the issue lies in dulwich's porcelain.clone().

Looking further into that, dulwich does a .write call here:
https://github.com/dulwich/dulwich/blob/e924e8bdfb71a7ecb6f88f4e6627cb7d2e344e1a/dulwich/porcelain.py#L349

Looking at
https://github.com/dulwich/dulwich/blob/e924e8bdfb71a7ecb6f88f4e6627cb7d2e344e1a/dulwich/porcelain.py#L289-L291

errstream defaults to default_bytes_err_stream, which is defined as follows:

https://github.com/dulwich/dulwich/blob/e924e8bdfb71a7ecb6f88f4e6627cb7d2e344e1a/dulwich/porcelain.py#L142-L143

Looking at the Python documentation, it is incorrect to assume sys.stdout and sys.stderr are not None on Windows (see https://docs.python.org/3/library/sys.html#sys.stderr):
>Under some conditions stdin, stdout and stderr as well as the original values __stdin__, __stdout__ and __stderr__ can be None. It is usually the case for Windows GUI apps that aren’t connected to a console and Python apps started with pythonw.

I can probably fix this myself, but I'd need to know how you'd prefer this fixed. We could do either of the following:
1. Add a small fallback class with a .write function that does nothing (seems cleanest to me, but risks having to make a very complex class at some point)
2. A try/except AttributeError block around every .write call on errstream or outstream
3. A None-check before any write call.

Which would you prefer? Or do you perhaps have a better idea altogether?

TheLastProject:
I think theoretically this should fix #652, but I'm not sure how to properly test this, as I don't have a Windows machine myself so can't try it under pythonw.exe myself.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -61,7 +61,7 @@ from contextlib import (
     closing,
     contextmanager,
 )
-from io import BytesIO
+from io import BytesIO, RawIOBase
 import datetime
 import os
 import posixpath
@@ -139,8 +139,27 @@ from dulwich.server import (
 GitStatus = namedtuple('GitStatus', 'staged unstaged untracked')
 
 
-default_bytes_out_stream = getattr(sys.stdout, 'buffer', sys.stdout)
-default_bytes_err_stream = getattr(sys.stderr, 'buffer', sys.stderr)
+class NoneStream(RawIOBase):
+    """Fallback if stdout or stderr are unavailable, does nothing."""
+    def read(size=-1):
+        return None
+
+    def readall():
+        return None
+
+    def readinto(b):
+        return None
+
+    def write(b):
+        return None
+
+
+default_bytes_out_stream = getattr(
+        sys.stdout, 'buffer', sys.stdout
+    ) or NoneStream
+default_bytes_err_stream = getattr(
+        sys.stderr, 'buffer', sys.stderr
+    ) or NoneStream
 
 
 DEFAULT_ENCODING = 'utf-8'
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -141,25 +141,25 @@ GitStatus = namedtuple('GitStatus', 'staged unstaged untracked')
 
 class NoneStream(RawIOBase):
     """Fallback if stdout or stderr are unavailable, does nothing."""
-    def read(size=-1):
+    def read(self, size=-1):
         return None
 
-    def readall():
+    def readall(self):
         return None
 
-    def readinto(b):
+    def readinto(self, b):
         return None
 
-    def write(b):
+    def write(self, b):
         return None
 
 
 default_bytes_out_stream = getattr(
         sys.stdout, 'buffer', sys.stdout
-    ) or NoneStream
+    ) or NoneStream()
 default_bytes_err_stream = getattr(
         sys.stderr, 'buffer', sys.stderr
-    ) or NoneStream
+    ) or NoneStream()
 
 
 DEFAULT_ENCODING = 'utf-8'
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -79,12 +79,6 @@ class ArchiveTests(PorcelainTestCase):
         self.addCleanup(tf.close)
         self.assertEqual([], tf.getnames())
 
-    def test_simple_outstream_errstream_autofallback(self):
-        c1, c2, c3 = build_commit_graph(
-                self.repo.object_store, [[1], [2, 1], [3, 1, 2]])
-        self.repo.refs[b"refs/heads/master"] = c3.id
-        porcelain.archive(self.repo.path, b"refs/heads/master")
-
 
 class UpdateServerInfoTests(PorcelainTestCase):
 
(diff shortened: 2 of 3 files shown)


TheLastProject:
A tester has confirmed that a modified build of Pext using a version of dulwich with that patch applied indeed works: https://github.com/Pext/Pext/issues/125#issuecomment-414628250

jelmer:
Thanks! Could you add a simple test to dulwich/tests/test_porcelain.py that uses the new NoneStream (e.g. by just passing it to an errstream argument somewhere) so we know it doesn't regress?

TheLastProject:
Hmm, I actually notice that all the porcelain tests directly use a new instance of BytesIO for errstream. Wouldn't it be better to not set errstream anywhere, and change the appveyor test script to use both python.exe and pythonw.exe, so that the fallback is automatically correctly tested?

TheLastProject:
Nevermind, I misunderstood the code a bit. This change should help test it though

jelmer:
On Tue, Aug 21, 2018 at 04:19:50AM -0700, Sylvia van Os wrote:
> Hmm, I actually notice that all the porcelain tests directly use a new instance of BytesIO for errstream. Wouldn't it be better to not set errstream anywhere, and change the appveyor test script to use both python.exe and pythonw.exe, so that the fallback is automatically correctly tested?

I'd like to pass in NoneStream explicitly, since not doing so has two disadvantages:

 * NoneStream only gets tested on pythonw
 * It makes the testsuite spammy on other platforms, since it will now
   be writing to standard error

-- 
PGP Key: https://www.jelmer.uk/D730A457.asc

dulwich/tests/test_porcelain.py
```diff
@@ -79,6 +79,12 @@ def test_simple(self):
         self.addCleanup(tf.close)
         self.assertEqual([], tf.getnames())
 
+    def test_simple_outstream_errstream_autofallback(self):
+        c1, c2, c3 = build_commit_graph(
+                self.repo.object_store, [[1], [2, 1], [3, 1, 2]])
+        self.repo.refs[b"refs/heads/master"] = c3.id
+        porcelain.archive(self.repo.path, b"refs/heads/master")
```
jelmer: Please pass in NoneStream to errstream explicitly, otherwise this will be spamming the testsuite output.

jelmer: Please use a different porcelain command than archive, since it doesn't actually write to errstream at the moment.

TheLastProject:
Hmm, you are indeed right that this fix doesn't completely work.

It's not a regression (because without my patch it always crashes when there is no available stdout/stderr), but Python 2 doesn't seem to like this fix:
```
ERROR: test_no_head_no_checkout_outstream_errstream_autofallback (dulwich.tests.test_porcelain.CloneTests)
  File "dulwich/tests/test_porcelain.py", line 238, in test_no_head_no_checkout_outstream_errstream_autofallback
    self.repo.path, target_path, checkout=True, errstream=errstream)
  File "dulwich/porcelain.py", line 350, in clone
    **kwargs)
  File "dulwich/porcelain.py", line 1103, in fetch
    fetch_result = client.fetch(path, r, progress=errstream.write)
  File "dulwich/client.py", line 1026, in fetch
    progress=progress)
  File "dulwich/repo.py", line 274, in fetch
    determine_wants, target.get_graph_walker(), progress)
  File "dulwich/repo.py", line 296, in fetch_pack_data
    return pack_objects_to_data(objects)
  File "dulwich/pack.py", line 1590, in pack_objects_to_data
    count = len(objects)
  File "dulwich/object_store.py", line 980, in __len__
    return len(list(self.itershas()))
  File "dulwich/object_store.py", line 954, in itershas
    for sha in self.sha_iter:
  File "dulwich/object_store.py", line 1153, in next
    len(self.sha_done)).encode('ascii'))
TypeError: unbound method write() must be called with NoneStream instance as first argument (got str instance instead)
```

I've tried to fix it, but haven't been managing to. I'm honestly unsure why it does this on Python 2 (and pypy) but not on Python 3.

dulwich/porcelain.py
```diff
@@ -139,8 +139,27 @@
 GitStatus = namedtuple('GitStatus', 'staged unstaged untracked')
 
 
-default_bytes_out_stream = getattr(sys.stdout, 'buffer', sys.stdout)
-default_bytes_err_stream = getattr(sys.stderr, 'buffer', sys.stderr)
+class NoneStream(RawIOBase):
+    """Fallback if stdout or stderr are unavailable, does nothing."""
+    def read(size=-1):
```
jelmer: These functions all appear to be missing a "self" argument.

dulwich/tests/test_porcelain.py
```diff
@@ -224,6 +224,20 @@ def test_no_head_no_checkout(self):
...
+    def test_no_head_no_checkout_outstream_errstream_autofallback(self):
+        f1_1 = make_object(Blob, data=b'f1')
+        commit_spec = [[1]]
+        trees = {1: [(b'f1', f1_1), (b'f2', f1_1)]}
+
+        (c1, ) = build_commit_graph(self.repo.object_store, commit_spec, trees)
+        self.repo.refs[b"refs/heads/master"] = c1.id
+        target_path = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, target_path)
+        errstream = porcelain.NoneStream
```
jelmer: This is passing in the class, rather than an instance of the class. You'd probably want to call 

NoneStream()

here, rather than

NoneStream

TheLastProject:
Can't believe I missed that, thanks. Interesting that it worked on Python 3 though.

jelmer:
On Tue, Aug 21, 2018 at 02:16:25PM -0700, Sylvia van Os wrote:
> Can't believe I missed that, thanks. Interesting that it worked on Python 3 though.
I think they changed the way method instances work, perhaps that's why
it did.

Anyway, thanks for the bug fix. Merged :)

-- 

PR #669, opened 2018-11-06
Fix setup.py:

EgorBu:
* read "README.md" with "utf-8" encoding
* consistant quotes

it should fix https://github.com/dulwich/dulwich/issues/668

Signed-off-by: egor <egor@sourced.tech>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,17 +10,18 @@ except ImportError:
     has_setuptools = False
 else:
     has_setuptools = True
+import io
 from distutils.core import Distribution
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


jelmer:
Thanks! I've cherry-picked the io.open fix, but left the quote fixes.

Dulwich's preferred style for strings is single quotes, as documented in CONTRIBUTING.md. It also more commonly uses single quotes for strings (~5k vs ~3k double quotes). We haven't been very proactive about fixing these, but I'd rather not move backward :)

EgorBu:
Thank you!

PR #674, opened 2018-12-10
Line ending convert support

Lothiraldan:
It's a work-in-progress


Lothiraldan:
The `get_line_ending_filters` code is likely wrong at the moment

jelmer:
Thanks for working on this!

Lothiraldan:
@jelmer I went to a different approach, inspired by JGit implementation (https://github.com/eclipse/jgit/blob/master/org.eclipse.jgit/src/org/eclipse/jgit/util/io/EolStreamTypeUtil.java). Here is the gist of it:

- We create a blob like usual.
- Then we call `repo.normalize_blob_on_checkout` which may or may not wrap the blob in `NormalizedBlob` with the right convert function based on the configuration.
- `NormalizedBlob` either apply the convert function unconditionally or try to detect if the file is binary first. The binary detection is not implemented yet.

This pull-request lays the foundation for full line-ending support but don't fully support all of the possibilities:
- `.git-attributes` parsing, merging and processing is not supported yet. With the current API, the best place to process it would be in `get_checkout_filter` and `get_checkin_filter`. There is a `#TODO` in the code for that.
- Binary detection is not supported yet. Where a file is marked as `text=auto` via the `.git-attributes` file, each file goes through a binary detection heuristic. If the file is detected as binary, it's line endings are not modified. If it's detected as a text file (not binary), it's line ending is converted to `LF` at checkin and the configured EOL at checkout.

Both of those features will be complex as they open a big bag of questions about how they interact with the other features and configurations. I think the current API should work fine for supporting them in later pull-requests.

I tried documenting my findings in the `line_ending.py` module docstring, please tell me if something is missing.

This PR is still WIP, I have the proper docstring to write for the new methods and functions and surely a lot of cosmetic change to do.

@jelmer If you could give me your opinion about the current API, wrapping blob based on configuration and tell me if I'm heading in the right direction, that would be great. The part of the code where I'm the less confident is the `NormalizedBlob`, I don't know if I should override more methods, your expertise on this would be greatly appreciated.

Thank you again for your work on Dulwich, it's a great library.

jelmer:
Thanks, I'll try to have a look this week.

dulwich/tests/test_line_ending.py
```diff
@@ -0,0 +1,175 @@
...
+
+class NormalizedBlobTestCase(TestCase):
+    def test_normalize_to_lf_no_op(self):
+        base_content = b"line1\nline2"
+        base_sha = "f8be7bb828880727816015d21abcbc37d033f233"
+
+        base_blob = Blob()
+        base_blob.set_raw_string(base_content)
+
+        assert base_blob.as_raw_chunks() == [base_content]
```
jelmer: Please use self.assertEqual

dulwich/config.py
```diff
@@ -143,7 +146,11 @@ def get_boolean(self, section, name, default=None):
             return True
         elif value.lower() == b"false":
             return False
-        raise ValueError("not a valid boolean string: %r" % value)
+
+        if strict:
+            raise ValueError("not a valid boolean string: %r" % value)
+        else:
+            return value
```
jelmer: This makes get_boolean() return values that are not boolean. Please just use Config.get() in that case.

dulwich/line_ending.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def convert_lf_to_crlf(text_hunk):
+    """Convert LF in text hunk into CRLF"""
+    # TODO find a more efficient way of doing it
+    intermediary = text_hunk.replace(CRLF, LF)
+    return intermediary.replace(LF, CRLF)
+
+
+class NormalizedBlob(Blob):
```
jelmer: Could this be a function rather than a wrapper class that pokes at internals?

dulwich/line_ending.py
```diff
@@ -0,0 +1,197 @@
...
+        return convert_lf_to_crlf
+
+    return None
+
+
+def get_checkin_filter_autocrlf(core_autocrlf):
+    """ Returns the correct checkin filter base on autocrlf value
+    """
+
+    if core_autocrlf is True or core_autocrlf == b"input":
```
jelmer: It seems odd that this either takes a boolean or a string. Perhaps just pass in a string, even if that is "true" or "input" ?

Lothiraldan:
This PR has been updated now that #678 has been merged

jelmer:
Thanks! I think the main thing left to do is to move more of the transformation logic out of Repo. I've left more specific comments in the code.

Lothiraldan:
@jelmer I don't see your comments neither on the PR, neither on a new commit on master. Where did you left them?

dulwich/line_ending.py
```diff
@@ -179,3 +199,26 @@ def get_checkin_filter_autocrlf(core_autocrlf):
...
+        super(NormalizedBlob, self).__init__()
+        self._original_blob = blob
+        self.conversion = conversion
+        self.binary_detection = binary_detection
+        self._needs_serialization = True
+
+    def _serialize(self):
+        original_chunked_text = self._original_blob._chunked_text
+
+        if self.binary_detection is True:
```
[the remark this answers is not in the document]
Lothiraldan: Now that I know that we have a binary detection function in dulwich, I think it should be easy to implement it and test it

dulwich/repo.py
```diff
@@ -1261,6 +1265,50 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.close()
 
+    def normalize_blob_on_checkout(self, blob, tree_path):
```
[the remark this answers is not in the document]
Lothiraldan: That's doable, one thing to keep in mind for the design is the .gitattributes support. With the .gitattributes we will need to resolve the rules that apply to each file path during checkin and checkout. So we will have some processing to do for each file before choosing the correct conversion to apply (or not).

What is your main concern here? That we will be reading the configuration too many times and have a performance issue? That it's too tightly coupled to the repo object? Your answers will be precious to iterate on the design, I will start thinking about it.

jelmer:
On Tue, Jan 08, 2019 at 12:46:02PM -0800, Boris Feld wrote:
> Lothiraldan commented on this pull request.
> 
> 
> 
> > @@ -1261,6 +1265,50 @@ def __enter__(self):
>      def __exit__(self, exc_type, exc_val, exc_tb):
>          self.close()
>  
> +    def normalize_blob_on_checkout(self, blob, tree_path):
> 
> That's doable, one thing to keep in mind for the design is the .gitattributes support. With the .gitattributes we will need to resolve the rules that apply to each file path during checkin and checkout. So we will have some processing to do for each file before choosing the correct conversion to apply (or not).
> 
> What is your main concern here? That we will be reading the configuration too many times and have a performance issue? That it's too tightly coupled to the repo object? Your answers will be precious to iterate on the design, I will start thinking about it.

Specifically:

 * That reading the configuration too many times will have a
   performance impact
 * The amount of functionality that lives directly on the repository object
 * Tight coupling to the repository object; this means that testing
   this functionality requires constructing a repository

One way of doing this would be to create an object that takes the
contents of a .gitattributes file (parsed or not) and can then be
consulted for the checked in/checked out fulltext of files. Such an
object can live for the duration of a working tree operation (commit
or checkout) and could be constructed in Repo.stage, reused for each
of the files that is processed and discarded when that function returns.

Lothiraldan:
@jelmer I have update the PR with a first design based on `BlobNormalizer`. It works well for staging but it's more ackward for checkout as we don't have a centralized `unstage` method.

dulwich/line_ending.py
```diff
@@ -179,3 +200,77 @@ def get_checkin_filter_autocrlf(core_autocrlf):
...
+        """
+        if self.fallback_read_filter is not None:
+            return NormalizedBlob(
+                blob, self.fallback_read_filter, binary_detection=False
+            )
+
+        return blob
+
+
+class NormalizedBlob(Blob):
```
jelmer: Can we avoid the subclass of Blob here that peeks at its internals? AFAICT there is no added benefit in having a somewhat lazy wrapper; all callers just directly consume the returned blob object anyway.

jelmer:
Thanks! My only remaining concern is with NormalizedBlob, otherwise I think we're very close to being able to merge this.

Lothiraldan:
@jelmer I have replaced the NormalizedBlob object by a normalize_blob function using the `blob.data` attribute to read and write the content, please tell me if that's okay or if I should iterate on the text hunks instead.

Also, we need to decide if we activate line ending normalization by default or not

jelmer:
Merged, thanks!

PR #678, opened 2018-12-30
Line ending convert support bases

Lothiraldan:
This is an extract of https://github.com/dulwich/dulwich/pull/674#pullrequestreview-188184940 with only the conversion functions and associated tests.

### DIFF
--- a/dulwich/line_ending.py
+++ b/dulwich/line_ending.py
@@ -17,8 +17,113 @@
 # and <http://www.apache.org/licenses/LICENSE-2.0> for a copy of the Apache
 # License, Version 2.0.
 #
-
 """ All line-ending related functions, from conversions to config processing
+
+Line-ending normalization is a complex beast. Here is some notes and details
+about how it seems to work.
+
+The normalization is a two-fold process that happens at two moments:
+
+- When reading a file from the index and to the working directory. For example
+  when doing a `git clone` or `git checkout` call. We call this process the
+  read filter in this module.
+- When writing a file to the index from the working directory. For example
+  when doing a `git add` call. We call this process the write filter in this
+  module.
+
+One thing to know is that Git does line-ending normalization only on text
+files. How does Git know that a file is text? We can either mark a file as a
+text file, a binary file or ask Git to automatically decides. Git has an
+heuristic to detect if a file is a text file or a binary file. It seems based
+on the percentage of non-printable characters in files.
+
+The code for this heuristic is here:
+https://git.kernel.org/pub/scm/git/git.git/tree/convert.c#n46
+
+Dulwich have an implementation with a slightly different heuristic, the
+`is_binary` function in `dulwich.patch`.
+
+The binary detection heuristic implementation is close to the one in JGit:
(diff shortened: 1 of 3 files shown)


dulwich/tests/test_line_ending.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# test_line_ending.py -- Tests for the line ending functions
```
jelmer: Please add this file to dulwich/tests/__init__.py

dulwich/line_ending.py
```diff
@@ -0,0 +1,173 @@
...
+
+    :param core_autocrlf: The bytes configuration value of core.autocrlf.
+        Valid values are: b'true', b'false' or b'input'.
+    :return: Either None if no filter has to be applied or a function
+        accepting a single argument, a binary text hunk
+    """
+
+    if core_autocrlf == b"true" or core_autocrlf == b"input":
+        return convert_crlf_to_lf
+
```
jelmer: Is this consistent with C Git, i.e. does it just assume no conversion for other values of autocrlf?

dulwich/line_ending.py
```diff
@@ -0,0 +1,173 @@
...
+  when doing a `git clone` or `git checkout` call. We call this process the
+  read filter in this module.
+- When writing a file to the index from the working directory. For example
+  when doing a `git add` call. We call this process the write filter in this
+  module.
+
+One thing to know is that Git does line-ending normalization only on text
+files. How does Git know that a file is text? We can either mark a file as a
+text file, a binary file or ask Git to automatically decides. Git has an
+heuristic to detect if a file is a text file or a binary file. It seems based
```
jelmer: Note that Dulwich already has a "is_binary" function that lives in dulwich.patch and that fulfils the same function (though with a slightly different eurostic).

dulwich/line_ending.py
```diff
@@ -0,0 +1,173 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# You should have received a copy of the licenses; if not, see
+# <http://www.gnu.org/licenses/> for a copy of the GNU General Public License
+# and <http://www.apache.org/licenses/LICENSE-2.0> for a copy of the Apache
+# License, Version 2.0.
+#
+""" Line-ending normalization is a complex beast. Here is some notes and
```
jelmer: Please start the docstring with a one-line introduction of the module, for the documentation generator.

dulwich/line_ending.py
Lothiraldan: I have updated the module docstring to mention it

dulwich/line_ending.py
Lothiraldan: Without support of `.gitattributes` yes.
When we will supports `.gitattributes`, a file manually marked as a text file will be converted no matter the value of `core.autocrlf`.
This additional logic will likely lives in https://github.com/dulwich/dulwich/pull/674/files#diff-66fa4724245bb7627512d754f4c8b41dR161 and https://github.com/dulwich/dulwich/pull/674/files#diff-66fa4724245bb7627512d754f4c8b41dR170, hence the two `#TODO`

PR #679, opened 2019-01-17
Activate the binary detection by default

Lothiraldan:
https://github.com/dulwich/dulwich/pull/674 added automatic binary detection but didn't exposed a way to activate it from the BlobNormalizer layer.

I've added the basic argument to allow to activate it. But there is some usage of BlobNormalizer internally (in the `repo.stage` method and in the `porcelain.status` function). Should we activate the binary detection by default? Should we expose a parameter in repo to set it?

One piece of design to keep in mind, it is possible through `.gitattributes` file to force a file to be detected as a binary file or as a text file, no matter what its actual content is. So there will be some cases where the `binary_detection` will be set to off when calling `normalize_blob` once the `.gitattributes` will be added.

### DIFF
--- a/dulwich/line_ending.py
+++ b/dulwich/line_ending.py
@@ -234,7 +234,7 @@ class BlobNormalizer(object):
         """
         if self.fallback_write_filter is not None:
             return normalize_blob(
-                blob, self.fallback_write_filter, binary_detection=False
+                blob, self.fallback_write_filter, binary_detection=True
             )
 
         return blob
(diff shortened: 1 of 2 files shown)


jelmer:
I don't think we should expose this the porcelain level. C Git doesn't provide this functionality either, and I don't see why you would ever want to specify this behaviour on a per-stage/checkout basis. Is there any reason why we wouldn't just want to do binary detection whenever we're converting files?

Lothiraldan:
@jelmer I am mostly worried about backward-compatibility with previous releases of Dulwich. Also, the heuristic doesn't looks the same.
If you think that it's not a too important change, I would be okay just turning it be default.

jelmer:
The behaviour in previous releases is to treat all files as binary files (no filtering). If anything,  detecting binary files and excluding them from filters is improving backwards compatibility.

jelmer:
Thanks!

PR #686, opened 2019-02-05
Resolves TypeError: can only concatenate str (not "NoneType") to str

tiny-dancer:
https://github.com/dulwich/dulwich/issues/685


jelmer:
This was merged into trunk.

PR #688, opened 2019-02-09
Update directory detection for `get_unstaged_changes` for Python 3 - …

### LINKED ISSUE #684: get_unstaged_changes directory detection is broken in Python 3
I have identified a scenario where calling `get_unstaged_changes` in Python 3 could results in a crash. Here is some reproduce steps:

```
git init test
cd test
touch ROOT
git add ROOT
git commit -m "ROOT"
rm ROOT
mkdir ROOT
python -c "import dulwich.repo; import dulwich.index; repo=dulwich.repo.Repo('.'); list(dulwich.index.get_unstaged_changes(repo.open_index(), repo.path))"
```

I think the code is handling this case (here: https://github.com/dulwich/dulwich/blob/master/dulwich/index.py#L634) but the exception seems to have changed in Python 3, it is now a `IsADirectoryError` which is a subclass of `OSError`.

Lothiraldan:
…fix #684

The exception raised when trying to read a directory changed from an `IOError`
in Python 2 to an `IsADirectoryError` in Python 3 (`IsADirectoryError` is a
subclass of `OSError`).

`IsADirectoryError` doesn't exists in Python 2 so we have to makes extra hops
to have a check compatible with both Python 2 and Python 3.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -604,6 +604,32 @@ def read_submodule_head(path):
         return None
 
 
+def has_directory_changed(error, tree_path, entry):
+    """ When handling an error trying to create a blob from a path, call this
+    function. It will check if the path is a directory. If it's a directory
+    and a submodule, check the submodule head to see if it's has changed. If
+    not, consider the file as changed as Git tracked a file and not a
+    directory.
+
+    Return true if the given path should be considered as changed and False
+    otherwise or if the path is not a directory.
+    """
+    if error.errno != errno.EISDIR:
+        return False
+
+    # This is actually a directory
+    if os.path.exists(os.path.join(tree_path, b'.git')):
+        # Submodule
+        head = read_submodule_head(tree_path)
+        if entry.sha != head:
+            return True
+    else:
+        # The file was changed to a directory, so consider it removed.
+        return True
+
+    return False
+
+
 def get_unstaged_changes(index, root_path, filter_blob_callback=None):
     """Walk through an index and check for differences against working tree.
 
@@ -625,23 +651,21 @@ def get_unstaged_changes(index, root_path, filter_blob_callback=None):
             if filter_blob_callback is not None:
                 blob = filter_blob_callback(blob, tree_path)
         except OSError as e:
-            if e.errno != errno.ENOENT:
-                raise
-            # The file was removed, so we assume that counts as
-            # different from whatever file used to exist.
-            yield tree_path
-        except IOError as e:
-            if e.errno != errno.EISDIR:
-                raise
-            # This is actually a directory
-            if os.path.exists(os.path.join(tree_path, '.git')):
-                # Submodule
-                head = read_submodule_head(tree_path)
-                if entry.sha != head:
-                    yield tree_path
-            else:
-                # The file was changed to a directory, so consider it removed.
+            directory_changed = has_directory_changed(e, tree_path, entry)
+            if directory_changed:
                 yield tree_path
+            else:
+                if e.errno != errno.ENOENT:
+                    raise
+                # The file was removed, so we assume that counts as
+                # different from whatever file used to exist.
+                yield tree_path
+        except IOError as e:
+            directory_changed = has_directory_changed(e, tree_path, entry)
+            if directory_changed:
+                yield tree_path
+            else:
+                raise
         else:
             if blob.id != entry.sha:
                 yield tree_path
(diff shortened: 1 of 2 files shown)


Lothiraldan:
@jelmer I'm not a big fan of the new control flow in this PR but the tests are passing, so it's a good starting point.

Also I wasn't able to figure out how to test the submodule check in `get_unstaged_changes`, does dulwich have helpers to create and track submodules?

dulwich/index.py
```diff
@@ -604,6 +604,32 @@ def read_submodule_head(path):
         return None
 
 
+def has_directory_changed(error, tree_path, entry):
+    """ When handling an error trying to create a blob from a path, call this
```
jelmer: Please start docstrings with a single line summarizing what the function does, for e.g. sphinx.

dulwich/index.py
```diff
@@ -625,23 +651,21 @@ def get_unstaged_changes(index, root_path, filter_blob_callback=None):
             if filter_blob_callback is not None:
                 blob = filter_blob_callback(blob, tree_path)
         except OSError as e:
```
jelmer: Can't we just change OSError here to (IOError, OSError)?

jelmer:
> Also I wasn't able to figure out how to test the submodule check in `get_unstaged_changes`, does dulwich have helpers to create and track submodules?
We could probably do with some helper functions for creating submodules for use in tests. Until then, the best way to create a submodule is something like what is done in test_index: https://github.com/dulwich/dulwich/blob/9599545b541a85fc82c26ebb520805f5b44b1c16/dulwich/tests/test_index.py#L538

dulwich/index.py
Lothiraldan: I have updated the code, thanks for the idea

brutus:
It seems to me, that _dulwich_ still stumbles over submodules when listing unstaged changes. Or am I missing something here?

## Example

If I setup a repo like this:

```shell
rm -rf dulwich && mkdir dulwich && cd dulwich && \
pipenv --three install dulwich && \
echo 'foo' > README.rst && \
git init && git add . && git commit -m init
```

And run

```shell
pipenv run python -c 'import dulwich.repo as repo; import dulwich.index as index; print(list(index.get_unstaged_changes(repo.Repo(".").open_index(), b"./")))'
```

I get `[]` as result; and running it again after

```shell
echo 'bar' > README.rst
```

results in `[b'README.rst']`.

### With Submodule

But if I use a sub module:

```shell
git submodule add https://github.com/pypa/pipenv.git somesubmodule
```

And run it one more time, it results in:

```
  File "<string>", line 1, in <module>
  File "/home/brutus/.local/share/virtualenvs/dulwich-zMWvdkda/lib/python3.7/site-packages/dulwich/index.py", line 622, in get_unstaged_changes
    full_path, os.lstat(full_path)
  File "/home/brutus/.local/share/virtualenvs/dulwich-zMWvdkda/lib/python3.7/site-packages/dulwich/index.py", line 571, in blob_from_path_and_stat
    with open(fs_path, 'rb') as f:
IsADirectoryError: [Errno 21] Is a directory: b'./somesubmodule'

```

### Using the Patch

If I apply the changes from <https://github.com/dulwich/dulwich/pull/688>, I still get:

```
  File "<string>", line 1, in <module>
  File "/home/brutus/.local/share/virtualenvs/dulwich-zMWvdkda/lib/python3.7/site-packages/dulwich/index.py", line 650, in get_unstaged_changes
    full_path, os.lstat(full_path)
  File "/home/brutus/.local/share/virtualenvs/dulwich-zMWvdkda/lib/python3.7/site-packages/dulwich/index.py", line 571, in blob_from_path_and_stat
    with open(fs_path, 'rb') as f:
IsADirectoryError: [Errno 21] Is a directory: b'./somesubmodule'
```

## Possible Cause

I'm not familiar with _dulwich_ yet and only glimpsed over the code, but shouldn't we ignore unchanged sub-modules, instead of re-raising the `IsADirectoryError` exception? If I  add this part:

```python
elif e.errno == errno.EISDIR:
	# The file is an unchanged directory
	pass
```

it works in this simple test case…

Lothiraldan:
@brutus Could you test with dulwich master? I'm not sure how you applied the changes from this pull-request but I cannot reproduce with dulwich master.

brutus:
Doh m) _master_ runs fine. Thanks!

PR #696, opened 2019-04-01
Update git diff format

Lothiraldan:
The goal of this PR is to make dulwich generate diff compatible with `git apply`.

### DIFF
--- a/dulwich/patch.py
+++ b/dulwich/patch.py
@@ -185,8 +185,8 @@ def write_object_diff(f, store, old_file, new_file, diff_binary=False):
     """
     (old_path, old_mode, old_id) = old_file
     (new_path, new_mode, new_id) = new_file
-    old_path = patch_filename(old_path, b"a")
-    new_path = patch_filename(new_path, b"b")
+    patched_old_path = patch_filename(old_path, b"a")
+    patched_new_path = patch_filename(new_path, b"b")
 
     def content(mode, hexsha):
         if hexsha is None:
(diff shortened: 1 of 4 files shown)


dulwich/patch.py
```diff
@@ -207,11 +207,17 @@ def lines(content):
     new_content = content(new_mode, new_id)
     if not diff_binary and (
             is_binary(old_content.data) or is_binary(new_content.data)):
-        f.write(b"Binary files " + old_path + b" and " + new_path +
-                b" differ\n")
+        binary_diff = (
```
jelmer: Please keep changes minimal, there shouldn't be a need to reflow these lines in a style that's different from the surrounding code.

dulwich/tests/test_porcelain.py
```diff
@@ -702,6 +706,81 @@ def test_empty(self):
...
+        git_command = ["-C", copy_path, "apply", patch_path]
+        run_git_or_fail(git_command)
+
+        # And now check that the files contents are exactly the same between
+        # the two repositories
+        original_files = set(os.listdir(self.repo_path))
+        new_files = set(os.listdir(copy_path))
+
+        # Check that we have the exact same files in both repositories
+        assert original_files == new_files
```
jelmer: Please use self.assertEqual rather than plain assert; the latter gets stripped out in -O2 and it's inconsistent with the rest of the Dulwich test code.

dulwich/tests/test_porcelain.py
```diff
@@ -702,6 +706,81 @@ def test_empty(self):
                             outstream=outstream)
         self.assertEqual(outstream.getvalue(), b"")
 
+    def test_diff_apply(self):
```
jelmer: This test is verifying compatibility with C git; please move it somewhere under dulwich/tests/compat. Tests not under compat/ should not require git itself to be installed.

Lothiraldan:
@jelmer I have update the PR and squashed my changes

PR #700, opened 2019-04-03
ignore: question mark should not match slash

pared:
According to [git docs](https://github.com/git/git/blob/041f5ea1cf987a4068ef5f39ba0a09be85952064/Documentation/gitignore.txt#L106)
`?` should not match `/`


jelmer:
Merged, thanks!

PR #703, opened 2019-04-11
Rename submodule into subproject in diff to match CGit behavior

Lothiraldan:
When generating a diff in a repository with dirty submodules, CGit is using
the subproject terminology. I couldn't find the reasoning why or since when it
has been doing it but https://stackoverflow.com/q/4873980/2996250 seems to
indicate it has been the case since 2011 at least.

It has no real impacts as submodules information in diff are mostly ignored by
`git apply` but this makes Dulwich behavior closer to CGit one.

### DIFF
--- a/dulwich/patch.py
+++ b/dulwich/patch.py
@@ -192,7 +192,7 @@ def write_object_diff(f, store, old_file, new_file, diff_binary=False):
         if hexsha is None:
             return Blob.from_string(b'')
         elif S_ISGITLINK(mode):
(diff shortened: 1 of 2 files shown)


Lothiraldan:
@jelmer This PR seems to have been merged but the code on master still uses the old terminology: https://github.com/dulwich/dulwich/blob/master/dulwich/patch.py#L201. Is that expected? I don't find traces of the commit being reverted

jelmer:
Huh, yes - that's weird. Cherry-picked (again?) now.

PR #704, opened 2019-04-15
Fix line-ending conversion support

### LINKED ISSUE #693: Files with CRLF and git configured to convert those to LF leads to incorrect porcelain status
If there is a file with Windows newlines (CRLF) and git is set to automatically convert this to unix-style (LF), which is enabled by default (`git config core.autocrlf` option),
```porcelain.status``` with show that that tree is dirty thought no local changes have been made.

[Repo with code to reproduce](https://github.com/AndrewLvov/test_porcelain)

### PR DESCRIPTION
**Pull request by Lothiraldan**
The first support of line-ending conversion was too aggressive. After further
manual testing and investigation of
https://github.com/dulwich/dulwich/issues/693, I realized that line-ending
conversions doesn't apply on every checkin or checkout operation, only on
checkin or checkout operation resulting in a new file being created.

CGit actually tells you when it's applying the line-ending normalization with
a message that looks like: `warning: CRLF will be replaced by LF in BRANCH2.`

It should hopefully fix #693

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -629,16 +629,27 @@ def _has_directory_changed(tree_path, entry):
     return False
 
 
-def get_unstaged_changes(index, root_path, filter_blob_callback=None):
+def get_unstaged_changes(index, repo, filter_blob_callback=None):
     """Walk through an index and check for differences against working tree.
 
     :param index: index to check
     :param root_path: path in which to find files
     :return: iterator over paths with unstaged changes
     """
+    root_path = repo.path
     # For each entry in the index check the sha1 & ensure not staged
     if not isinstance(root_path, bytes):
         root_path = root_path.encode(sys.getfilesystemencoding())
+
+    head_tree = None
+    lookup_obj = repo.object_store.__getitem__
+
+    try:
+        head_tree = repo[repo[b"HEAD"].tree]
+    except KeyError:
+        # In case we don't have a HEAD refs, the line-ending filters will
+        # be executed
+        pass
 
     for tree_path, entry in index.iteritems():
         full_path = _tree_to_fs_path(root_path, tree_path)
@@ -652,7 +663,19 @@ def get_unstaged_changes(index, root_path, filter_blob_callback=None):
             blob = blob_from_path_and_stat(full_path, st)
 
             if filter_blob_callback is not None:
-                blob = filter_blob_callback(blob, tree_path)
+
+                if not head_tree:
+                    new_file = True
+                else:
+                    try:
+                        head_tree.lookup_path(lookup_obj, tree_path)
+                        new_file = False
+                    except KeyError:
+                        # Line-ending conversion is done only for files
+                        # not in store yet
+                        new_file = True
+
+                blob = filter_blob_callback(blob, tree_path, new_file)
         except EnvironmentError as e:
             if e.errno == errno.ENOENT:
                 # The file was removed, so we assume that counts as
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -641,16 +641,6 @@ def get_unstaged_changes(index, repo, filter_blob_callback=None):
     if not isinstance(root_path, bytes):
         root_path = root_path.encode(sys.getfilesystemencoding())
 
-    head_tree = None
-    lookup_obj = repo.object_store.__getitem__
-
-    try:
-        head_tree = repo[repo[b"HEAD"].tree]
-    except KeyError:
-        # In case we don't have a HEAD refs, the line-ending filters will
-        # be executed
-        pass
-
     for tree_path, entry in index.iteritems():
         full_path = _tree_to_fs_path(root_path, tree_path)
         try:
@@ -663,17 +653,12 @@ def get_unstaged_changes(index, repo, filter_blob_callback=None):
             blob = blob_from_path_and_stat(full_path, st)
 
             if filter_blob_callback is not None:
-
-                if not head_tree:
+                # Check if the file is already in the index
+                try:
+                    index[tree_path]
+                    new_file = False
+                except KeyError:
                     new_file = True
-                else:
-                    try:
-                        head_tree.lookup_path(lookup_obj, tree_path)
-                        new_file = False
-                    except KeyError:
-                        # Line-ending conversion is done only for files
-                        # not in store yet
-                        new_file = True
 
                 blob = filter_blob_callback(blob, tree_path, new_file)
         except EnvironmentError as e:
--- a/dulwich/line_ending.py
+++ b/dulwich/line_ending.py
@@ -31,6 +31,9 @@ The normalization is a two-fold process that happens at two moments:
   when doing a `git add` call. We call this process the write filter in this
   module.
 
+The normalization only happens when the resulting file does not exists yet.
+For the write filter, they are files that are shown as added in status.
+
 One thing to know is that Git does line-ending normalization only on text
 files. How does Git know that a file is text? We can either mark a file as a
 text file, a binary file or ask Git to automatically decides. Git has an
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by Lothiraldan** [comment]
@jelmer It's not fully ready yet, I need to do more manual testing and think about possible edge-cases but if you have time, I would like your opinion on the API changes.

**Comment by Lothiraldan** [comment]
Wait, I just understood that it's only the `input` mode of `autocrlf` which is converting on new files only, not `autocrlf=true`, I need to update this PR.

File: dulwich/index.py [hunk 1]
```diff
@@ -629,17 +629,28 @@ def _has_directory_changed(tree_path, entry):
     return False
 
 
-def get_unstaged_changes(index, root_path, filter_blob_callback=None):
+def get_unstaged_changes(index, repo, filter_blob_callback=None):
     """Walk through an index and check for differences against working tree.
 
     :param index: index to check
     :param root_path: path in which to find files
```
AndrewLvov: Name change not reflected in docstring

dulwich/tests/test_porcelain.py
```diff
@@ -1200,10 +1200,51 @@ def test_status_crlf_convert(self):
...
+        # the staging area
         results = porcelain.status(self.repo)
         self.assertDictEqual(
             {'add': [], 'delete': [], 'modify': []},
             results.staged)
+        self.assertListEqual(results.unstaged, [b"crlf"])
+        self.assertListEqual(results.untracked, [])
+
+    def test_status_crlf_convert(self):
+        # With an emty repo
```
AndrewLvov: empty*

dulwich/index.py
```diff
@@ -629,17 +629,28 @@ def _has_directory_changed(tree_path, entry):
     return False
 
 
-def get_unstaged_changes(index, root_path, filter_blob_callback=None):
+def get_unstaged_changes(index, repo, filter_blob_callback=None):
```
jelmer: It seems like a layer violation to pass in a repo here. Do you really need the whole repo, or could you e.g. just pass in a different filter_blob_callback?

dulwich/line_ending.py
```diff
@@ -229,19 +232,29 @@ def __init__(self, config_stack, gitattributes):
             core_eol, core_autocrlf, self.gitattributes
         )
 
-    def checkin_normalize(self, blob, tree_path):
+    def checkin_normalize(self, blob, tree_path, new_file):
```
jelmer: Couldn't we just call checkin_normalize for new files only?

dulwich/line_ending.py
```diff
@@ -229,19 +232,29 @@ def __init__(self, config_stack, gitattributes):
...
                 blob, self.fallback_write_filter, binary_detection=True
             )
 
         return blob
 
-    def checkout_normalize(self, blob, tree_path):
+    def checkout_normalize(self, blob, tree_path, new_file):
         """ Normalize a blob during a checkout operation
         """
+        if not new_file:
```
jelmer: Same: couldn't we just call checkout_normalize for new files only?

dulwich/repo.py
```diff
@@ -1080,7 +1090,21 @@ def stage(self, fs_paths):
             else:
                 if not stat.S_ISDIR(st.st_mode):
                     blob = blob_from_path_and_stat(full_path, st)
-                    blob = blob_normalizer.checkin_normalize(blob, fs_path)
+
+                    if not head_tree:
+                        new_file = True
+                    else:
+                        try:
+                            head_tree.lookup_path(lookup_obj, tree_path)
```
jelmer: Are you sure git actually looks in HEAD? "git add" is supposed to be an index-only operation.

dulwich/line_ending.py
Lothiraldan: Places that calls `checkin_normalize` don't actually know that the callback is checkin_normalize. It was designed this way to support more filter callbacks in order to support Git custom filters in gitattributes.

dulwich/repo.py
Lothiraldan: I tried adding a new file with CRLF line-ending to the staging area and it was correctly converted into LF (with autocrlf input). Then I tried modifying it by adding new CRLF and when adding the new content, it also got converted.
When introducing new CRLF lines in an already tracked file, CGit seems to not run the line-ending conversion so I suspect CGit has a way to detect when it's a new file or not.
My guess is it either check the store on the current branch (I tried adding a file already tracked on another branch and it got converted) or it saves some extra information in the index (saying that file X is new) and discard this information after a commit. Do you have ideas how we could find how CGit is doing it?

dulwich/line_ending.py
jelmer: All filters should work at the same moments. See https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes

Lothiraldan: I'm not sure if line-endings are exactly like filters in CGit but they operate at about the same time

dulwich/repo.py
Lothiraldan: After more test, maybe CGit is comparing the line-endings between the working directory and the index and apply the conversion if the file in index does not contains CRLF and the file in working directory does, I will try to validate this hypothesis

Lothiraldan:
I have updated the PR by checking if the file is new in the index rather than if the file is new in the store.

There is still one case where it's failing, with the status command. If we add a file, the file will be converted as it's not in the index yet. But after, when calling status, the file is in the index. I'm not sure how to detect such case. Also trying to diff the file should returns nothing.

Do you have an idea how CGit might handle this case? I started digging in CGit test files and source code but didn't found the answer yet.

Lothiraldan:
So I tried testing CGit in many scenarios and I am more confused than ever, I don't understand when CGit decides to convert LF to CRLF and CRLF to LF. I have started a test file https://github.com/Lothiraldan/cgit-line-ending-conversions-tests/blob/master/test-line_ending.t and some of the scenarios doesn't make sense for me at all. For me, using only autocrlf, CGit should nevers convert LF to CRLF but it says otherwise: https://github.com/Lothiraldan/cgit-line-ending-conversions-tests/blob/master/test-line_ending.t#L417

@jelmer Do you know someone that could shed some light on this? If not I may know someone that could help.

jelmer:
On Tue, May 14, 2019 at 04:50:32AM -0700, Boris Feld wrote:
> So I tried testing CGit in many scenarios and I am more confused than ever, I don't understand when CGit decides to convert LF to CRLF and CRLF to LF. I have started a test file https://github.com/Lothiraldan/cgit-line-ending-conversions-tests/blob/master/test-line_ending.t and some of the scenarios doesn't make sense for me at all. For me, using only autocrlf, CGit should nevers convert LF to CRLF but it says otherwise: https://github.com/Lothiraldan/cgit-line-ending-conversions-tests/blob/master/test-line_ending.t#L417
> 
> @jelmer Do you know someone that could shed some light on this? If not I may know someone that could help.
Sorry, I've been traveling. I'll try and have a look at it this weekend.

jelmer:
Sorry, still traveling - but I haven't forgotten about this.

Lothiraldan:
@jelmer how can we make this PR go forward? Will you have time to take a look at it or do you know anybody that could help me catch what I am missing in the line-ending conversion support?

jelmer:
Sorry, I'll take another look this week.

jelmer:
I think the git config manpage shines the most light on how these options work. 

*core.autocrlf*:

           Setting this variable to "true" is the same as setting the text attribute to "auto" on all files and core.eol to "crlf". Set to true
           if you want to have CRLF line endings in your working directory and the repository has LF line endings. This variable can be set to
           input, in which case no output conversion is performed.

*core.eol*:

           Sets the line ending type to use in the working directory for files that are marked as text (either by having the text attribute set,
           or by having text=auto and Git auto-detecting the contents as text). Alternatives are lf, crlf and native, which uses the platform’s
           native line ending. The default value is native. See gitattributes(5) for more information on end-of-line conversion. Note that this
           value is ignored if core.autocrlf is set to true or input.

pmrowla:
@jelmer this PR can be closed (replaced by #867)

Issue #222, opened 2014-09-15
Binary wheels for Windows


dahlia:
Although Dulwich provides `setup.py install --pure` option, the option cannot be enabled when it’s installed as a dependency of `install_requires`, or using `pip install`.  On Windows which doesn’t have Visual Studio Dulwich installation fails without `--pure` option.

So it would be great if Dulwich provides binary [wheels](http://pythonwheels.com/) for Windows.

jelmer:
Any news on this?

garyvdm:
Actually started on this today. Will probably have a patch ready tomorrow.

ankostis:
Having *wheels* autogenerated would be much appreciated.
It is possible to automate *Appveyor* to use [*twine*](https://pypi.org/project/twine/) to push wheels into PyPi, with an `appveyor.yml` like this:

```yaml
environment:
  PYPI_USER: "FOO"
  PYPI_PSWD:
    # See https://www.appveyor.com/docs/build-configuration/#secure-variables
    secure: <paste-your-encrypted-value>

#... regular build here

on_success:
  - ps: >-
      $cmtid = git rev-parse HEAD
      if (git describe --exact-match $cmtid) {
        twine upload -u PYPI_USER -p PYPI_PSWD dist/*
      }
```

Of course the situation becomes a bit more complex if gpg-signing is needed.
I could provide a PR but the user/password have to be set by the owner of the [PyPi account](https://pypi.org/project/dulwich/).

jelmer:
On Sun, Oct 02, 2016 at 04:18:11AM -0700, Kostis Anagnostopoulos wrote:
> Having *wheels* autogenerated would be much appreciated.
> It is possible to automate *Appveyor* to use [*twine*](https://pypi.org/project/twine/) to push wheels into PyPi, with an `appveyor.yml` like this:
> 
> ```yaml
> environment:
>   PYPI_USER: "FOO"
>   PYPI_PSWD:
>     # See https://www.appveyor.com/docs/build-configuration/#secure-variables
>     secure: <paste-your-encrypted-value>
> 
> #... regular build here
> 
> on_success:
>   - ps: >-
>       $cmtid = git rev-parse HEAD
>       if (git describe --exact-match $cmtid) {
>         twine upload -u PYPI_USER -p PYPI_PSWD dist/*
>       }
> ```
> 
> Of course the situation becomes a bit more complex if gpg-signing is needed.
[...]

PR #707, opened 2019-05-17
ci: deploy to pypi

### LINKED ISSUE #593: allow depending on "dulwich[pure]" to disable C extension build
There isn't an easy way to do this at the moment; you can only do this
using e.g. requirements.txt.

For setup.py, ideally we'd want to support something like
"dulwich[pure]" in your setup.py's install_requires, but I haven't
worked out the right magic to make that work in dulwich' setup.py yet. See
https://setuptools.readthedocs.io/en/latest/setuptools.html#id14 for
background on extras.

efiop:
- [x] make travis upload source and linux wheel to pypi;

- <s> make travis upload osx wheel to pypi; </s> (tests are failing on osx, so need to sort those out first, so we don't accidentally upload something that doesn't actually work on osx properly. Out of the scope for this PR.)

 - [x] make appveyor upload windows wheel to pypi;

Signed-off-by: Ruslan Kuprieiev <ruslan@iterative.ai>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,3 +47,18 @@ script:
 after_success:
   - python -m coverage combine
   - codecov
+
+deploy:
+  provider: pypi
+  user: dulwich-bot
+  password:
+    secure: e7DTu4CwOCARfN/mwA8lXQ==
+  skip_cleanup: true
+  skip_existing: true
+  file_glob: true
+  file:
+    - dist/dulwich*.whl
+    - dist/dulwich*.tar.gz
+  on:
+    tags: true
+    repo: dulwich/dulwich
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,7 @@
 environment:
 
-  PYPI_USER: "dulwich-bot"
-  PYPI_PSWD:
+  TWINE_USERNAME: "dulwich-bot"
+  TWINE_PASSWORD:
     # See https://www.appveyor.com/docs/build-configuration/#secure-variables
     secure: e7DTu4CwOCARfN/mwA8lXQ==
 
@@ -88,6 +88,10 @@ after_test:
   # - "python setup.py bdist_wininst"
   - "build.cmd %PYTHON%\\python.exe setup.py bdist_msi"
   - ps: "ls dist"
+
+deploy_script:
+  - if "%APPVEYOR_REPO_TAG%"=="true" pip install twine
+  - if "%APPVEYOR_REPO_TAG%"=="true" twine upload dist\dulwich-*.whl
 
 artifacts:
   - path: dist\*


efiop:
Hm, looks like some windows test builds have failed with
```
Installed c:\python34\lib\site-packages\certifi-2019.3.9-py3.4.egg
Searching for urllib3>=1.24.1
Reading https://pypi.python.org/simple/urllib3/
Best match: urllib3 1.25.2
Downloading https://files.pythonhosted.org/packages/9a/8b/ea6d2beb2da6e331e9857d0a60b79ed4f72dcbc4e2c7f2d2521b0480fda2/urllib3-1.25.2.tar.gz#sha256=a53063d8b9210a7bdec15e7b272776b9d42b2fd6816401a0d43006ad2f9902db
Processing urllib3-1.25.2.tar.gz
Writing C:\Users\appveyor\AppData\Local\Temp\1\easy_install-ef4rhdjc\urllib3-1.25.2\setup.cfg
Running urllib3-1.25.2\setup.py -q bdist_egg --dist-dir C:\Users\appveyor\AppData\Local\Temp\1\easy_install-ef4rhdjc\urllib3-1.25.2\egg-dist-tmp-exkzlmjg
C:\Python34\lib\distutils\dist.py:260: UserWarning: Unknown distribution option: 'project_urls'
  warnings.warn(msg)
C:\Python34\lib\site-packages\setuptools\dist.py:285: UserWarning: Normalizing '2019.03.09' to '2019.3.9'
  normalized_version,
C:\Python34\lib\distutils\dist.py:260: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
error: Setup script exited with error in urllib3 setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```
Doesn't seem related to my patch.

jelmer:
I'm not sure about this change - This makes it really easy to accidentally install Dulwich without C extensions, which results in much worse performance.

efiop:
Hi @jelmer !

Thank you so much for working on this amazing project! 🙂 

From https://github.com/dulwich/dulwich/issues/593#issuecomment-394774705
> I'd like the default to having C extensions enabled. If we do it the other way around, most people are going to end up with the slower version of Dulwich.

But this patch doesn't disable C extensions by default. It only allows dulwich to fallback to pure python implementation, so that anyone who is using dulwich as a dependency, is not forcing their users to explicitly use `--pure` for their project if they happen to not have a compiler installed.

If you still don't like this fallback approach, maybe we could find another way to make this more livable? As proposed in #593 , an obvious choice would be to create a separate package `dulwich-pure`. Since it would differ from `dulwich` package by a mere few lines of code in setup.py that could be enabled/disabled by something like an env `DULWICH_PURE=1` var, it shouldn't be too much hustle to maintain. As to discoverability, we could simply put a note about available `dulwich-pure` package in the same section that `--pure` is currently described in the README. What do you think?

Thanks,
Ruslan

ei-grad:
Wouldn't it be better to build the `manylinux`-wheels for dulwich?

jelmer:
On Fri, May 17, 2019 at 04:25:39AM -0700, Ruslan Kuprieiev wrote:
> Hi @jelmer !
> 
> Thank you so much for working on this amazing project! 🙂 
> 
> From https://github.com/dulwich/dulwich/issues/593#issuecomment-394774705
> > I'd like the default to having C extensions enabled. If we do it the other way around, most people are going to end up with the slower version of Dulwich.
> 
> But this patch doesn't disable C extensions by default. It only allows dulwich to fallback to pure python implementation, so that anyone who is using dulwich as a dependency, is not forcing their users to explicitly use `--pure` for their project if they happen to not have a compiler installed.
For some users, it will disable C extensions by default (if they don't
have a compiler). It should be an explicit choice by the user to 
run Dulwich in pure mode. 

Is there a particular reason why you need Dulwich without the C
extensions?

> If you still don't like this fallback approach, maybe we could find another way to make this more livable? As proposed in #593 , an obvious choice would be to create a separate package `dulwich-pure`. Since it would differ from `dulwich` package by a mere few lines of code in setup.py that could be enabled/disabled by something like an env `DULWICH_PURE=1` var, it shouldn't be too much hustle to maintain. As to discoverability, we could simply put a note about available `dulwich-pure` package in the same section that `--pure` is currently described in the README. What do you think?
> https://github.com/dulwich/dulwich/pull/707#issuecomment-493420048
Supporting an env variable may be an option - I'd be happy
to support that; as far as I know, that should not require a separate
package.

Jelmer

jelmer:
On Fri, May 17, 2019 at 05:11:06AM -0700, Andrew Grigorev wrote:
> Wouldn't it be better to build the `manylinux`-wheels for dulwich?
I'm open to other suggestions on how to deal with this, but unfamiliar
with manylinux-wheels - how would this work?

efiop:
@jelmer 

> For some users, it will disable C extensions by default (if they don't
have a compiler). It should be an explicit choice by the user to
run Dulwich in pure mode.

Agreed :+1: 

> Is there a particular reason why you need Dulwich without the C
extensions?

[Just some users trying to install dvc (which depends on dulwich) on systems without gcc :slightly_smiling_face:](https://github.com/iterative/dvc/issues/2010) Trying to provide the best experience for our users.

> Supporting an env variable may be an option - I'd be happy
to support that; as far as I know, that should not require a separate
package.

Perfect! I'll update my patch ASAP. Thanks a lot for your help! :slightly_smiling_face:

efiop:
> Supporting an env variable may be an option - I'd be happy
to support that; as far as I know, that should not require a separate
package.

@jelmer  Ah, sorry, I've misread that :) I meant that that env option would make setup.py not include extensions into `setup()` and will change name from `dulwich` to `dulwich-pure`, so when you are building and uploading to pypi you would do something like:
```
# build dulwich pkgs
python setup.py sdist bdist_wheel --universal

# build dulwich-pure pkgs
DULWICH_PURE=1 python setup.py sdist bdist_wheel --universal

# upload both pkgs to pypa
twine upload dist/*
```
What do you think about it?

jelmer:
Adding a separate dulwich-pure project is problematic for a couple of reasons:

* users can no longer choose to install the fast version of Dulwich; they're stuck with pure Python version
* it doesn't solve the problem for applications/libraries that depend on Dulwich don't depend on the pure-Python version. Users are still stuck having to install gcc
* other applications installed in the same environment also get the slow version

If the problem is just that installing gcc is too complicated, then I think the best path forward is to provide wheels so users don't have to install gcc.

efiop:
@jelmer Agreed. Building wheels wasn't my first choice because you have `pure` option already and I find bulding wheels more complex, since, ideally, you'll have to provide them for each os: Windows, mac and linux (but manylinux mentioned by @ei-grad would at least make it a little easier for linux) and for each arch. So your release workflow will no longer be as simple as:
```
python setup.py sdist
twine upload dist/*
```
But on the other hand, we could setup appveyor and travis to handle a lot of that for us once and for all. Let's do that then. How would you like us to proceed with it? Should we setup CIs to auto-deploy to pypi on tags? Or would you like to do that manually? I might be missing something, but it seems like you are currently releasing manually, right?

jelmer:
I'd rather avoid manual building on multiple platforms, that would make the release process significantly more complicated. 

Help setting up CIs to autodeploy to pypi on tags would be great. See also #222, which is specifically about having binary wheels on Windows.

efiop:
@jelmer Got it. So I'll need a few encrypted values from you. Please go to your local clone of dulwich repo and run
```
travis encrypt $PYPI_USERNAME
travis encrypt $PYPI_PASSWORD
```
and then go to appveyor->account->ecrypt yaml and encrypt your username and password.

efiop:
Btw, @jelmer is dulwich meant to run on osx? I've enabled osx build on travis, but tests show a lot of errors. Just wondering if that are simply some tests that are not meant to run on osx.

Also, what is up with half of the windows builds failing?

efiop:
Can also confirm that `make check` is failing because of multiple tests on my mac. So looks like tests were never run on mac before :slightly_frowning_face:

jelmer:
Thanks!

I'm pretty sure the tests did pass on Mac OS X at one point. Some tests have probably regressed since, since I don't have access to a Mac OS X machine and thus I won't notice if a change breaks Mac OS X but not Linux. Hopefully having CI for Mac OS X will make it harder for new regressions to be introduced in the future once we fix them...

I've set $TWINE_USER / $TWINE_PASSWORD earlier for pypi uploads; see also https://github.com/dulwich/dulwich/pull/701/files

efiop:
@jelmer Thanks! Will use that for appveyor build. How about travis-encrypted password then? I don't think travis and appveyor encrypted values are compatible, are they?

jelmer:
It is meant to run on OSX - the tests did pass at one point, but they have probably regressed since I don't have access to a OSX machine.

Having continuous CI for OSX will hopefully prevent them from regressing again.

efiop:
@jelmer and how about travis-encrypted password? The PR you've linked is appveyor-only.

efiop:
@jelmer I guess osx upload will have to wait until tests are fixed. Please take a look at this PR.

efiop:
@jelmer ?

efiop:
Btw @jelmer , forgot to clarify that we are only using `dulwich/ignore.py` in our project, without any need for C extensions, so our users won't suffer from that. Maybe we could have both automatic fallback and wheels? So that it works no matter what? Or maybe dulwich-pure is not a bad idea for that purpose?

jelmer:
@efiop Is there a particular reason why wheels wouldn't work in your case? I'm trying to avoid creating a dulwich-pure if I at all can, since I'm pretty sure it's going to lead to issues down the line as people will start depending on it.

jelmer:
Merged, thanks. I'll try to kick off builds and fix the password variables some time this week.

efiop:
> @efiop Is there a particular reason why wheels wouldn't work in your case? I'm trying to avoid creating a dulwich-pure if I at all can, since I'm pretty sure it's going to lead to issues down the line as people will start depending on it.

Not really, I'm just thinking that wheels will take a bit longer time to make them build for every platform, considering that osx needs to be fixed as well. I'm willing to go either way. Let's try wheels for now :slightly_smiling_face:

> Merged, thanks. I'll try to kick off builds and fix the password variables some time this week.

Thank you! :slightly_smiling_face: Please ping me when builds are ready or if there is anything I can help with to make them work. I'll try to fix tests on osx this weekend or next week so that we could enable wheel deployment for it as well.

Thanks again for working on this amazing project! :slightly_smiling_face:

jelmer:
ping @efiop 

The appveyor / .travis.yml passwords are now in place.  

There are still some open issues building wheels on the individual platforms:

* #222: urllib3 fails to install with Python3 on Windows
* #710: need to build manylinux wheels on Linux
* #711: build wheels on Mac OS X

efiop:
Thanks @jelmer ! I will take a look this week. Cheers! :slightly_smiling_face:

efiop:
Sorry for the delay @jelmer , have more important tasks on my plate right now, will try to get back to helping out with this in the upcoming weekends.

efiop:
Again, sorry for the delay. We've modified the part of our code that was using dulwich and we no longer need it :( So that moves the priority for those patches for me to very low for now 🙁

jelmer:
No worries. Looks like you just moved to GitPython?

efiop:
@jelmer we've used both actually :) So we were using dulwich for dvcignore implementation, but with the new developments we've moved to another library instead of hacking bits from dulwich 🙂We are still considering abandoning gitpython because of the resource leaks, and will be choosing between pygit and dulwich, so if we chose the latter, we will probably come back to help with builds and stuff 🙂 Thank you so much for working on this amazing project!

PR #724, opened 2019-09-17
Optimize idempotent ref file write

manueljacob:
The advantage is that it saves one `fsync()` if the content wouldn’t change.

The disadvantage is that it introduces another read from the ref file. However, the file is already read shortly before and very likely to be in the file system cache.

### DIFF
--- a/dulwich/refs.py
+++ b/dulwich/refs.py
@@ -717,8 +717,7 @@ class DiskRefsContainer(RefsContainer):
                 # read again while holding the lock
                 orig_ref = self.read_loose_ref(realname)
                 if orig_ref is None:
-                    orig_ref = self.get_packed_refs().get(
-                            realname, ZERO_SHA)
+                    orig_ref = self.get_packed_refs().get(realname, ZERO_SHA)
             except (OSError, IOError):
                 f.abort()
                 raise
--- a/dulwich/refs.py
+++ b/dulwich/refs.py
@@ -721,6 +721,9 @@ class DiskRefsContainer(RefsContainer):
             except (OSError, IOError):
                 f.abort()
                 raise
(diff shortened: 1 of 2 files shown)


jelmer:
I would much rather have callers do the read and then skip setting of the ref if nothing has changed, rather than require a read for every ref change. This is likely to slow down other callers, in particular when a packed-refs file is present.

dulwich/refs.py
```diff
@@ -713,19 +713,21 @@ def set_if_equals(self, name, old_ref, new_ref, committer=None,
 
         ensure_dir_exists(os.path.dirname(filename))
         with GitFile(filename, 'wb') as f:
+            try:
+                # read again while holding the lock
+                orig_ref = self.read_loose_ref(realname)
+                if orig_ref is None:
```
jelmer: There is no lock on the packed refs, how does this ensure that packed refs isn't changing at the same time? Since it will no longer write the new ref file.

jelmer:
Closing this for now, since it's stale. Very interested in performance improvements in general, but we can't sacrifice correctness for performance. If you have more updates for this PR, please reopen.

manueljacob:
I’m considering to re-visit this.

> This isn't just an optimization, you're changing the behaviour of the code.
> 
> If the ref is present in packed refs with the value you're setting, this makes it no longer write a loose ref file.

Does that change the refs for commands using them later? If not, in which case is not writing a loose ref file a bad thing?

jelmer:
> Does that change the refs for commands using them later? If not, in which case is not writing a loose ref file a bad thing?
What if somebody else was trying to write that ref at the same time and had it locked? 

(I don't think these issues are fixable FWIW, but it's tricky to get right)

PR #728, opened 2019-10-07
Handle Windows path separators

### LINKED ISSUE #727: Path handling incorrect on Windows
When looking up paths, the current code expects paths to be splittable on '/':

https://www.dulwich.io/code/dulwich/blob/master/dulwich/objects.py#L-1075

This causes path lookup failures when run on Windows systems where '\' is the path separator.

stmcginnis:
This updates a few spots where assuming path separators of '/' results in the failure to find files.

Closes #727

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -25,6 +25,7 @@
 from io import BytesIO
 import errno
 import os
+import os.path
 import stat
 import sys
 import tempfile
@@ -1265,7 +1266,7 @@ def commit_tree_changes(object_store, tree, changes):
     nested_changes = {}
     for (path, new_mode, new_sha) in changes:
         try:
-            (dirname, subpath) = path.split(b'/', 1)
+            (dirname, subpath) = path.split(str.encode(os.path.sep), 1)
         except ValueError:
             if new_sha is None:
                 del tree[path]
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -25,6 +25,7 @@ import binascii
 from io import BytesIO
 from collections import namedtuple
 import os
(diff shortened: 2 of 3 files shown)


dulwich/object_store.py
```diff
@@ -1265,7 +1266,7 @@ def commit_tree_changes(object_store, tree, changes):
     nested_changes = {}
     for (path, new_mode, new_sha) in changes:
         try:
-            (dirname, subpath) = path.split(b'/', 1)
```
jelmer: Paths passed in here should always be using / as a separator; it's the responsibility of the caller to convert them to git-style paths (with a /)

dulwich/objects.py
```diff
@@ -1072,7 +1073,7 @@ def lookup_path(self, lookup_obj, path):
         :param path: Path to lookup
         :return: A tuple of (mode, SHA) of the resulting path.
         """
-        parts = path.split(b'/')
+        parts = path.split(str.encode(os.path.sep))
```
jelmer: Paths here are always git-style paths with a / - it's the responsibility of the caller to convert them.

stmcginnis:
Unfortunate that each consumer needs to know this and the library can't just handle it, but so be it.

jelmer:
This is not a matter of Dulwich not handling it. Backslash is a valid character in Git filenames.

This patch makes it impossible to access such filenames on windows.

The correct fix is for functions that convert filesystem paths to Git paths to also translate os.sep to b'/'. Patches that do that would be great.

jelmer:
This change also breaks any existing code that relies on being able to pass in a /.

PR #737, opened 2020-01-15
Fix #691 by supporting a password manager argument

### LINKED ISSUE #691: HTTP auth with urllib3 does not prompt for credentials
Since 0.19.0, pushing to a repo with HTTP authentication seems not to work, because it fails with HTTP 401 rather than prompting me for credentials. (I am using dulwich mostly via hg-git version 0.8.12.)

With versions prior to 0.19.0, I see something like the following when pushing to one of my repos.

```
$ hg push git+https://github.com/stephenrkell/liballocs.git
pushing to git+https://github.com/stephenrkell/liballocs.git
http authorization required for https://github.com/stephenrkell/liballocs.git/info/refs
realm: GitHub
user: stephenrkell
password: 
```
whereas from 0.19.0 onwards, I get the following.
```
$ hg push git+https://github.com/stephenrkell/liballocs.git
pushing to git+https://github.com/stephenrkell/liballocs.git
abort: git remote error: unexpected http resp 401 for https://github.com/stephenrkell/liballocs.git/info/refs?service=git-receive-pack
```
As an experiment, I backed out 465addec72ee and 0c99ef824c74 (sorry, Mercurial hashes :-) ) and the problem seems to go away.

Let me know if I can help debug further....

byteit101:
If a password manager is provided, retry requests with auth if they fail with 401 errors.
This still requires some changes on hg-git's side to get hg-git working.

I wasn't quite sure how to organize it or what names were best, so let me know if you have suggestions.

`password_mgr` must be an object that has a function of signature `find_user_password(realm, origin) -> (username, password)`. Not sure where the best place to document this is

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1502,10 +1502,11 @@ def default_urllib3_manager(config, **override_kwargs):
 class HttpGitClient(GitClient):
 
     def __init__(self, base_url, dumb=None, pool_manager=None, config=None,
-                 username=None, password=None, **kwargs):
+                 username=None, password=None, password_mgr=None, **kwargs):
         self._base_url = base_url.rstrip("/") + "/"
         self._username = username
         self._password = password
+        self._password_mgr = password_mgr
         self.dumb = dumb
 
         if pool_manager is None:
@@ -1514,12 +1515,7 @@ class HttpGitClient(GitClient):
             self.pool_manager = pool_manager
 
         if username is not None:
-            # No escaping needed: ":" is not allowed in username:
-            # https://tools.ietf.org/html/rfc2617#section-2
-            credentials = "%s:%s" % (username, password)
-            import urllib3.util
-            basic_auth = urllib3.util.make_headers(basic_auth=credentials)
-            self.pool_manager.headers.update(basic_auth)
+            self._update_auth()
 
         GitClient.__init__(self, **kwargs)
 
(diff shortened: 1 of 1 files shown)


dulwich/client.py
```diff
@@ -1588,6 +1592,31 @@ def _http_request(self, url, headers=None, data=None,
...
+            import urllib3.util
+            current_url = urllib3.util.parse_url(resp.geturl())
+            # Create the http origin for the password manager
+            safe_origin = str(urllib3.util.Url(scheme=current_url.scheme,
+                                           host=current_url.host,
+                                           port=current_url.port).url)
+            # Extract the realm (if it exists)
+            realm = None
+            try:
+                realm = resp.headers["WWW-Authenticawte"]
```
jelmer: This looks like it can't work

dulwich/client.py
```diff
@@ -1588,6 +1592,31 @@ def _http_request(self, url, headers=None, data=None,
...
+                if realm is not None:
+                    realm = realm[13:-1] # Basic realm=".."
+            except KeyError:
+                pass
+
+            self._username, self._password = \
+                self._password_mgr.find_user_password(realm, safe_origin)
```
jelmer: This breaks if find_user_password was not specified.

jelmer:
Would it make more sense to just have HttpGitClient raise an exception when it gets a 401 so the caller can call it again with credentials? That way, the caller can also support other kinds of authentication like negotiate, without having to add support in Dulwich.

This adds quite a bit of untested code, and it will only work for basic auth. The existing (basic-specific) code exists as a compromise because it allows the "password" argument to work for both the HTTP and SSH transports.

dulwich/client.py
byteit101: Oh whoops, looks like a save key slip.

byteit101:
I created a patch for hg-git that does the same, so I think this PR can probably be closed

https://groups.google.com/forum/#!topic/hg-git/MRzj-CihO9g

jelmer:
Ack, that looks like a much better way of doing this. Thanks, closing this PR.

PR #738, opened 2020-01-27
SendPackError exceptions have no message attribute

brechtm:
The message is stored as the first element of the args attribute.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -917,9 +917,9 @@ def push(repo, remote_location, refspecs,
             errstream.write(
                 b"Push to " + remote_location_bytes + b" successful.\n")
         except (UpdateRefsError, SendPackError) as e:
+            message, = e.args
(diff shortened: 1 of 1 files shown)


dulwich/porcelain.py
```diff
@@ -917,9 +917,9 @@ def update_refs(refs):
             errstream.write(
                 b"Push to " + remote_location_bytes + b" successful.\n")
         except (UpdateRefsError, SendPackError) as e:
+            message, = e.args
```
jelmer: Handling for UpdateRefsError and SendPackError probably needs to be split out here.

UpdateRefsError has more than 1 element in e.args, and e.args[0] is a unicode string.

brechtm:
Sorry for the very basic nature of this patch. It was made without giving it much though, merely to support issue #739.

jelmer:
Ack, thanks for reporting this - fixed in master.

PR #741, opened 2020-02-01
Feature: Pull remote url

sijis:
Lookup remote origin url when using `porcelain.pull()`.

This should allow `porcelain.pull()` work when git repository has a remote origin defined.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -940,9 +940,12 @@ def pull(repo, remote_location=None, refspecs=None,
     # Open the repo
     with open_repo_closing(repo) as r:
         if remote_location is None:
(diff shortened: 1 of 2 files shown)


sijis:
@jelmer Thanks for merging.
I noticed you added another TODO regarding [getting the config](https://github.com/dulwich/dulwich/blob/master/dulwich/porcelain.py#L943-L946).

I suspect you are wanting to look if there is a section that starts with 'remote'?

I had initially looked for ways to use `config.get()` but its always required a set as the key. I eventually settled with using `has_section()`.

Is there another way you can think of know if `remote` exists? I'm willing to address this properly.

Issue #747, opened 2020-02-28
Shallow fetches are broken


xavierd:
In the following code, depth can be None (due to the `or` condition above), which will send "deepen None" to the server. A regular git server will choke on this and fail the pull.

https://github.com/dulwich/dulwich/blob/24dcd78c8997539c067f09507bc2bff6529de666/dulwich/client.py#L644

PR #749, opened 2020-03-03
Don't send a depth of 0 on a shallow fetch

xavierd:
The git server code that parse the deepen packet (https://github.com/git/git/blob/2f268890c2cd2f115424936bbee27f8269080e5e/upload-pack.c#L811) will fail if the deepen is set to 0, special case this too to not send it.

This is the second fix for #747


jelmer:
I've reverted this, since it breaks the tests. This should be raising an exception if depth is set but shallow is not present, rather than silently skipping shallow.

Issue #702, opened 2019-04-08
Python error when cloning: AttributeError: 'HTTPResponse' object has no attribute 'geturl'


plafosse:
I'm getting the following error which seems to have been introduced by https://github.com/dulwich/dulwich/commit/5148c0c22d8a35ad094b431b116a35b4fa41cba0#diff-28cd0777376826bef5605561210673e2
```/usr/local/lib/python3.7/site-packages/dulwich-0.19.12-py3.7.egg/dulwich/client.py in _http_request(self, url, headers, data, allow_compression)
   1518
   1519         resp.content_type = resp.getheader("Content-Type")
-> 1520         resp_url = resp.geturl()
   1521         resp.redirect_location = resp_url if resp_url != url else ''
   1522

AttributeError: 'HTTPResponse' object has no attribute 'geturl'```

jelmer:
You need a newer version of urllib3; older versions don't have geturl()

plafosse:
Thanks for the response, adding a version check might be handy. I googled that error and couldn't find anything.

jelmer:
There is a version check in setup.py; I'm open to patches adding a run-time check as well, though we don't really do that for our other dependencies either.

havatv:
After the "merge" of #750, earlier versions of urllib3 than 1.23 are supported again, meaning that dulwich can be used on Ubuntu 18.04.

PR #750, opened 2020-03-03
Fallback to get_redirect_location() for urrlib3 versions lower than 1.23

havatv:
``geturl()`` was introduced for ***HTTPResponse*** in versions 1.23 of ***urllib3***.
This PR will allow a fallback to ``get_redirect_location()`` for earlier *urllib3* versions.

Reason: We would like the *QGIS Resource Sharing* plugin to include a recent release of dulwich.
It enables resource sharing in QGIS, and should work for the most popular platforms, including Ubuntu 18.04 (Bionic) that ships with *urrlib3* version 1.22.1 (https://packages.ubuntu.com/bionic/python3-urllib3).

See also #702 (``Python error when cloning: AttributeError: 'HTTPResponse' object has no attribute 'geturl'``)

Backporting should be fine, but I don't know if it is relevant.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1603,7 +1603,13 @@ class HttpGitClient(GitClient):
         read = BytesIO(resp.data).read
 
         resp.content_type = resp.getheader("Content-Type")
-        resp_url = resp.geturl()
+        resp_url = ''
+        # Check if geturl() is available (urllib3 version >= 1.23)
+        try:
+            resp_url = resp.geturl()
+        except AttributeError:
+            # get_redirect_location() is available for urllib3 >= 1.1
+            resp_url = resp.get_redirect_location()
         resp.redirect_location = resp_url if resp_url != url else ''
 
         return resp, read
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1607,11 +1607,10 @@ class HttpGitClient(GitClient):
         # Check if geturl() is available (urllib3 version >= 1.23)
         try:
             resp_url = resp.geturl()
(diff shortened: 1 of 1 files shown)


jelmer:
The behaviour of get_redirect_location() is not the same as that of geturl():

 |  get_redirect_location(self)
 |      Should we redirect and where to?
 |      
 |      :returns: Truthy redirect location string if we got a redirect status
 |          code and valid location. ``None`` if redirect status and no
 |          location. ``False`` if not a redirect status code.

havatv:
Hi @jelmer, and thanks for responding!

> The behaviour of get_redirect_location() is not the same as that of geturl():
> 
> | get_redirect_location(self)
> | Should we redirect and where to?
> |
> | :returns: Truthy redirect location string if we got a redirect status
> | code and valid location. `None` if redirect status and no
> | location. `False` if not a redirect status code.

I was aware of this. My mistakes was to just reuse some of the "old" code:

        resp.redirect_location = resp.get_redirect_location()
        return resp, read

I should have seen that there were other changes in ``_http_request``...
Big mistake - sorry! I will try to fix. Thanks again for your help.

havatv:
Hi again @jelmer. Could you have a look at the updates I have made.
I have now used the code from master in the ``try:`` and the code from 0.19.10 in the ``catch``.

master:
>         resp_url = resp.geturl()
>         resp.redirect_location = resp_url if resp_url != url else ''

0.19.10:
>        resp.redirect_location = resp.get_redirect_location()

jelmer:
Merged, thanks.

PR #753, opened 2020-04-02
add simple diffstat implementation

kevinhendricks:
I found a MIT license version of patch.py that I could extract the diffstat routine from and made a very simple diffstat.py replacement by a quick and dirty emulation the original Patch class parse behaviour.

If this is something you are interested in please feel free to use it, it is very simple and only meant to work with git style patches that dulwich generates.

If you like it but want changes, I would be happy to try to merge the emulation parser code into the diffstat routine itself but wanted to you to see how it relates to its origins so there are no licensing issues.  I assume MIT license can work for Apache as well.

Hope this helps.

### DIFF
--- a/dulwich/diffstat.py
+++ b/dulwich/diffstat.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+# vim:ts=4:sw=4:softtabstop=4:smarttab:expandtab
+
+# Copyright (c) 2020 Kevin B. Hendricks, Stratford Ontario Canada
+# All rights reserved.
+#
+# This diffstat code was extracted and heavily modified from:
+#
+#    https://github.com/techtonik/python-patch
+#        Under the following license:
+#
+#    Patch utility to apply unified diffs
+#    Brute-force line-by-line non-recursive parsing
+#
+#    Copyright (c) 2008-2016 anatoly techtonik
+#    Available under the terms of MIT license
+#
+# and falls under the exact same MIT license
+
+import sys
+import re
+
+# only needs to detect git style diffs as this is for
+# use with dulwich
+
+_git_header_name = re.compile(br'diff --git a/(.*) b/(.*)')
+
+_GIT_HEADER_START = b'diff --git a/'
+_GIT_BINARY_START = b'Binary file'
+_GIT_RENAMEFROM_START = b'rename from'
+_GIT_RENAMETO_START = b'rename to'
+_GIT_CHUNK_START = b'@@'
+_GIT_ADDED_START = b'+'
+_GIT_DELETED_START = b'-'
+_GIT_UNCHANGED_START = b' '
+
+# emulate original full Patch class by just extracting
+# filename and minimal chunk added/deleted information to
+# properly interface with diffstat routine
+
+
+def _parse_patch(lines):
+    """An internal routine to parse a git style diff or patch to generate
+       diff stats
+    Args:
+      lines: list of byte strings "lines" from the diff to be parsed
+    Returns: A tuple (names, nametypes, counts) of three lists:
+             names = list of repo relative file paths
+             nametypes - list of booolean values indicating if file
+                         is binary (True means binary file)
+             counts = list of tuples of (added, deleted) counts for that file
+    """
+    names = []
+    nametypes = []
+    counts = []
+    in_patch_chunk = in_git_header = binaryfile = False
+    currentfile = None
+    added = deleted = 0
+    for line in lines:
+        if line.startswith(_GIT_HEADER_START):
+            if currentfile is not None:
+                names.append(currentfile)
+                nametypes.append(binaryfile)
+                counts.append((added, deleted))
+            currentfile = _git_header_name.search(line).group(2)
+            binaryfile = False
+            added = deleted = 0
+            in_git_header = True
+            in_patch_chunk = False
+        elif line.startswith(_GIT_BINARY_START) and in_git_header:
+            binaryfile = True
+            in_git_header = False
+        elif line.startswith(_GIT_RENAMEFROM_START) and in_git_header:
+            currentfile = line[12:]
+        elif line.startswith(_GIT_RENAMETO_START) and in_git_header:
+            currentfile += b' => %s' % line[10:]
+        elif line.startswith(_GIT_CHUNK_START) and \
+                (in_patch_chunk or in_git_header):
+            in_patch_chunk = True
+            in_git_header = False
+        elif line.startswith(_GIT_ADDED_START) and in_patch_chunk:
+            added += 1
+        elif line.startswith(_GIT_DELETED_START) and in_patch_chunk:
+            deleted += 1
+        elif not line.startswith(_GIT_UNCHANGED_START) and in_patch_chunk:
+            in_patch_chunk = False
+    # handle end of input
+    if currentfile is not None:
+        names.append(currentfile)
+        nametypes.append(binaryfile)
+        counts.append((added, deleted))
+    return names, nametypes, counts
+
+
+# note must all done using bytes not string because on linux filenames
+# may not be encodable even to utf-8
+def diffstat(lines, max_width=80):
+    """Generate summary statistics from a git style diff ala
+       (git diff tag1 tag2 --stat)
+    Args:
+      lines: list of byte string "lines" from the diff to be parsed
+      max_width: maximum line length for generating the summary
+                 statistics (default 80)
+    Returns: A byte string that lists the changed files with change
+             counts and histogram
+    """
+    names, nametypes, counts = _parse_patch(lines)
+    insert = []
+    delete = []
+    namelen = 0
+    maxdiff = 0  # max changes for any file used for histogram width calc
+    for i, filename in enumerate(names):
+        i, d = counts[i]
+        insert.append(i)
+        delete.append(d)
+        namelen = max(namelen, len(filename))
+        maxdiff = max(maxdiff, i+d)
+    output = b''
+    statlen = len(str(maxdiff))  # stats column width
+    for i, n in enumerate(names):
+        binaryfile = nametypes[i]
+        # %-19s | %-4d %s
+        # note b'%d' % namelen is not supported until Python 3.5
+        # To convert an int to a format width specifier for byte
+        # strings use str(namelen).encode('ascii')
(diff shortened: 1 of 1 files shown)


jelmer:
Thanks, this is great. Can you add unit tests and docstrings?

kevinhendricks:
I will take a shot at adding docstrings similar to those in porcelain.py.  But I have never been a big fan of separate unit tests for standalone python code.  The tests never stay with the code.  And the test code ends up being longer and more complex than the code it is supposed to be testing and so just grows the overall support burden.   

I could possibly extend the main routine to run a simple set of tests to show basic functionality.  That way anyone who wants to can reuse that file without worrying about needing to extract tests as the basic tests are built into __main__.

I will look into both as soon as I get some extra time from working on Sigil.  Hopefully in a week or two at most.

kevinhendricks:
one question, are you okay with using dulwich's own github repo to dynamically generate test case diffs from dulwich's own history on the fly and compare to actual git output.   I was thinking of testing by running actual git in a full log of the github dulwich repo commit changes and comparing diffstat outputs with what git produces.

jelmer:
Hi Kevin,

Thanks again for working on this.

On Thu, 2020-04-02 at 14:59 -0700, Kevin Hendricks wrote:
> I will take a shot at adding docstrings similar to those in
> porcelain.py. But I have never been a big fan of separate unit tests
> for standalone python code. The tests never stay with the code. And
> the test code ends up being longer and more complex than the code it
> is supposed to be testing and so just grows the overall support
> burden.
I agree that that is true for single independent files that can be
copied between projects...

> I could possibly extend the main routine to run a simple set of tests
> to show basic functionality. That way anyone who wants to can reuse
> that file without worrying about needing to extract tests as the
> basic tests are built into main.
... but please add a module under dulwich.tests.test_diffstat. If you
add it in a separate main function then the tests won't usually be run
and it will almost certainly bitrot. 

Alternatively, it may make sense to just upload this module as a
separate module on pip (rather than as part of dulwich) since it
doesn't appear to be specific to Dulwich in any way.

Cheers,

Jelmer

jelmer:
On Thu, 2020-04-02 at 16:02 -0700, Kevin Hendricks wrote:
> one question, are you okay with using dulwich's own github repo to
> dynamically generate test case diffs from dulwich's own history on
> the fly and compare to actual git output. I was thinking of testing
> by running actual git in a full log of the github dulwich repo commit
> changes and comparing diffstat outputs with what git produces.

Unfortunately the problem with using Dulwich' own repository is that it
makes tests harder to debug, but it also breaks when you run the tests
in a copy of Dulwich not from Git (such as when downloaded as a
tarball).

This approach also adds a dependency on C Git for running the tests,
which Dulwich currently doesn't have.

There are some functions for creating simple repositories in the
testsuite that the rest of the Dulwich tests use.

Jelmer

kevinhendricks:
To test things tonight I used "git diff 1.0.0 1.2.1" on Sigil-Ebooks/Sigil (but any 2 historic tags would do)  and captured that diff to a file.  It is 9,911,116 bytes.  I ran "git diff 1.0.0 1.2.1 --stat on the same repo and saved it 33,824 bytes.

Then I used the diffstat.py code on these standalone saved diff file and compared the output to what git generated.  By using this approach I was able to improve diffstat.py to better handle binary files, and better handle rename from and rename to output.  This particular test saw 480 files changed, 55870 insertions and 46641 deletions which matches what git produced.

So do you have a place to put larger test diff data file (zipped say 1.5 meg worth) of stand alone pre generated git diff and a 33k git stat file.  If so, we can use a reasonably exhaustive real piece of empirical data to test with without requiring C git and it would run from downloaded archive, if that amount of test data would be allowed.  Would that work?

kevinhendricks:
Also, when you get a free moment I posted a potential patch for porcelain.add in my original bug report. Please let me know if it looks okay to you.

jelmer:
On Thu, 2020-04-02 at 18:18 -0700, Kevin Hendricks wrote:
> To test things tonight I used "git diff 1.0.0 1.2.1" on Sigil-
> Ebooks/Sigil (but any 2 historic tags would do) and captured that
> diff to a file. It is 9,911,116 bytes. I ran "git diff 1.0.0 1.2.1 --
> stat on the same repo and saved it 33,824 bytes.
> 
> Then I used the diffstat.py code on these standalone saved diff file
> and compared the output to what git generated. By using this approach
> I was able to improve diffstat.py to better handle binary files, and
> better handle rename from and rename to output. This particular test
> saw 480 files changed, 55870 insertions and 46641 deletions which
> matches what git produced.
> 
> So do you have a place to put larger test diff data file (zipped say
> 1.5 meg worth) of stand alone pre generated git diff and a 33k git
> stat file. If so, we can use a reasonably exhaustive real piece of
> empirical data to test with without requiring C git and it would run
> from downloaded archive, if that amount of test data would be
> allowed. Would that work?
I'd rather have a much smaller (purposefully crafted) test that just
covers the important use cases. Debugging any kind of difference in a
1.5 meg file is going to be painful, and if all of the tests were that
large the repo would be huge. :-)

Dulwich bundled a couple of git repositories for testing in its early
days, and they have been a massive source of pain.

There is also the problem of shipping a part of the Sigil repository if
it's GPLv2ed.

Jelmer

kevinhendricks:
It would not be shipping code, just a git generated diff.  We could of course do the same with a diff generated from the dulwich repo.  I chose to use Sigil because Sigil is quite active and would provide a larger number of interesting/different diff chunks to test with.  I will try to find a similar diff from the dulwich repo and strip it down to 2 or 3 interesting cases (a rename or two, a couple of binary files, and a couple of small file diffs, one add and one delete maybe).  A will inline them into diffstat main and you can add a test_diffstat routine to invoke main to run the self-contained tests.  Note, this is not a general diffstat program, it really only handles git diff files so it really is only useful to dulwich.  Sigil has its own version to use.  If you do not want it, that is of corse your call.

kevinhendricks:
Okay, cleaned up the code, tried to add docstrings and a self test.  If that works for you, great.  If not please just close this.  The code is yours either way in case you want it for the future.

jelmer:
Unfortunately including a diff extracted from the Sigil repository still means that we're including GPLv3-only code.

The unit test will need more work before merging, it needs to be moved into dulwich/tests (rather than just being invoked from there). I'll mark this merge proposal as WIP for the moment, and may come back to it.

kevinhendricks:
If you read the comments in the we the authors of Sigil give permission to use the diffs of  the patches we wrote, and the docs we wrote.  Your choice though.

jelmer:
Ah, I missed that - thanks. That just leaves the moving of the unit tests.

jelmer:
It also looks like this introduces a slew of flake8 errors, but that's not shown because the pull request removes the travis.yml file.

kevinhendricks:
I had problems with your travis triggering in our repo of dulwich.  I probably killed it by accident in ourrepo of dulwich.  Sorry about that.  I will make a clean checkout  and repost the PR.  Please go ahead and close this in the interim.

Not a flake8 user but I will take a look.

kevinhendricks:
Fixed the flake8 warnings (sad to think we have advanced so little over fortran 72 character limits). Disabled them on the literal test strings.
Created a clean pull from dulwich to prevent travis issues and added in flake8 "warning" fixed version of diffstat.py.

I have no plans to create "unit tests" for this thing.  If that means you do not want the code then please state that and I will close this.  We can just use it in our own fork of dulwich.

Ditto for the diff3merge.py code that I offered in issue 754:

  https://github.com/dulwich/dulwich/issues/754

that does a 3 way merge that I wrote specifically to fit an MIT license to try to help you out. 
The lack of a merge capability has been around for far too long and really impacts the usefulness of dulwich.  But again, we will just use it (or a gnu2 version) internally.

I must say dulwich (you) and Sigil (I) have very different approaches to getting community involvement in our projects.  I literally accept and welcome almost any help and do not put hurdles in front of accepting working code (other than license concerns) in order to foster and grow the community.  Something you might want to think about.

Take care (and stay healthy!).

jelmer:
On Sun, Apr 05, 2020 at 05:15:13PM -0700, Kevin Hendricks wrote:
> Fixed the flake8 warnings (sad to think we have advanced so little over fortran 72 character limits). Disabled them on the literal test strings.
> Created a clean pull from dulwich to prevent travis issues and added in flake8 "warning" fixed version of diffstat.py.
> 
> I have no plans to create "unit tests" for this thing.  If that means you do not want the code then please state that and I will close this.  We can just use it in our own fork of dulwich.
> 
> I must say dulwich (you) and Sigil (I) have very different approaches to getting community involvement in our projects.  I literally accept and welcome almost any help and do not put hurdles in front of accepting working code (other than license concerns) in order to foster and grow the community.  Something you might want to think about.

Thanks again for the patch. I do feel that leaving the patch on the
table is not a great outcome. However, unfortunately I really do want to
make sure we have unit tests in the standard test suite before merging
new functionality like this.

An alternative is that we merge it into dulwich/contrib; this
directory is basically a grab-bag of code related to Dulwich that may
be useful to Dulwich users. It contains code that isn't regularly
tested, and thus more likely to be broken.

I may come back to it to merge the tests into the regular testsuite at
some point.

Or, as I suggested earlier, since this code is independent of the
rest of Dulwich perhaps this would be useful enough to submit as a
separate Python package on pypi.

On community engagement:

I don't think I have an unreasonably high bar for contributions to
Dulwich. Requirements for contributions are documented in
CONTRIBUTING.rst: contributions don't have to be perfect, they just
need to be an improvement over the current state of the code and not a
hindrance in the future.  In particular, I am wary of anything that
makes the maintenance burden heavier; non-standard tests or untested
code are a big pain in a large Python code base like Dulwich.

The test coverage currently isn't perfect and the lack of test
coverage in some places already bites us regularly. It also makes any
refactoring painful.

There have been 127 different contributors to Dulwich over the course
of the project, many with more than one contribution. Maybe there
would have been more if it was easier to contribute, but it would
also be harder to make changes and the project would be buggier.

(I'll follow up separately about the merge3 implementation)

-- 

kevinhendricks:
Understood.  As I said, this diffstat will only work with git style diffs so it is only valuable to a git clone project written in python.   And  the code to count adds and deletes itself is quite trivial.  I will close this PR and if you ever decide you want to use it you can just let me know.

jelmer:
Let me see if we can at least see if we can get this into dulwich.contrib.

jelmer:
Thanks, merged into dulwich.contrib.

Issue #705, opened 2019-04-16
Cannot push from shallow fetch


ckclark:
## Environment

OS: linux
version: dulwich==0.19.11

## Steps to reproduce
```python
from dulwich.porcelain import push, clone, add, commit
repo = clone('<git-url>', 'out', depth=1)
fp = open('out/.gitignore', 'w')
fp.write('*.pyc')
fp.close()
add(repo, 'out/.gitignore')
commit(repo, b'Add .gitignore')
push(repo, '<git-url>', 'master')
```
This yields
```
Traceback (most recent call last):
  File "filename.py", line 197, in <module>
    push(repo, '<git-url>', 'master')
  File "/library/path/to/dulwich/porcelain.py", line 822, in push
    progress=errstream.write)
  File "/library/path/to/dulwich/client.py", line 784, in send_pack
    ofs_delta=(CAPABILITY_OFS_DELTA in negotiated_capabilities))
  File "/library/path/to/dulwich/object_store.py", line 248, in generate_pack_data
    self.generate_pack_contents(have, want, progress))
  File "/library/path/to/dulwich/object_store.py", line 236, in generate_pack_contents
    return self.iter_shas(self.find_missing_objects(have, want, progress))
  File "/library/path/to/dulwich/object_store.py", line 211, in find_missing_objects
    get_parents=get_parents)
  File "/library/path/to/dulwich/object_store.py", line 1110, in __init__
    have_commits, get_parents=self._get_parents)[0]
  File "/library/path/to/dulwich/object_store.py", line 288, in _collect_ancestors
    cmt = self[e]
  File "/library/path/to/dulwich/object_store.py", line 120, in __getitem__
    type_num, uncomp = self.get_raw(sha)
  File "/library/path/to/dulwich/object_store.py", line 482, in get_raw
    raise KeyError(hexsha)
KeyError: '<parent-commit-id>'
```
The object identified by parent-commit-id is not pulled due to the `depth=1` config. But this operation works if git cmdline is used.

```shell
$ git clone --depth=1 <git-url> out
Cloning into 'out'...
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 5 (delta 0), reused 4 (delta 0), pack-reused 0
Receiving objects: 100% (5/5), done.
$ cd out
$ echo "*.pyc" > .gitignore
$ git add .gitignore
$ git commit -m 'Add .gitignore'
[...]

PR #757, opened 2020-04-23
Quick hack to "fix" pushing from a shallow clone

brechtm:
Just something to get your feedback on, @jelmer.

This includes a unit test that triggers #705.

As far as I can tell, the proper way to fix this is to not have `_collect_ancestors` "descend" into the parent commits if the current commit is a shallow commit. However, the list of shallow commits is not available in _ObjectStore_. Can we pass a _Repo_ reference to the _ObjectStore_? Or perhaps you can suggest a better approach?

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -361,7 +361,7 @@ class GitClient(object):
         """
         raise NotImplementedError(cls.from_parsedurl)
 
-    def send_pack(self, path, update_refs, generate_pack_data,
+    def send_pack(self, path, update_refs, shallow, generate_pack_data,
                   progress=None):
         """Upload a pack to a remote repository.
 
@@ -370,6 +370,7 @@ class GitClient(object):
           update_refs: Function to determine changes to remote refs. Receive
             dict with existing remote refs, returns dict with
             changed refs (name -> sha, where sha=ZERO_SHA for deletions)
+          shallow: Set of shallow commits generate_pack_data should skip
           generate_pack_data: Function that can return a tuple
             with number of objects and list of pack data to include
           progress: Optional progress function
@@ -772,7 +773,7 @@ class TraditionalGitClient(GitClient):
         """
         raise NotImplementedError()
 
-    def send_pack(self, path, update_refs, generate_pack_data,
+    def send_pack(self, path, update_refs, shallow, generate_pack_data,
                   progress=None):
         """Upload a pack to a remote repository.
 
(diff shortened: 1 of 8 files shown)


jelmer:
Hi Brecht, 

Sorry for the slow reply.

> As far as I can tell, the proper way to fix this is to not have _collect_ancestors "descend" into the parent commits if the current commit is a shallow commit. However, the list of shallow commits is not available in ObjectStore. Can we pass a Repo reference to the ObjectStore? Or perhaps you can suggest a better approach?

Passing a repo reference to Repo into ObjectStore would create a circular reference and break the abstractions, so I'd rather avoid doing that.

Perhaps we can simply pass in the list of shallow ids along with have and want?

jelmer:
Thanks for the unit test - I've at least merged that, but marked it as known failing for the moment.

brechtm:
> Sorry for the slow reply.

No problem. What's three days? And it's not like I'm paying for professional support! 😄 

> Passing a repo reference to Repo into ObjectStore would create a circular reference and break the abstractions, so I'd rather avoid doing that.
> 
> Perhaps we can simply pass in the list of shallow ids along with have and want?

I have implemented this in the updated branch. This did require passing the list of shallow commits along quite a number of function call levels. It would be cleaner if the list of shallow commits was also stored in ObjectStore, but I guess it doesn't belong there.

I'm not fully aware of what exactly is happening in MissingObjectFinder, so I invite you to review that. It seems to be doing the right thing, though.

jelmer:
On Mon, Apr 27, 2020 at 03:15:39AM -0700, Brecht Machiels wrote:
> > Sorry for the slow reply.
> No problem. What's three days? And it's not like I'm paying for professional support! 😄 
Hah, for some reason I thought it was a lot more than that. :-)

> > Passing a repo reference to Repo into ObjectStore would create a circular reference and break the abstractions, so I'd rather avoid doing that.
> > 
> > Perhaps we can simply pass in the list of shallow ids along with have and want?
> 
> I have implemented this in the updated branch. This did require passing the list of shallow commits along quite a number of function call levels. It would be cleaner if the list of shallow commits was also stored in ObjectStore, but I guess it doesn't belong there.
> 
> I'm not fully aware of what exactly is happening in MissingObjectFinder, so I invite you to review that. It seems to be doing the right thing, though.
Thanks! This broadly LGTM.

For backwards compatibility reasons, we probably want to default to an
empty set of shallow commits.

I would think that in the case of send-pack the remote repository
would tell you what shallows it has so that you can avoid sending
those. However, such functionality doesn't exist and https://git-scm.com/docs/shallow says:

* It is unclear how "push into/from a shallow repo" should behave.

I'll see if I can merge this with some additional refactoring to hide
the shallow argument.

Jelmer

-- 

jelmer:
Merged with some changes - mostly to avoid making API changes. Rather than move shallow data into the object store, I've added a convenience method on Repo that calls .get_shallow().

matdrapeau:
It looks like pulling from shallow clone is raising the same `keyerror` exception

PR #758, opened 2020-05-03
Use utf-8 instead of file system encoding to convert path to str.

manueljacob:
This seems the correct thing to do because it enables the user to control the exact bytes that are used to encode the URL.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1557,10 +1557,9 @@ class HttpGitClient(GitClient):
 
     def _get_url(self, path):
         if not isinstance(path, str):
-            # TODO(jelmer): this is unrelated to the local filesystem;
-            # This is not necessarily the right encoding to decode the path
-            # with.
-            path = path.decode(sys.getfilesystemencoding())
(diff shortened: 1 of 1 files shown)


dulwich/client.py
```diff
@@ -1557,10 +1557,9 @@ def __repr__(self):
 
     def _get_url(self, path):
         if not isinstance(path, str):
-            # TODO(jelmer): this is unrelated to the local filesystem;
-            # This is not necessarily the right encoding to decode the path
-            # with.
-            path = path.decode(sys.getfilesystemencoding())
+            # urllib3.util.url._encode_invalid_chars() converts the path back
+            # to bytes using the same parameters.
+            path = path.decode('utf-8', 'surrogatepass')
```
jelmer, line 1562: Can we drop the second argument? Replacing characters seems like a recipe for confusing problems - it's better to just raise an error here and let the user pass in the right arguments.

manueljacob: I dropped it.

PR #762, opened 2020-05-11
 Make it possible to pass custom classes to default_urllib3_manager()

manueljacob:


jelmer:
LGTM, I'll merge this later this week.

dulwich/client.py
```diff
@@ -57,6 +57,8 @@
 except ImportError:
     import urllib.parse as urlparse
 
+import urllib3
```
jelmer, line 60: Please don't import urllib3 as a top-level import - both for performance reason, but also to allow people to use dulwich without having urllib3 installed if they don't need http support.

manueljacob: Makes sense. Fixed.

PR #764, opened 2020-05-21
Coding problem

timmy61109:
To solve the `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe9 
in position 27: ordinal not in range (128)` problem, change `ascii` to` 
utf8`.

Before modification.

2020-05-21 10-57-27 的螢幕擷圖

After modification.

2020-05-21 10-57-55 的螢幕擷圖

If it is determined that this is the solution, please let it merge, because the company cannot pass the CI check because of this error.

Issue 763

### DIFF
--- a/dulwich/patch.py
+++ b/dulwich/patch.py
@@ -117,15 +117,15 @@ def unified_diff(a, b, fromfile='', tofile='', fromfiledate='',
             fromdate = '\t{}'.format(fromfiledate) if fromfiledate else ''
             todate = '\t{}'.format(tofiledate) if tofiledate else ''
             yield '--- {}{}{}'.format(
-                fromfile.decode("ascii"),
+                fromfile.decode("utf8"),
                 fromdate,
                 lineterm
-                ).encode('ascii')
+                ).encode('utf8')
             yield '+++ {}{}{}'.format(
-                tofile.decode("ascii"),
+                tofile.decode("utf8"),
                 todate,
                 lineterm
-                ).encode('ascii')
+                ).encode('utf8')
 
         first, last = group[0], group[-1]
         file1_range = _format_range_unified(first[1], last[2])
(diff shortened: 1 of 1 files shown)


jelmer:
These paths aren't necessarily utf-8. Can you make the file encoding an argument instead?

timmy61109:
@jelmer Yes, I can, but what do I need to do after providing the parameters? For example, how can I use it for other functions?

jelmer:
I'm not sure I follow?

timmy61109:
@jelmer So you want me to change the name of the file to what the program can support?

timmy61109:
> I'm not sure I follow?

I don't quite understand the meaning of this sentence.

jelmer:
Add an argument with the file encoding to the diff function, and allow the caller to pass that encoding in. You can pass in utf-8, but others may want to specify a different encoding.

timmy61109:
> 將帶有文件編碼的參數添加到diff函數，並允許調用方傳遞該編碼。您可以傳遞utf-8，但其他人可能希望指定其他編碼。

Understand, can I use `self` as a variable?

If this submission is approved, do I need to write about the changes?

timmy61109:
> Add an argument with the file encoding to the diff function, and allow the caller to pass that encoding in. You can pass in utf-8, but others may want to specify a different encoding.

Understand, can I use `self` as a variable?

If this submission is approved, do I need to write about the changes?

jelmer:
I don't understand - this is not a method on an object so self is irrelevant.

timmy61109:
> I don't understand - this is not a method on an object so self is irrelevant.

OK. I will use the parameter method, just how to provide the use of this parameter, this will be the problem.

jelmer:
Please unmark this as draft when it's ready for another review.

jelmer:
Superseded by a2d13e546a7f977a90ab1479b445b3ebe95bb36e

PR #770, opened 2020-05-27
Only return files from the loose store that look like git objects

olasd:
In some situations (e.g. a git crash), one can end up with temporary files in
the git loose object store. These temporary files pollute dulwich's loose object
iterator.

Ref. implementation in git: https://github.com/git/git/blob/2d5e9f31ac46017895ce6a183467037d29ceb9d3/sha1-file.c#L2292

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -47,6 +47,7 @@ from dulwich.objects import (
     hex_to_filename,
     S_ISGITLINK,
     object_class,
+    valid_hexsha,
     )
 from dulwich.pack import (
     Pack,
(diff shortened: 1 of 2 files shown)


dulwich/object_store.py
```diff
@@ -644,7 +644,12 @@ def _iter_loose_objects(self):
             if len(base) != 2:
                 continue
             for rest in os.listdir(os.path.join(self.path, base)):
-                yield os.fsencode(base+rest)
+                if len(rest) != 38:
+                    continue
+                sha = os.fsencode(base+rest)
```
jelmer, line 649: Would it make sense to use valid_hexsha() here instead?

jelmer:
Merged, thanks.

PR #774, opened 2020-06-10
fix command-line parsing for cli.py

onyxware:
Only the arguments of the subcommand (i.e dulwich clone ...) should be passed to the function, not the subcommand itself as well:
(continuing the example) 'github.com/dulwich/dulwich.git', not 'clone github.com/dulwich/dulwich.git'


onyxware:
Nevermind, the indice was correct.

PR #775, opened 2020-06-11
Supplementary ignore list files

timmy61109:
I added and added most of the files that Python will need to ignore. Especially not ignoring the directories and files generated by virtualenv is actually inconvenient for me. If there are repetitions, please let me improve and simplify.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,108 @@ docs/api/*.txt
 .mypy_cache/
 .eggs
 dulwich.dist-info
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.hypothesis/
+.pytest_cache/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+local_settings.py
+db.sqlite3
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# pyenv
+.python-version
+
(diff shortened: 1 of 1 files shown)


jelmer:
There are lots of duplicates here that were already in the ignore file, and a lot of files that are not relevant to Dulwich at all.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
...
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.hypothesis/
+.pytest_cache/
+
+# Translations
```
jelmer, line 78: These are not relevant.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
...
+*.mo
+*.pot
+
+# Django stuff:
```
jelmer, line 82: These are not relevant.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
...
+*.log
+local_settings.py
+db.sqlite3
+
+# Flask stuff:
```
jelmer, line 87: These are not relevant.

.gitignore
timmy61109: In fact, these are the ignore lists that are often encountered in Python development. For example, `.mo` and `.pot` can provide localized languages.

.gitignore
timmy61109: It is very likely that it will be used in the future, such as the use of APIs provided by the Git server.

.gitignore
timmy61109: It is very likely that it will be used in the future, such as the use of APIs provided by the Git server.

.gitignore
jelmer: yes, but for dulwich it's not relevant. And when it becomes relevant we can add these ignores.

.gitignore
jelmer: No, it's not likely that would happen with django. And even if, we can add that when it becomes relevant.

.gitignore
jelmer: No, we won't use flask in Dulwich.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
...
 .eggs
 dulwich.dist-info
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
```
jelmer, line 35: Already ignored.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
 .mypy_cache/
 .eggs
 dulwich.dist-info
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
```
jelmer, line 30: Already ignored.

.gitignore
```diff
@@ -25,3 +25,108 @@ docs/api/*.txt
...
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
```
jelmer, line 39: Already ignored.

jelmer:
Sorry, but this pull request is noisy and doesn't provide any value that I can see. If there are specific files that you have actually seen appear that dulwich' current ignore file doesn't cover, then I'm happy to take updates. But the bulk of the lines that this adds are just irrelevant.

timmy61109:
I will reorganize and re-request

PR #779, opened 2020-06-17
Use chain.from_iterable in compat/test_repository.py

cool-RR:
This is a continuation of #778 . I don't know whether this file is important enough for this change, given it's a test file in the compat package. Your call.


jelmer:
Yeah, this code would only be used with specific small repositories anyway. But sure, why not.

PR #809, opened 2020-10-30
gha: move linux wheels to wheelhouse dir

efiop:
Just following the convention for wheel location. Clearly the workflow
file needs some refactoring, but it is out of scope for this particular PR.


efiop:
Ah, I see this is no longer needed https://github.com/dulwich/dulwich/commit/4d8df87f8fe5bda176eb52f7e8b97d5f8ea4d5d7 . Thanks for a quick fix!

PR #812, opened 2020-11-12
[setup] Add py.typed to allow type checking

david-caro:
This enables PEP 561 compatibility and allows other packages that use
this one as third party to type check.

See https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages


jelmer:
Merged, thanks!

PR #814, opened 2020-11-23
Fix minor stash bugs

pmrowla:
Fixes minor bugs in `stash.Stash`

* use `os.fsencode`/`os.fsdecode` when needed so that bytes/strings are not mixed
* make `__getitem__` and `__len__` work properly (use `list(stashes())` instead of `_stashes()`)

### DIFF
--- a/dulwich/stash.py
+++ b/dulwich/stash.py
@@ -47,7 +47,7 @@ class Stash(object):
 
     def stashes(self):
         reflog_path = os.path.join(
-            self._repo.commondir(), 'logs', self._ref)
+            self._repo.commondir(), 'logs', os.fsdecode(self._ref))
         try:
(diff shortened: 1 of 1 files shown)


pmrowla:
@jelmer we have some downstream issues in https://github.com/iterative/dvc that are blocked by this, and would appreciate it if we could get a new pypi release w/this fix whenever you get the chance

jelmer:
On Wed, Nov 25, 2020 at 06:21:41PM -0800, Peter Rowlands (변기호) wrote:
> @jelmer we have some downstream issues in https://github.com/iterative/dvc that are blocked by this, and would appreciate it if we could get a new pypi release w/this fix whenever you get the chance
I'll do one before the weekend, with a bit of luck later today.

pmrowla:
Much appreciated, thanks!

PR #815, opened 2020-11-24
Fix for byte problems (TypeError) reading alternate paths 

gracinet:
On Python 3, dulwich crashes with `TypeError` if there is a relative path in `info/alternates`.

This PR reproduces and fixes the problem in tests, and also an issue with comments uncovered while investigating my actual issue.

For more context, here's the [issue reported by one of my users](https://foss.heptapod.net/heptapod/heptapod/-/issues/370). Happens in GitLab because  the path to pool is relative.

Since this is I believe my first contribution to Dulwich, do you want an issue to supplement that?


jelmer:
An issue isn't necessary - thanks for the patch & the unit test!

PR #817, opened 2020-11-25
upload sdist

dtrodrigues:
Uploads the sdist to PyPI, not just wheels. 

Fixes https://github.com/dulwich/dulwich/issues/816


jelmer:
Thanks, but I manually upload to pypi - so that I can upload the PGP signature as well. For some reason this broke for 0.20.13.

PR #831, opened 2021-01-29
I find a logic error start with line 476

xiayulu:


jelmer:
I don't understand what is wrong with the logic there. If this is a bug, please provide a test case as well.

xiayulu:
Oh, I misunderstood with `sha` and `hexsha`.
Any way, thanks for your time, dulwich is a cool project.

PR #832, opened 2021-01-29
client/credentials: ignore end-of-line character

gracinet:
Before this change, credential files consisting of only one
line without end-of-line would work.

The change is actually stripping any leading and trailing whitespace,
but that shouldn't be a problem. The shared test setup is updated with
a newline because that should be a fairly common case.


gracinet:
Great, thanks for the merge

PR #833, opened 2021-02-02
Setup.cfg Update

timmy61109:
If this update is adopted and Flake8, pylint, and pydocstyle are added, the following problems will be avoided.

2021-01-04 18-10-49 的螢幕擷圖

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,7 +2,6 @@
 exclude =
     .git,
     __pycache__,
-    docs/source/conf.py,
     docs/conf.py,
     old,
     build,
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,21 @@
+[flake8]
+exclude =
+    .git,
+    __pycache__,
+    docs/source/conf.py,
+    docs/conf.py,
+    old,
+    build,
+    dist,
+    build-pypy,
+    .tox,
+    .eggs,
+    venv
+
 [mypy]
 ignore_missing_imports = True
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,5 +1,21 @@
 [flake8]
-exclude = build,.git,build-pypy,.tox
+exclude =
+    .git,
```
jelmer, line 3: flake8 doesn't complain about any of these things, why do they need to be added to the exclude list?

setup.cfg
```diff
@@ -1,5 +1,21 @@
...
+    .eggs,
+    venv
 
 [mypy]
 ignore_missing_imports = True
+
+[pydocstyle]
+
+[pylint]
+disable = R0205
```
jelmer, line 20: Please also add pylint to make / GitHub actions, otherwise this will just regress.

jelmer:
(I've converted this back to draft for the moment, please click the "Ready for review" button once it's ready)

setup.cfg
timmy61109: Eliminating unnecessary inspections can reduce the time and computing resources spent in inspections.

timmy61109: These are common places that do not need to be checked, and this project is also applicable

jelmer:
CLosing this for the moment since it's been stalled for > 6 months; feel free to reopen if you'd like to continue working on it.

Issue #836, opened 2021-02-09
gitignore manpage example does not work with IgnoreFilterManager


courentin:
Hello all,

I figured out a bug while using https://github.com/iterative/dvc, after some investigations, I think the root issue comes from dulwich.

### The issue

Let's say I have this file structure:

```
data/
	.gitignore <- gitignore A
	foo/
		.gitignore <- gitignore B
		bar
```

In `gitignore A`:
```
/*
!/foo
```

In `gitignore B`:
```
/bar
```

Running `is_ignored("data/foo/bar")` on this kind of repository gives `False` while gitignore B actually ignore this file.

### To reproduce

So I tried to reproduce this test: https://github.com/dulwich/dulwich/blob/bc0ca9e58814cb95961b42992a4011ced8673add/dulwich/tests/test_ignore.py#L176-L188

with the `IgnoreFilterManager` and it seems to fails.
I'll join a PR with a failing test next.

Is it the intented behaviour or am I pointing a real issue here?

PS: I really don't know the dulwhich codebase as I discovered it 30min ago.

PR #837, opened 2021-02-09
Failing test for manpage gitignore example with IgnoreFilterManager

courentin:
It is a demonstration of #836
However, I don't think I'll be able to fix it on my own.


jelmer:
Thanks for adding a test to reproduce the issue! Can you split it into a separate test function and then mark it as known failing by adding the @unittest.expectedFailure decorator?

courentin:
Done!

PR #838, opened 2021-02-10
Fix find_matching to stop returning after the first match

### LINKED ISSUE #836: gitignore manpage example does not work with IgnoreFilterManager

courentin:
This PR should fix #836
Tests are passing but what was the reason for `find_matching` to stop after the first match?

### DIFF
--- a/dulwich/ignore.py
+++ b/dulwich/ignore.py
@@ -328,8 +328,6 @@ class IgnoreFilterManager(object):
     def find_matching(self, path: str) -> Iterable[Pattern]:
         """Find matching patterns for path.
 
-        Stops after the first ignore file with matches.
-
         Args:
           path: Path to check
         Returns:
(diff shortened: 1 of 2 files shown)


jelmer:
That was for performance reasons - but clearly that wasn't the right call. Thanks for the patch!

courentin:
Right!
Can we merge this PR and release a new version? So that dulwich could be upgraded in the dvc package and close https://github.com/iterative/dvc/issues/5438 🙏

PR #841, opened 2021-02-17
stash: implement drop

pmrowla:
* Implements `Stash.drop`
* Adds `drop_reflog_entry` method in `dulwich/reflog` for removing and rewriting reflog entries

### DIFF
--- a/dulwich/reflog.py
+++ b/dulwich/reflog.py
@@ -93,3 +93,62 @@ def read_reflog(f):
     """
     for line in f:
         yield parse_reflog_line(line)
+
+
+def drop_reflog_entry(f, index, rewrite=False):
+    """Drop the specified reflog entry.
+
+    Args:
+        f: File-like object
+        index: Reflog entry index (in Git reflog reverse 0-indexed order)
+        rewrite: If a reflog entry's predecessor is removed, set its
+            old SHA to the new SHA of the entry that now precedes it
+    """
+    if index < 0:
+        raise ValueError("Invalid reflog index %d" % index)
+
+    log = []
+    offset = f.tell()
+    for line in f:
+        log.append((offset, parse_reflog_line(line)))
+        offset = f.tell()
+
+    inverse_index = len(log) - index - 1
+    write_offset = log[inverse_index][0]
+    f.seek(write_offset)
+
+    if index == 0:
+        f.truncate()
+        return
+
+    del log[inverse_index]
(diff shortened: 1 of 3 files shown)


dulwich/reflog.py
```diff
@@ -93,3 +93,62 @@ def read_reflog(f):
     """
     for line in f:
         yield parse_reflog_line(line)
+
+
+def drop_reflog_entry(f, index, rewrite=False):
```
pmrowla, line 98: This could have a porcelain equivalent to `git reflog delete ...` but I'm not sure it's needed

dulwich/stash.py
```diff
@@ -62,7 +62,14 @@ def from_repo(cls, repo):
...
+        with open(self._reflog_path, "rb+") as f:
+            drop_reflog_entry(f, index, rewrite=True)
+        if len(self) == 0:
+            os.remove(self._reflog_path)
+            del self._repo.refs[self._ref]
+            return
+        if index == 0:
+            self._repo.refs[self._ref] = self[0].new_sha
 
     def pop(self, index):
```
pmrowla, line 74: I left `pop` unimplemented since it's not entirely clear what pop should do when we aren't touching the working tree.

dulwich/stash.py
```diff
@@ -62,7 +62,14 @@ def from_repo(cls, repo):
 
     def drop(self, index):
         """Drop entry with specified index."""
-        raise NotImplementedError(self.drop)
+        with open(self._reflog_path, "rb+") as f:
```
pmrowla, line 65: Wasn't sure if calling `open` directly here is a problem (vs using `GitFile`)? `drop_reflog_entry` needs a read+write file handle, and I noticed that when writing reflog entries for commits in `Repo` open is also used directly (but in append mode)

dulwich/reflog.py
jelmer: That would be nice, but I don't think that needs to be in this pull request.

dulwich/stash.py
jelmer: today, stash doesn't update the working tree (as per its docstring) but it should.

dulwich/stash.py
jelmer: it's good to be consistent. GitFile provides locking, which seems like it would be beneficial. We should double check what Git does here.

jelmer (approved):

PR #851, opened 2021-03-15
porcelain.clone: add branch option equivalent to 'git clone --branch'

pmrowla:
- unifies behavior between `porcelain.clone` and `Repo.clone` in `dulwich/clone.py`
- cloned repo now uses the remote's default branch name if it differs from dulwich `DEFAULT_REF` (previously `porcelain.clone` would always use `refs/heads/master` regardless of the remote branch name)
- adds `branch=...` parameter for porcelain and repo clone equivalent to `git clone --branch` for checking out specific branch or tag name after clone (instead of checking out default branch)

### DIFF
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -652,7 +652,10 @@ class CloneTests(PorcelainTestCase):
         self.assertEqual(0, len(target_repo.open_index()))
         self.assertEqual(c1.id, target_repo.refs[b"refs/heads/else"])
         self.assertEqual(c1.id, target_repo.refs[b"HEAD"])
-        self.assertEqual({b"HEAD": b"refs/heads/else"}, target_repo.refs.get_symrefs())
+        self.assertEqual(
+            {b"HEAD": b"refs/heads/else", b"refs/remotes/origin/HEAD": b"refs/remotes/origin/else"},
+            target_repo.refs.get_symrefs(),
+        )
 
 
 class InitTests(TestCase):
@@ -2385,6 +2388,8 @@ class FetchTests(PorcelainTestCase):
             for k, v in remote_refs.items()
             if k.startswith(local_ref_prefix)
         }
+        if b"HEAD" in locally_known_remote_refs and b"HEAD" in remote_refs:
+            normalized_remote_refs[b"HEAD"] = remote_refs[b"HEAD"]
 
         self.assertEqual(locally_known_remote_refs, normalized_remote_refs)
 
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -630,9 +630,12 @@ class CloneTests(PorcelainTestCase):
         r.close()
 
     def test_source_broken(self):
-        target_path = tempfile.mkdtemp()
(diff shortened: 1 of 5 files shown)


pmrowla:
Still needs tests, will also cleanup/unify behavior for `porcelain.clone` and `Repo.clone`

pmrowla:
@jelmer just a heads up that this is ready for review now

pmrowla:
@jelmer I know you mentioned at one point that some of what's in this PR should maybe get moved out of `repo.py`, but it still needs an initial review

alchem1ster:
I could be wrong, but if you use here porcelain.clone() twice in the same folder, when you get `FileExistsError` the folder with the already cloned repository will be deleted, because your `to_clone()` causes an exception about the existing `.git`.  
The original version does not execute `rmtree()` when it receives a `FileExistsError`

pmrowla:
Looks like this is because in the original porcelain implementation, the `Repo.init` call happens before the `porcelain.clone` `try/except: rmtree` block, but with the unified behavior the exception in `init` now gets caught by the `try/except`.

I think we can just move the `rmtree` behavior from porcelain into the unified `do_clone`, and structure it so that we only do the `rmtree` if `Repo.init` actually created the target directory (and nothing would get removed if `init` fails due to a pre-existing git repo). This would be consistent with the original porcelain version, but would be a change for  `Repo.clone` (which did not do any `rmtree` at all).

WDYT @jelmer

alchem1ster:
@pmrowla can you please merge `main` branch into this PR to update?
@jelmer what about this PR?

pmrowla:
This has been rebased on latest master and now includes the `rmtree` change described in https://github.com/dulwich/dulwich/pull/851#issuecomment-988927245

dulwich/tests/test_porcelain.py
```diff
@@ -630,9 +630,12 @@ def test_no_head_no_checkout_outstream_errstream_autofallback(self):
         r.close()
 
     def test_source_broken(self):
-        target_path = tempfile.mkdtemp()
-        self.assertRaises(Exception, porcelain.clone, "/nonexistant/repo", target_path)
-        self.assertFalse(os.path.exists(target_path))
+        with tempfile.TemporaryDirectory() as parent:
+            target_path = os.path.join(parent, "target")
```
pmrowla, line 634: With the adjusted `rmtree` behavior, the target repo directory will only be removed when `mkdir=True` (which for porcelain is when the directory does not exist before `porcelain.clone()` is called). So `target_path` can't be the dir generated by `mkdtemp()` `target_path` has to be a child of that dir instead.

I used `TemporaryDirectory` instead of `mkdtemp` since the context manager will clean everything up afterwards whether or not the test succeeds.

pmrowla:
`test_shell_hook_pre_commit_add_files` also fails in master, everything else in CI related to this PR is passing.

jelmer (approved):
I think this could do with some more refactoring (see comments below), but let's land it is as is and iterate rather than letting perfect be the enemy of good enough. Sorry for the delay in merging.

pmrowla:
Thanks for merging this @jelmer. I'll send a follow up PR addressing the new comments when I can, probably not until later this week or early next week though.

PR #852, opened 2021-03-16
client: support deepening/unshallowing refs on fetch

pmrowla:
Currently, dulwich supports shallow clone/fetch from non-local servers, but there is no way to deepen or unshallow via `fetch`. `determine_wants` will see that the tip commit for a given ref is up to date and skip the fetch.

After this PR:

- fetch `determine_wants` now takes an optional `depth` parameter containing fetch depth.
- if depth is set to `DEPTH_INFINITE` (requesting a complete unshallowing of the repo), all refs will be fetched
- otherwise, if depth is set and nonzero, the current available depth for a given ref will be checked, and the ref will be fetched if the requested fetch depth is greater than the current depth.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -502,7 +502,10 @@ class GitClient(object):
 
         """
         if determine_wants is None:
-            determine_wants = target.object_store.determine_wants_all
+            if depth is not None and depth > 1:
+                determine_wants = target.object_store.determine_wants_force
+            else:
+                determine_wants = target.object_store.determine_wants_all
         if CAPABILITY_THIN_PACK in self._fetch_capabilities:
             # TODO(jelmer): Avoid reading entire file into memory and
             # only processing it after the whole file has been fetched.
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -502,10 +502,7 @@ class GitClient(object):
 
         """
         if determine_wants is None:
-            if depth is not None and depth > 1:
-                determine_wants = target.object_store.determine_wants_force
-            else:
-                determine_wants = target.object_store.determine_wants_all
+            determine_wants = target.object_store.determine_wants_all
         if CAPABILITY_THIN_PACK in self._fetch_capabilities:
             # TODO(jelmer): Avoid reading entire file into memory and
             # only processing it after the whole file has been fetched.
@@ -1026,7 +1023,7 @@ class TraditionalGitClient(GitClient):
                 return FetchPackResult(refs, symrefs, agent)
 
             try:
-                wants = determine_wants(refs)
+                wants = determine_wants(refs, depth=depth)
             except BaseException:
                 proto.write_pkt_line(None)
                 raise
@@ -2045,7 +2042,7 @@ class HttpGitClient(GitClient):
             symrefs,
             agent,
         ) = self._negotiate_upload_pack_capabilities(server_capabilities)
-        wants = determine_wants(refs)
+        wants = determine_wants(refs, depth=depth)
         if wants is not None:
             wants = [cid for cid in wants if cid != ZERO_SHA]
         if not wants:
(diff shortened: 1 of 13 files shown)


dulwich/protocol.py
```diff
@@ -107,6 +107,8 @@
     ]
 )
 
+DEPTH_INFINITE = 0x7FFFFFFF
```
pmrowla, line 110: This is documented here: https://git-scm.com/docs/shallow

> The special depth 2147483647 (or 0x7fffffff, the largest positive number a signed 32-bit integer can contain) means infinite depth.

And can also be seen in git source `commit.h`. When `fetch --unshallow` is used, CLI git just calls `git-fetch-pack` with depth set to this value

jelmer: This seems fine, but is unrelated to the rest of the pull request?

dulwich/client.py
```diff
@@ -502,7 +502,10 @@ def fetch(self, path, target, determine_wants=None, progress=None, depth=None):
 
         """
         if determine_wants is None:
-            determine_wants = target.object_store.determine_wants_all
+            if depth is not None and depth > 1:
+                determine_wants = target.object_store.determine_wants_force
```
jelmer, line 506: Do we need to call determine_wants at all in this case, if we're going to want all refs?

jelmer, line 506: determine_wants is meant to be used by the caller to specify what refs they're interested in. What's the point of asking them, if we're going to fetch the ref anyway?

ideally, this should be checking what depth we're currently at and whether that's less than what was requested.

pmrowla: My main issue here was that `porcelain.fetch` does not take a `determine_wants` callback at all, and only uses the default  `object_store.determine_wants_all`. If it would be preferred to update `porcelain.fetch` to accept the callback (and pass it through to `client.fetch` I can go ahead and make that change

jelmer: Perhaps an option would be to add an optional depth argument to determine_wants, which gets specified when there is a depth?

That way, target.object_store.determine_wants() can return refs depending on how deep the history of target.object_store is.

jelmer: (to start off with, you could use the current logic - always fetch if depth is > 1, but that way we can make it more correct later by actually checking the current depth)

dulwich/repo.py
```diff
@@ -698,7 +698,12 @@ def update_shallow(self, new_shallow, new_unshallow):
             shallow.update(new_shallow)
         if new_unshallow:
             shallow.difference_update(new_unshallow)
-        self._put_named_file("shallow", b"".join([sha + b"\n" for sha in shallow]))
+        if shallow:
+            self._put_named_file(
+                "shallow", b"".join([sha + b"\n" for sha in shallow])
+            )
+        else:
+            self._del_named_file("shallow")
```
pmrowla, line 706: CLI git removes this file if it is empty after a `fetch --depth=...` or `fetch --unshallow`

pmrowla:
Just a heads up that I'm on vacation next week, so I'm marking this as a draft.

Will add unit tests for the new behavior and then reopen this for review once I am back.

dulwich/client.py
pmrowla: Updated this so that depth is now passed into `determine_wants`. For the default callback, we will always fetch everything if a complete unshallow is requested via `depth=DEPTH_INFINITE`, otherwise we will check the current available depth for a ref

jelmer:
Anything I can do to help move this along? Changes look reasonable, but tests to verify the new behaviour would be awesome.

pmrowla:
The remaining codecov diff complaints are for places where the kwargs for calling `determine_wants` needed to be updated, but were also not being covered prior to this PR. (The actual new behavior is covered in tests, and there is no overall coverage change after this PR)

PR #853, opened 2021-03-17
porcelain.status: skip traversal of gitignore untracked dirs

pmrowla:
When using `porcelain.get_untracked_paths` with `exclude_ignored`, currently every untracked path is checked to see if it is gitignored. If a directory is gitignored, we can skip checking the remainder of the files inside that directory - from the documentation for [gitignore](https://git-scm.com/docs/gitignore):

> Git doesn’t list excluded directories for performance reasons, so any patterns on contained files have no effect, no matter where they are defined.

Changes after this PR:

- `porcelain.get_untracked_files`: no longer walks ignored directories for performance reasons. The path to the ignored dir (including trailing slash) will be yielded if `exclude_ignored=False`, but individual files within the directory will not be included.
- `porcelain.add`: now only returns path to ignored directory in the (`added`, `ignored`) tuple (previously all files inside ignored dirs would be returned)
- `porcelain.status`: now only returns path to ignored directory when called with `ignored=True` (previously all files inside ignored dirs would be returned)

### DIFF
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -1678,7 +1678,7 @@ class StatusTests(PorcelainTestCase):
             set(porcelain.status(self.repo, ignored=True).untracked),
         )
 
-    def test_get_untracked_paths_nested(self):
+    def test_get_untracked_paths_subrepo(self):
         with open(os.path.join(self.repo.path, ".gitignore"), "w") as f:
             f.write("nested/\n")
         with open(os.path.join(self.repo.path, "notignored"), "w") as f:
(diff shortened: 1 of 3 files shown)


pmrowla:
porcelain.add could probably also benefit from this?

Although the current behavior is to return the tuple of `(added, ignored)` where ignored contains all ignored paths, so it would need to be changed to return ignored directory paths instead of the files inside the directory

jelmer:
The changes look good - any chance you could add a test to make sure this doesn't regress?

jelmer:
I think it would make sense to change the behaviour of add as well - we should make sure to document the change in behaviour in NEWS.

dulwich/tests/test_porcelain.py
```diff
@@ -1705,32 +1720,66 @@ def test_get_untracked_paths_nested(self):
...
+
+        self.assertEqual(
+            set(
+                [
+                    ".gitignore",
+                    "notignored",
+                    "ignored",
+                    os.path.join("subdir", ""),
+                ]
+            ),
```
pmrowla, line 1764: This test should catch any future regression, if we recurse into the ignored `subdir` here, we will end up yielding `subdir/ignored` as well.

pmrowla:
@jelmer this is also ready for review

PR #859, opened 2021-03-29
Fix two bugs in signed tag support and add ability to specify a key id to use for signing

DanHatesNumbers:
This PR fixes two compatibility issues with git around signed tags, and adds support for specifying a signing Key ID to use when creating a signed tag.

The first issue is the serialisation logic for tags. When git creates a signed tag, it includes a new line at the end of the buffer being signed. If this isn't included, then git will not parse the tag signature and if it's included when serialising the tag signature itself, git will be performing signature verification on different data to what was signed, leading to signature verification failure.

The second issue this PR fixes is with the GPG signature mode used. By specifying that a detached signature is being made, Dulwich will create the same format of tag signature as git (see https://github.com/git/git/blob/84d06cdc06389ae7c462434cb7b1db0980f63860/gpg-interface.c#L453 relevant option is -b, which tells GPG to make a detached signature).

Lastly I've added support for overloading the sign parameter to porcelain.tag_create, which when passed a str containing a key id will use that key id to perform signing.

Because there aren't existing tests for tag signing, I'm not sure how to structure these, any advice would be greatly appreciated

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -35,7 +35,7 @@ jobs:
       run: sudo apt-get update && sudo apt-get install -y libgpgme-dev libgpg-error-dev
       if: "matrix.os == 'ubuntu-latest'"
     - name: Install native dependencies (MacOS)
-      run: brew install gpgme
+      run: brew install swig gpgme
       if: "matrix.os == 'macos-latest'"
     - name: Install dependencies
       run: |
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -32,7 +32,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install native dependencies (Ubuntu)
-      run: apt update && apt install -y libgpgme libgpg-error-dev
+      run: sudo apt-get update && sudo apt-get install -y libgpgme libgpg-error-dev
       if: "matrix.os == 'ubuntu-latest'"
     - name: Install native dependencies (MacOS)
       run: brew install gpgme
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -31,6 +31,9 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install native dependencies (Ubuntu)
+      run: apt update && apt install -y libgpgme libgpg-error-dev
+      if: "matrix.os == 'ubuntu-latest'"
     - name: Install dependencies
       run: |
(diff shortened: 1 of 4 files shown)


DanHatesNumbers:
It looks like my change for tag serialisation fixes tag signing, but breaks everything else related. Converted to draft while I investigate further

jelmer:
I suspect what you want to do is add an extra newline to the tag message when you're creating the tag, not on serialization.

DanHatesNumbers:
> I suspect what you want to do is add an extra newline to the tag message when you're creating the tag, not on serialization.

You're absolutely correct! Just made that change and everything is working and I can verify a tag created with Dulwich using git. Going to try and figure out how to sort out test coverage and I'll mark as ready for review when that's finished.

DanHatesNumbers:
Update: in order to run the gpg tests, the gpgme package needs to be installed on the actions runner. Need to dig into this further to make sure it can be setup on macos and windows.

DanHatesNumbers:
@jelmer I'm not sure why the PyPy build is failing to import gpg, how to tackle the issue of Windows needing libgpgme or how to fix the small decrease in project coverage :/

DanHatesNumbers:
One possible option would be to switch from the gnupg python wrapper of gpgme, which requires users to also install libgpgme (which becomes more complex on Windows) to either the [cryptography](https://cryptography.io/en/latest/index.html) package or [pretty-bad-protocol](https://github.com/isislovecruft/python-gnupg), both of which appear to provide statically linked builds and support PyPy.

If you're open to me trying either of these modules as a replacement for the current gpg module @jelmer, I'd be happy to raise the change as a separate PR or include it in this one.

jelmer:
> One possible option would be to switch from the gnupg python wrapper of gpgme, which requires users to also install libgpgme (which becomes more complex on Windows) to either the [cryptography](https://cryptography.io/en/latest/index.html) package or [pretty-bad-protocol](https://github.com/isislovecruft/python-gnupg), both of which appear to provide binary wheels (to make installing on windows easy) and support PyPy.
> 
> If you're open to me trying either of these modules as a replacement for the current gpg module @jelmer, I'd be happy to raise the change as a separate PR or include it in this one.
Perhaps it would make sense to do that as a separate change? We can skip the tests on the platform where gpgme isn't easily available, but at least means that we can get your fix in. And then we can consider other libraries?

DanHatesNumbers:
That sounds sensible to me. I'll push a change when I get a chance tomorrow to skip the tests on PyPy and Windows and then if either of the alternative libraries look like a promising option they could be re-enabled in that followup PR

jelmer:
I missed that this introduces a dependency on C Git in the main testsuite :( We'll have to resolve that before the next release.

PR #862, opened 2021-04-05
Add LICENSE

timmy61109:


jelmer:
The license file is already present as COPYING.

PR #864, opened 2021-04-11
Performance of porcelain status

POD666:
Looks like `porcelain.status()` stuck at my `node_modules`. I took a brief look and looks like the following changes work for me.

You might also want to use it in other places like `clean` and `add`.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -992,17 +992,14 @@ def status(repo=".", ignored=False):
             get_unstaged_changes(index, r.path, filter_callback)
         )
         ignore_manager = IgnoreFilterManager.from_repo(r)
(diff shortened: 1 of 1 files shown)


jelmer:
Did you mean to propose this against master? 0.19 is no longer maintained.

jelmer:
Also /cc @pmrowla  as somebody who has touched this code recently.

pmrowla:
@POD666 have you tried the latest master branch? This looks like the same issue which should be fixed in https://github.com/dulwich/dulwich/pull/853

POD666:
@pmrowla, you are right, that PR looks very similar.

I stay on 0.19.* as it seems to support py2.

PR #866, opened 2021-04-29
config: check default system paths on windows

### LINKED ISSUE #865: porcelain.status(): reports unmodified files as containing unstaged changes on Windows
On Windows, `porcelain.status()` shows all files as changed even when the working tree and index should both be clean.

```
(.venv-win) PS C:\users\vagrant\example-get-started> git ls-files
.dvc/.gitignore
.dvc/config
.dvc/plots/confusion.json
.dvc/plots/confusion_normalized.json
.dvc/plots/default.json
.dvc/plots/linear.json
.dvc/plots/scatter.json
.dvc/plots/smooth.json
.dvcignore
.gitignore
README.md
data/.gitignore
data/data.xml.dvc
dvc.lock
dvc.yaml
params.yaml
prc.json
roc.json
scores.json
src/evaluate.py
src/featurization.py
src/prepare.py
src/requirements.txt
src/train.py
(.venv-win) PS C:\users\vagrant\example-get-started> git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
(.venv-win) PS C:\users\vagrant\example-get-started> pip freeze
certifi==2020.12.5
dulwich==0.20.21
urllib3==1.26.4
(.venv-win) PS C:\users\vagrant\example-get-started> python
Python 3.9.1 (tags/v3.9.1:1e5d33e, Dec  7 2020, 17:08:21) [MSC v.1927 64 bit (AMD64)] on win32
>>> porcelain.status(".")
GitStatus(staged={'add': [], 'delete': [], 'modify': []}, unstaged=[b'.dvc/.gitignore', b'.dvc/config', b'.dvc/plots/confusion.json', b'.dvc/plots/confusion_normalized.json', b'.dvc/plots/default.json', b'.dvc/plots/linear.json', b'.dvc/plots/scatter.json', b'.dvc/plots/smooth.json', b'.dvcignore', b'.gitignore', b'README.md', b'data/.gitignore', b'data/data.xml.dvc', b'dvc.lock', b'dvc.yaml', b'params.yaml', b'prc.json', b'roc.json', b'scores.json', b'src/evaluate.py', b'src/featurization.py', b'src/prepare.py', b'src/requirements.txt', b'src/train.py'], untracked=[])

```

For reference, this is using a completely fresh clone (via CLI `git clone`) of https://github.com/iterative/example-get-started. I tested in `0.20.21` and current master and got the same result.

pmrowla:
Will close #865 

from https://github.com/dulwich/dulwich/issues/865#issuecomment-828274471
> I guess checking for the system config on windows isn't exactly straightforward since it depends on where the user installed Git (it will be in `<path-to-git-installation>/etc/gitconfig`
> 
> But it reasonable for dulwich to at least check `C:\Program Files\Git\etc\gitconfig` (current default Git for Windows install location) and `C:\ProgramData\Git\config` (alternate system config location in vista+ from the [git docs](https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup))

### DIFF
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -498,6 +498,57 @@ def get_xdg_config_home_path(*path_segments):
     return os.path.join(xdg_config_home, *path_segments)
 
 
+def _find_git_in_win_path():
+    for exe in ("git.exe", "git.cmd"):
+        for path in os.environ.get("PATH", "").split(";"):
+            if os.path.exists(os.path.join(path, exe)):
+                # exe path is .../Git/bin/git.exe or .../Git/cmd/git.exe
+                git_dir, _bin_dir = os.path.split(path)
+                yield git_dir
+                break
+
+
+def _find_git_in_win_reg():
+    import platform
+    import winreg
+
+    if platform.machine() == "AMD64":
+        subkey = (
+            "SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\"
+            "CurrentVersion\\Uninstall\\Git_is1"
+        )
+    else:
+        subkey = (
+            "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\"
+            "Uninstall\\Git_is1"
+        )
+
+    for key in (winreg.HKEY_CURRENT_USER, winreg.HKEY_LOCAL_MACHINE):
+        try:
+            with winreg.OpenKey(key, subkey) as k:
+                val, typ = winreg.QueryValueEx(k, "InstallLocation")
+                if typ == winreg.REG_SZ:
+                    yield val
+        except OSError:
+            pass
+
+
+# There is no set standard for system config dirs on windows. We try the
+# following:
+#   - %ProgramData%/Git/config - (deprecated) Windows config dir per CGit docs
+#   - %ProgramFiles%/Git/etc/gitconfig - Git for Windows (msysgit) config dir
+#     Used if CGit installation (Git/bin/git.exe) is found in PATH in the
+#     system registry
+def get_win_system_paths():
+    if "ProgramData" in os.environ:
+        yield os.path.join(os.environ["ProgramData"], "Git", "config")
+
+    for git_dir in _find_git_in_win_path():
+        yield os.path.join(git_dir, "etc", "gitconfig")
+    for git_dir in _find_git_in_win_reg():
+        yield os.path.join(git_dir, "etc", "gitconfig")
+
+
 class StackedConfig(Config):
     """Configuration which reads from multiple config files.."""
 
@@ -525,12 +576,7 @@ class StackedConfig(Config):
         if "GIT_CONFIG_NOSYSTEM" not in os.environ:
             paths.append("/etc/gitconfig")
             if sys.platform == "win32":
-                paths.extend(
-                    [
-                        "C:/Program Files/Git/etc/gitconfig",
-                        "C:/ProgramData/Git/config",
-                    ]
-                )
+                paths.extend(get_win_system_paths())
 
         backends = []
         for path in paths:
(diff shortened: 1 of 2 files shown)


pmrowla:
linter (flake8) is failing in files unrelated to this PR

dulwich/config.py
```diff
@@ -524,6 +524,13 @@ def default_backends(cls):
 
         if "GIT_CONFIG_NOSYSTEM" not in os.environ:
             paths.append("/etc/gitconfig")
+            if sys.platform == "win32":
+                paths.extend(
+                    [
+                        "C:/Program Files/Git/etc/gitconfig",
```
jelmer, line 530: I'm not a big fan of peeking at another applications' directory.

https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup suggests C:\Documents and Settings\All Users\Application Data\Git\config and C:/ProgramData/Git/config instead. Can we use those?

pmrowla: The `Application Data\...` path is listed as the default in XP so I wasn't sure if it was worth including.

My concern here is that the listed paths in the git docs aren't up to date with the current Git for Windows installer. If you download the latest version from https://git-scm.com/download/win and do a fresh installation with all of the default settings, it puts the system config (containing `core.autocrlf=true`) in `C:/Program Files/Git/etc/gitconfig` and does not generate any `C:/ProgramData/Git/` directory at all.

I understand that dulwich is intended to work without relying on CGit, but in the event that I do have a default CGit installation on windows, I would expect dulwich to read the system `autocrlf` setting from the default config location.

pmrowla: The default system config location was changed away from `C:/ProgramData` here: https://github.com/git-for-windows/git/pull/2358

jelmer: I'm not necessarily opposed to this merge proposal, but I'd like to have a concise strategy - if we change the behaviour now people will start depending on it.

The other question is what to do in the case that we don't have a CGit installation. Should we have a Dulwich-specific global configuration file? What does e.g. libgit2 read?

pmrowla: libgit2 checks to see if `git.exe` is in the current PATH, or if certain registry keys set by msysgit exist, and then sets the system config dir based on that path or installation dir

https://github.com/libgit2/libgit2/blob/508361401fbb5d87118045eaeae3356a729131aa/src/win32/findfile.c#L168

and they also check for the `%PROGRAMDATA%` (`C:/ProgramData/Git`) dir mentioned in the git docs

pmrowla: They document the behavior for `git_config_find_system()` as
```
 * If /etc/gitconfig doesn't exist, it will look for
 * %PROGRAMFILES%\Git\etc\gitconfig.
 ```

pmrowla: Not sure that dulwich needs to go as far as checking registry keys, but looking for CGit in the PATH seems like a reasonable solution to me

jelmer: I'm a little annoyed by the move away from %PROGRAMDATA%, since that was an easy to find location for a bit of config that is shared between multiple Git applications/libraries. Anyway, that's hard to revert now. This way we'll end up searching for msysgit2, C Git for Windows, libgit2 installations and the behaviour becomes more opaque and unpredictable to the user.

I do think we should rely on registry settings if we're going to find the C Git installation. E.g. on localized systems the paths might vary. It might also be nice to e.g. provide a global list of paths that will be tried (a list of callables) so that Dulwich users can register/tweak git installation find functions - but not necessary for this PR.

pmrowla: after 0530d81 behavior in this PR should be consistent with libgit2 (try all of %PROGRAMDATA%, CGit from PATH, CGit from registry)

PR #867, opened 2021-04-29
line_ending: fix autocrlf=input handling

### LINKED ISSUE #693: Files with CRLF and git configured to convert those to LF leads to incorrect porcelain status

pmrowla:
This is a rebased follow up to https://github.com/dulwich/dulwich/pull/704

Should close #693

To summarize:
- with `autocrlf=true`, normalization should be done all the time (CRLF->LF when staging into index, LF->CRLF when checking out index into working tree)
- with `autocrlf=input`, only CRLF->LF normalization is only done when staging into index
  - in this case CGit only does normalization on files which are newly added. Since existing files are not normalized at all on checkout into working tree, these files also remain unmodified when staging modifications to those existing files.
  - this is a bit more complex than just only doing normalization on files which do not yet exist in the index - the original PR did not account for the case where you add (stage) a new file, then make further modifications to that new file.
  - essentially, the check for a "new file" needs to be based on whether or not the file being staged exists in the original HEAD tree, not whether or not it currently exists in the index.

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -83,7 +83,7 @@ from dulwich.hooks import (
     PostReceiveShellHook,
 )
 
-from dulwich.line_ending import BlobNormalizer
+from dulwich.line_ending import BlobNormalizer, TreeBlobNormalizer
 
 from dulwich.refs import (  # noqa: F401
     ANNOTATED_TAG_SUFFIX,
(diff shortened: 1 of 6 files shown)


dulwich/repo.py
```diff
@@ -1532,7 +1532,17 @@ def get_blob_normalizer(self):
         """Return a BlobNormalizer object"""
         # TODO Parse the git attributes files
         git_attributes = {}
-        return BlobNormalizer(self.get_config_stack(), git_attributes)
+        config_stack = self.get_config_stack()
+        try:
+            tree = self.object_store[self.refs[b"HEAD"]].tree
```
pmrowla, line 1537: this was also discussed in the original PR https://github.com/dulwich/dulwich/pull/704#discussion_r275740141

rather than comparing paths to what's in the `HEAD` tree, we could check what's in the index's blobs instead (if the index already has CRLF line endings in the existing blob, then assume we aren't supposed to convert it)

but that doesn't seem too much cleaner of a solution to me

PR #869, opened 2021-05-18
objects.Tag: add verify() for gpg signature verification

pmrowla:
- `tag.verify()` is used in porcelain tests and replaces CGit calls
- Checks that dulwich & CGit can verify each others' signed tags moved to `tests.compat.test_porcelain`

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -881,6 +881,23 @@ class Tag(ShaFile):
                     self.as_raw_string(), mode=gpg.constants.sig.mode.DETACH
                 )
 
+    def verify(self):
+        """Verify GPG signature for this tag (if it is signed).
+
+        Raises:
+          gpg.errors.BadSignatures: if GPG signature verification fails
+        """
+        if self._signature is None:
+            return
+
+        import gpg
+
+        with gpg.Context() as ctx:
+            unused_data, unused_result = ctx.verify(
+                self.as_raw_string()[: -len(self._signature)],
+                self._signature,
+            )
+
 
 class TreeEntry(namedtuple("TreeEntry", ["path", "mode", "sha"])):
     """Named tuple encapsulating a single tree entry."""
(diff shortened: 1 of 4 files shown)


dulwich/objects.py
```diff
@@ -881,6 +881,23 @@ def sign(self, keyid: Optional[str] = None):
...
+        Raises:
+          gpg.errors.BadSignatures: if GPG signature verification fails
+        """
+        if self._signature is None:
+            return
+
+        import gpg
+
+        with gpg.Context() as ctx:
+            unused_data, unused_result = ctx.verify(
```
jelmer, line 896: AFAICT this just verifies that the tag has a valid signature, but doesn't provide a way to verify that it's signed with the right key?

dulwich/tests/compat/test_porcelain.py
```diff
@@ -0,0 +1,97 @@
+# test_utils.py -- Tests for git compatibility utilities
```
jelmer, line 1: Please update the header.

dulwich/objects.py
pmrowla: By right key do you mean checking that the email on the key matches the commit author?

pmrowla: I'll double check and add a test case, but if it's signed by a key which has not been imported into gpg (meaning its untrusted) then my understanding is that `ctx.verify` will fail due to `GPG_ERR_NO_PUBKEY`?

jelmer: The key being imported locally doesn't imply that it's trusted, or trusted for this particular purpose. Verifying that the signature was made by the expected key could either be done by allowing the caller to pass in key ids into ctx.verify() or by return the resultant key ids from Tag.verify().

pmrowla:
test failures are unrelated

```
Run python -m flake8
./examples/rename-branch.py:16:1: E302 expected 2 blank lines, found 1
Error: Process completed with exit code 1.
```

PR #870, opened 2021-05-18
tests: fix windows crlf CI issues

pmrowla:
Fixes windows specific CI test failures that resulted from the crlf & windows system config changes

Now that dulwich reads CGit global config on windows, it now picks up `autocrlf = True` from the system wide config that was ignored before. This made some windows autocrlf test bugs show up (the tests assumed autocrlf was always disabled by default on all platforms which is no longer the case)

### DIFF
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -1820,11 +1820,6 @@ class StatusTests(PorcelainTestCase):
         with open(file_path, "wb") as f:
             f.write(b"line1\r\nline2")
 
-        if sys.platform == "win32":
-            c = self.repo.get_config()
-            c.set("core", "autocrlf", "false")
-            c.write_to_path()
-
         results = porcelain.status(self.repo)
         self.assertDictEqual({"add": [], "delete": [], "modify": []}, results.staged)
         self.assertListEqual(results.unstaged, [b"crlf"])
@@ -1861,10 +1856,6 @@ class StatusTests(PorcelainTestCase):
 
     def test_status_autocrlf_input(self):
         # Commit existing file with CRLF
-        c = self.repo.get_config()
-        c.set("core", "autocrlf", "false")
-        c.write_to_path()
-
         file_path = os.path.join(self.repo.path, "crlf-exists")
         with open(file_path, "wb") as f:
             f.write(b"line1\r\nline2")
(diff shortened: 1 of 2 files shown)


dulwich/tests/test_porcelain.py
```diff
@@ -1820,6 +1820,11 @@ def test_status_crlf_mismatch(self):
         with open(file_path, "wb") as f:
             f.write(b"line1\r\nline2")
 
+        if sys.platform == "win32":
+            c = self.repo.get_config()
+            c.set("core", "autocrlf", "false")
+            c.write_to_path()
```
pmrowla, line 1826: On windows, this case is only a mismatch if autocrlf is false (if it's true/auto, then `\r\n` in the working dir copy is expected and would be normalized to `\n` in the status call)

dulwich/tests/test_porcelain.py
```diff
@@ -1856,6 +1861,10 @@ def test_status_autocrlf_true(self):
 
     def test_status_autocrlf_input(self):
         # Commit existing file with CRLF
+        c = self.repo.get_config()
+        c.set("core", "autocrlf", "false")
+        c.write_to_path()
```
pmrowla, line 1866: We want to explicitly commit the `\r\n` here, autocrlf needs to be disabled

jelmer:
> Fixes windows specific CI test failures that resulted from the crlf & windows system config changes
> 
> Now that dulwich reads CGit global config on windows, it now picks up `autocrlf = True` from the system wide config that was ignored before. This made some windows autocrlf test bugs show up (the tests assumed autocrlf was always disabled by default on all platforms which is no longer the case)

The tests should run isolated; can we prevent reading of the global config?

> Fixes windows specific CI test failures that resulted from the crlf & windows system config changes
> 
> Now that dulwich reads CGit global config on windows, it now picks up `autocrlf = True` from the system wide config that was ignored before. This made some windows autocrlf test bugs show up (the tests assumed autocrlf was always disabled by default on all platforms which is no longer the case)

I think the longer-term fix here is to prevent the test isolation leak - we shouldn't be reading the system global configuration when running the tests (and already take measures to avoid reading the user home directory). Can we set GIT_CONFIG_NOSYSTEM in dulwich/tests/__init__.py ?

PR #883, opened 2021-06-12
Add local .git/config support

0x2b3bfa0:
This pull request allows Dulwich to load local configuration from `.git/config` through `StackedConfig.default()` by inferring the configuration file path from either the `GIT_DIR` environment variable or the current directory. The added functions are simplified implemetations of the following Git functions:

* `_parse_git_file` ➔ [`read_gitfile_gently`](https://github.com/git/git/blob/211eca0895794362184da2be2a2d812d070719d3/setup.c#L720-L803)
* `_is_git_directory` ➔ [`is_git_directory`](https://github.com/git/git/blob/211eca0895794362184da2be2a2d812d070719d3/setup.c#L328-L377)
* `_find_git_directory` ➔ [`setup_git_directory_gently_1`](https://github.com/git/git/blob/211eca0895794362184da2be2a2d812d070719d3/setup.c#L1042-L1151)

I don't know if taking into account the current working directory is too unpredictable for non-porcelain code, though. Any feedback would be highly appreciated.

### DIFF
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -29,7 +29,8 @@ TODO:
 import os
 import sys
 
-from typing import BinaryIO, Tuple, Optional
+from pathlib import Path
+from typing import BinaryIO, Iterator, Tuple, Optional
 
 from collections import (
     OrderedDict,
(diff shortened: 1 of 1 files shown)


jelmer:
Thanks for the pull request.

The three helper functions belong in dulwich.repo, since they're not config-specific. The functionality already exists today, so if you factor out into separate functions please update the existing code to use those functions as well.

Config.default_backends() intentionally doesn't use try to load the local repository configuration. See Repo.get_config_stack() for that.

0x2b3bfa0:
🙏🏼  Thank you very much for your reply!

> _The three helper functions belong in dulwich.repo, since they're not config-specific. The functionality already exists today, so if you factor out into separate functions please update the existing code to use those functions as well._

I've noticed that [`read_gitfile`](https://github.com/dulwich/dulwich/blob/5559440256dc231637896d97f1016f9052cc7dc3/dulwich/repo.py#L1010-L1022) already exists on [`dulwich/repo.py`](https://github.com/dulwich/dulwich/blob/5559440256dc231637896d97f1016f9052cc7dc3/dulwich/repo.py) and [`Repo.__init__`](https://github.com/dulwich/dulwich/blob/5559440256dc231637896d97f1016f9052cc7dc3/dulwich/repo.py#L1041-L1107) replicates a considerable amount of the proposed functionality. If any variant of this pull request gets accepted, I'll have to update the code to avoid duplication as you suggest.

> _`Config.default_backends()` intentionally doesn't use try to load the local repository configuration. See `Repo.get_config_stack()` for that._

As per https://github.com/dulwich/dulwich/pull/638#issuecomment-400055422, `dulwich.porcelain.<command>` should work like `git <command>` [...] including behaviour around `os.getcwd` 🤔  If `Config.default_backends()` is not intended to replicate the functionality of [`git-config`](https://git-scm.com/docs/git-config) as suggested in the code, where should this logic be implemented?

jelmer:
On Sat, Jun 12, 2021 at 09:04:47AM -0700, Helio Machado wrote:
> > _`Config.default_backends()` intentionally doesn't use try to load the local repository configuration. See `Repo.get_config_stack()` for that._
> As per https://github.com/dulwich/dulwich/pull/638#issuecomment-400055422, `dulwich.porcelain.<command>` should work like `git <command>` [...] including behaviour around `os.getcwd` 🤔  If `Config.default_backends()` is not intended to replicate the functionality of [`git-config`](https://git-scm.com/docs/git-config) as suggested in the code, where should this logic be implemented?

The intended code path for obtaining the config for a repository is Repo('.').get_config_stack().

-- 
Jelmer Vernooĳ ***@***.***>

0x2b3bfa0:
Understood! This pull request doesn't make any sense at all: `Repo.discover().get_config_stack()` does exactly the same, and it would just be matter of connecting it to the `porcelain` functions. 🤦🏼  Sorry for the noise.

PR #887, opened 2021-07-18
use list() to change fs_paths to list

DedSecer:
Using `list()` to format fs_paths will not only access type `str` but also access type `tuple` and type `set`.

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -1263,7 +1263,7 @@ class Repo(BaseRepo):
         root_path_bytes = os.fsencode(self.path)
 
         if not isinstance(fs_paths, list):
-            fs_paths = [fs_paths]
+            fs_paths = list(fs_paths)
         from dulwich.index import (
             blob_from_path_and_stat,
             index_entry_from_stat,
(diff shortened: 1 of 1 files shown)


jelmer:
The existing code is meant to e.g. allow a string to be passed in; this PR would change the behaviour in that case.

jelmer (changes requested):
The existing code is meant to e.g. allow a string to be passed in; this PR would change the behaviour in that case.

DedSecer:
I think it also work using `list()`(both python2 and python3).
```
>>> list('s')
['s']
```

jelmer:
It works, but not necessarily in the way you expect:

```
>>> list('foo')
['f', 'o', 'o']
```

DedSecer:
Sorry.I have pushed a new commit.

jelmer (approved):

DedSecer:
by the way.I hope it can be merge into branch 0.19(for python2)

jelmer:
That branch really only exists for critical bug fixes. Why do you need Python 2?

DedSecer:
I have some scripts using python2.

PR #888, opened 2021-07-20
use list() to change fs_paths to list

DedSecer:


jelmer:
Sorry, Python 2 is no longer maintained and neither is the 0.19 branch. If there are e.g. security fixes I might consider applying those to 0.19 but other changes (such as this) would certainly not qualify.

Issue #889, opened 2021-07-25
porcelain.status() output different type of file names.


DedSecer:
porcelain.status() output untracked file as string when others output as bytes.
It looks like:
```
>>> porcelain.status('./')
GitStatus(staged={'add': [b'a'], 'delete': [b'README.md'], 'modify': [b'demo.py']}, unstaged=[b'requirements.txt'], untracked=['b'])
```

jelmer:
On Sat, Jul 24, 2021 at 05:15:21PM -0700, DedSecer wrote:
> porcelain.status() output untracked file as string when others output as bytes.
> It looks like:
> ```
> >>> porcelain.status('./')
> GitStatus(staged={'add': [b'a'], 'delete': [b'README.md'], 'modify': [b'demo.py']}, unstaged=[b'requirements.txt'], untracked=['b'])
> ```
This is sort of intentional; untracked files may not be representable
in the git index encoding.

We could consider consistently returning strings here (rather than
bytestrings).

Jelmer

-- 
Jelmer Vernooĳ ***@***.***>
PGP Key: https://www.jelmer.uk/D729A457.asc

PR #890, opened 2021-07-29
update porcelain.status to output string

DedSecer:
Issues #889

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -1210,11 +1210,14 @@ def status(repo=".", ignored=False):
     with open_repo_closing(repo) as r:
         # 1. Get status of staged
         tracked_changes = get_tree_changes(r)
+        for key in tracked_changes.keys():
+            tracked_changes[key] = [file.decode() for file in tracked_changes[key]]  #decode
         # 2. Get status of unstaged
         index = r.open_index()
         normalizer = r.get_blob_normalizer()
         filter_callback = normalizer.checkin_normalize
         unstaged_changes = list(get_unstaged_changes(index, r.path, filter_callback))
+        unstaged_changes = [file.decode() for file in unstaged_changes]  # decode
 
         untracked_paths = get_untracked_paths(
             r.path, r.path, index, exclude_ignored=not ignored


jelmer:
Unfortunately it's not as simple as this - this changes the API of porcelain.status(), breaking any existing callers.

jelmer:
This still changes the API - any code in applications that use dulwich that expected bytestrings will now receive plain strings and break.

DedSecer:
So shouldn’t we change the porcelain.status?

dulwich/porcelain.py
```diff
@@ -1210,11 +1210,14 @@ def status(repo=".", ignored=False):
...
         # 1. Get status of staged
         tracked_changes = get_tree_changes(r)
+        for key in tracked_changes.keys():
+            tracked_changes[key] = [file.decode() for file in tracked_changes[key]]
         # 2. Get status of unstaged
         index = r.open_index()
         normalizer = r.get_blob_normalizer()
         filter_callback = normalizer.checkin_normalize
         unstaged_changes = list(get_unstaged_changes(index, r.path, filter_callback))
+        unstaged_changes = [file.decode() for file in unstaged_changes]
```
jelmer, line 1220: This is going to decode with the current encoding (e.g. utf-8) which may not work if the filename contains bytes that can't be decoded with the current encoding (e.g. utf-8) or even if they can it may not result in the string you were expecting (file system encoding and display encoding may be different).

jelmer:
We need to figure out a path forward - if we change this, it's going break everybody who relies on this API.

There are a couple of possibilities:

* We break it, but do so at a major release and announce it as part of the API changes
* We add a flag to enable the new behaviour; this is a little bit icky since we'll have to support both behaviours going forward
* We can add another status() function that has the new behaviour

dulwich/porcelain.py
DedSecer: Can we set the default encoding in the script which using dulwich to resolve it?

jelmer: That would be inconsistent with the way the rest of Dulwich works. And if you're making the caller worry about the encoding, why not let them do the decoding as well?

jelmer: For some uses cases, bytes are sufficient. You also can choose what to do in that case if you want to display the output - you can ignore invalid characters, replace them with surrogates or e.g. use chardet to try to figure out the proper encoding if it isn't utf8.

jelmer: I'd be open to returning a surrogateescape-decoded string here, since it's possible for the caller to undo that and retrieve the original bytes but that would still be a breaking change for existing users of the API. We should also force utf8 in that case (since that's the convention for git repositories) rather than whatever the users' encoding is.

DedSecer: I think the function in porcelain should work like the git subcommands(a higher function), or I should decode the bytes after using `porcelain.status`.

PR #898, opened 2021-08-14
porcelain.checkout

DedSecer:
Improve from [ pythonista stash's git cli](https://github.com/ywangd/stash/blob/master/bin/git.py)
This seemed to work in my simple testing, but I am not sure about if it actually work as `git checkout [branch]`


dulwich/porcelain.py
```diff
@@ -1753,6 +1755,136 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+                print('file not in index.', path.decode())
+            return
+
+        try:
+            index_entry = list(index[path])
+        except KeyError:
+            # if index_entry doesnt exist, this file was being removed. readd it
+            # update index entry stats to reflect commit
+            index_entry = [(0, 0), (0,0), 15, 0, 0, 1000, 1000, 0, tree_entry[1], 0, 0]
+        index_entry[0] = (repo[b'HEAD'].commit_time, 0)  #ctime
```
jelmer, line 1785: index_entry is a namedtuple nowadays, so you should be able to use index_entry.ctime here - which is slightly more readable.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,136 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+            except KeyError:
+                print('file not in index.', path.decode())
+            return
+
+        try:
+            index_entry = list(index[path])
+        except KeyError:
+            # if index_entry doesnt exist, this file was being removed. readd it
+            # update index entry stats to reflect commit
+            index_entry = [(0, 0), (0,0), 15, 0, 0, 1000, 1000, 0, tree_entry[1], 0, 0]
```
jelmer, line 1784: This is harcoding several values that don't necessarily make sense, e.g. uid/gid.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,136 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+        tree_id = repo[b'HEAD']._tree
+        try:
+            tree_entry = repo[tree_id].lookup_path(lambda x: repo[x], path)
+        except KeyError:
+            # if tree_entry didnt exist, this file was being added, so remove index entry
+            try:
+                del index[path]
+                index.write()
+            except KeyError:
+                print('file not in index.', path.decode())
```
jelmer, line 1776: this should probably raise an exception?

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,136 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+            par_dir_tree_id = get_entry(repo, tree, os.path.split(file_path)[0])[1]
+            tree = parse_tree(repo, treeish=par_dir_tree_id)
+            tree_id = tree[os.path.split(file_path)[1].encode()]
+            return tree_id
+        else:
+            try:
+                return tree[file_path.encode()]
+            except KeyError:
+                # untracked file not in tree.
+                print('file not in tree')
```
jelmer, line 1843: this should probably raise an exception rather than write to standard out (which may not be visible). Another option would be for the function to return the list of paths for files it actually changed.

jelmer (changes requested):
Thanks for the PR. A couple of high-level comments:

* Please add unit tests for any functionality you add (in this case, in dulwich/tests/test_porcelain.py)
* Dulwich uses a particular style for docstrings (Google style docstrings), this means e.g. double quotes rather than single quotes

dulwich/porcelain.py
DedSecer: Where can I get the other value in index_entry such as uid, gid etc. ? And it is necessary to set them?

dulwich/porcelain.py
DedSecer: I have considered about it. But I am confusing about which type of Error to raise.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,136 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+            # update index entry stats to reflect commit
+            index_entry = [(0, 0), (0,0), 15, 0, 0, 1000, 1000, 0, tree_entry[1], 0, 0]
+        index_entry[0] = (repo[b'HEAD'].commit_time, 0)  #ctime
+        index_entry[1] = (repo[b'HEAD'].commit_time, 0)  #mtime
+        index_entry[4] = tree_entry[0]  #mode
+        index_entry[7] = len(repo[tree_entry[1]].data)  #size
+        index_entry[8] = tree_entry[1]  #sha
+
+        index[path] = index_entry
+        index.write()
```
jelmer, line 1791: You probably want to write the index outside of the loop, rather than once for every entry.

dulwich/porcelain.py
jelmer: Either KeyError or FileNotFoundError seem like they could be appropriate, or we could define a new exception type.

dulwich/porcelain.py
DedSecer: OK.That's my mistake.

dulwich/porcelain.py
DedSecer: It will raise KeyError in the new commit

dulwich/porcelain.py
DedSecer: OK,It will raise KeyError in the new commit.

dulwich/porcelain.py
jelmer: We could set them to the current users' identity, or alternatively just to 0 (which may be easiest).

DedSecer:
Do you think if it will be better to move the function `unstage` from `porcelain` under `dulwich.repo.Repo`?

jelmer:
> Do you think if it will be better to move the function `unstage` from `porcelain` under `dulwich.repo.Repo`?

We should keep an eye on the number of methods on Repo, since it's already quite large. But adding Repo.unstage makes sense I think.

DedSecer:
I have added unittest for `stage`, `reset_file` and `checkout` now.

dulwich/errors.py
```diff
@@ -84,6 +84,10 @@ class NotBlobError(WrongObjectException):
     type_name = "blob"
 
 
+class CheckoutError(Exception):
```
jelmer, line 87: Can you give this a less generic name? There are all kinds of errors that can occur during checkout.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
             r.refs.set_symbolic_ref(b"HEAD", to_set)
 
 
+def reset_file(repo, file_path: str , target: bytes = b'HEAD'):
+    """
+    reset the file to specific commit or branch
```
jelmer, line 1760: Per google doc string style and the rest of Dulwich, please Put the summary immediately after the triple quotes, capitalize and end with a trailing period.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+    Args:
+        repo: dulwich Repo object
```
jelmer, line 1762: For consistency, just two spaces.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+        file_path: file to reset, relative to the repository path
+        target: branch or commit or b'HEAD' to reset
+    """
+    if target in branch_list(repo):
```
jelmer, line 1766: you probably want just parse_tree() here rather than duplicating the logic (see elsewhere in dulwich/porcelain.py)

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+        sha = repo.refs[b'refs/heads/' + target]
+    elif target == b'HEAD':
+        sha = repo.head()
+    else:
+        sha = target
+    tree = parse_tree(repo, treeish=sha)
+
+    def get_entry(repo, tree, file_path: str):
```
jelmer, line 1774: get_entry could be replaced with tree.lookup_path(repo.object_store.__getitem__, file_path)

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+    file_entry = get_entry(repo, tree, file_path)
+    
+    full_path = os.path.join(repo.path, file_path)
+    blob = repo.object_store[file_entry[1]]
+    mode = file_entry[0]
+    build_file_from_blob(blob, mode, full_path.encode())
+
+
+def checkout(repo, branch: bytes, force: bool = False):
+    """
```
jelmer, line 1796: See other comment about docstrings.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+    # unstage all files in the index
+    tracked_changes = []
+    for file in get_tree_changes(repo).values():
+        tracked_changes += file
+    repo.unstage(tracked_changes)
+
+    # reset all unstaged file to target
+    normalizer = repo.get_blob_normalizer()
+    filter_callback = normalizer.checkin_normalize
+    unstaged_files = list(get_unstaged_changes(repo.open_index(), repo.path, filter_callback))
```
jelmer, line 1818: why list(), couldn't you just iterate through the entries of get_unstaged_changes?

dulwich/repo.py
```diff
@@ -1305,6 +1305,59 @@ def stage(self, fs_paths):
                     index[tree_path] = index_entry_from_stat(st, blob.id, 0)
         index.write()
 
+    def unstage(self, relpaths: List[bytes] = []):
+        """
+        unstage specific file in the index
```
jelmer, line 1310: See earlier comment about docstrings.

dulwich/tests/test_porcelain.py
```diff
@@ -1317,6 +1320,222 @@ def test_hard_commit(self):
         self.assertEqual([], changes)
 
 
+class UnstageTests(PorcelainTestCase):
```
jelmer, line 1323: Since unstage lives on Repo, please add these tests to dulwich/tests/test_repo.py

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+    """
+    switch branches or restore working tree files
+    Args:
+        repo: dulwich Repo object
+        branch: branch name to checkout
+    """
+
+    # check repo status
+    if not force:
+        if status(repo) != GitStatus(staged={'add': [], 'delete': [], 'modify': []}, unstaged=[], untracked=[]):
```
jelmer, line 1805: What about ignored files? They could be versioned in the target.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,76 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+    for file in get_tree_changes(repo).values():
+        tracked_changes += file
+    repo.unstage(tracked_changes)
+
+    # reset all unstaged file to target
+    normalizer = repo.get_blob_normalizer()
+    filter_callback = normalizer.checkin_normalize
+    unstaged_files = list(get_unstaged_changes(repo.open_index(), repo.path, filter_callback))
+    for file in unstaged_files:
+        reset_file(repo, file.decode(), b'HEAD')
```
jelmer, line 1820: wouldn't this need to update the index as well?

dulwich/porcelain.py
DedSecer: It seem that ignored files will not appear in the GitSatus, and `git checkout` will not modify the ignored files

jelmer: The branch you're changing to can contain files that were ignored (and not versioned) in the original branch you were on.

DedSecer: Does your meaning like this?
```
$ git init
$ # add new thing to file `foo`
$ git add foo; git commit -m 'empty'
$ git branch bran
$ # add new thing to file `bar`
$ # add file `bar` to .gitignore
$ git add .gitignore; git commit -m 'empty1'
```
Here's the output of `git status` in branch `master`:
```
On branch master
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        bar

nothing added to commit but untracked files present (use "git add" to track)
```

And the output of `git status` in branch `bran`:
```
On branch bran
nothing to commit, working tree clean
```

I think both `git checkout` and  `porcelain.checkout` will not do any thing to the file `bar` after the first commit

dulwich/errors.py
DedSecer: raising DirNotCleanError in the new commit

dulwich/porcelain.py
DedSecer: I have replace `get_entry` with `tree.lookup_path(repo.object_store.__getitem__, file_path)`, 
but the test rasing error in the `dulwich.tests.test_porcelain.ResetFileTests`:
```
> python -m unittest dulwich.tests.test_porcelain.ResetFileTests
...E
ERROR: test_resetfile_with_dir (dulwich.tests.test_porcelain.ResetFileTests)
  File "D:\Repo\dulwich\dulwich\tests\test_porcelain.py", line 1497, in test_resetfile_with_dir
  File "D:\Repo\dulwich\dulwich\porcelain.py", line 1781, in reset_file
    file_entry = tree.lookup_path(repo.object_store.__getitem__, file_path.encode())
  File "D:\Repo\dulwich\dulwich\objects.py", line 1191, in lookup_path
    mode, sha = obj[p]
  File "D:\Repo\dulwich\dulwich\objects.py", line 1058, in __getitem__
    return self._entries[name]
KeyError: b'new_dir\\foo'
```
I think we should change `parts = path.split(b"/")` with `parts = os.path.split(path)` in the [objects.py, line 1182](https://github.com/dulwich/dulwich/blob/80bffaca1dfe59a5ba57a045f3ce7122227a7647/dulwich/objects.py#L1182)

dulwich/porcelain.py
DedSecer: I have remove other Conditionals except
```
if target in branch_list(repo):
        target = b'refs/heads/' + target
```
I think we should keep this to match branch name

jelmer: If we'd like to support branch names here (and I agree that would be a good idea), then we should fix that in parse_tree rather than here.

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,61 @@ def update_head(repo, target, detached=False, new_branch=None):
             r.refs.set_symbolic_ref(b"HEAD", to_set)
 
 
+def reset_file(repo, file_path: str , target: bytes = b'HEAD'):
```
jelmer, line 1758: extra space after "str"

jelmer:
Thanks for your patience; I think we're close to being able to merge this now.

dulwich/porcelain.py
DedSecer: If I remove extra space after "str", the flake8 will output: `missing whitespace after ':'flake8(E231)`

dulwich/porcelain.py
DedSecer: Update parser_tree in the new commit

dulwich/porcelain.py
DedSecer: reset_file will update index now

DedSecer: Where we should update index?After reset_file or at the end of checkout?

DedSecer:
@jelmer you can have a review now.

dulwich/porcelain.py
```diff
@@ -1753,6 +1756,91 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+        
+        index = repo.open_index()
+        normalizer = repo.get_blob_normalizer()
+        filter_callback = normalizer.checkin_normalize
+        unstaged_changes = list(get_unstaged_changes(index, repo.path, filter_callback))
+        for file in unstaged_changes:
+            if file in repo.open_index():
+                raise DirNotCleanError('trying to checkout when working directory not clean')
+ 
+    current_file_set = set(parse_tree(repo, repo.head()))
```
jelmer, line 1813: this will only return files in the root; how would this work for files under directories?

dulwich/porcelain.py
jelmer: After the checkout, otherwise this will be very slow for larger repositories.

dulwich/objectspec.py
```diff
@@ -54,6 +54,8 @@ def parse_tree(repo, treeish):
       KeyError: If the object can not be found
     """
     treeish = to_bytes(treeish)
+    if treeish in repo.refs.keys(base=b"refs/heads/"):
```
jelmer, line 57: perhaps just call parse_ref() here, so it also e.g. handles tags?

dulwich/porcelain.py
DedSecer: Have  I update index in the stage()?Or I should update index on more time?

dulwich/porcelain.py
jelmer: presumably if you delete the space before rather than after "str"?

dulwich/porcelain.py
jelmer: So ideally we'd just update the index once for each checkout operation, since it can be expensive for larger indexes. I'm okay with compromising for now and e.g. doing one write for staging and one for unstaging, but one write for each will be noticeable.

DedSecer: sorry, I can’t understand you.Since we have update index in the Repo.unstage, should we update index in the end of checkout?

dulwich/porcelain.py
DedSecer: using tree.lookup_path to check if the file in the tree now.

dulwich/objectspec.py
DedSecer: using parse_ref() now

dulwich/porcelain.py
jelmer: Yeah, we should update the index at the end of checkout - but we can't be updating the index for every reset_file() call, since that simply won't scale.

DedSecer: Repo.unstage will update the different file's index between tatget branch and current branch, should we update the index at the end of checkout.

jelmer: Yes -  we shouldn't be modifying the index every time reset_file() gets called. The changes *after* unstage still need to be properly written to the index.

DedSecer: So I think we just update index after unstage and needn't update index at the end of checkout.

jelmer:
Sorry, I've been travelling. I'll take a look at this tomorrow or on Tuesday.

dulwich/objects.py
```diff
@@ -1179,7 +1179,12 @@ def lookup_path(self, lookup_obj, path):
           path: Path to lookup
         Returns: A tuple of (mode, SHA) of the resulting path.
         """
-        parts = path.split(b"/")
```
jelmer, line 1182: Why is this necessary? At the very least, there should be accompanying test changes - but I don't understand what bug in lookup_path this is fixing.

dulwich/objectspec.py
```diff
@@ -54,6 +54,8 @@ def parse_tree(repo, treeish):
       KeyError: If the object can not be found
     """
     treeish = to_bytes(treeish)
+    if treeish not in repo.object_store.__iter__():
```
jelmer, line 57: Please don't use __iter__ but __contains__ - __iter__ can be very slow on large object stores. Why is the code on line 59 not sufficient? Tha twould have the same ffect.

dulwich/repo.py
```diff
@@ -1305,6 +1305,58 @@ def stage(self, fs_paths):
...
+
+        from dulwich.index import IndexEntry
+        from dulwich.porcelain import remove
+
+        index = self.open_index()
+        try:
+            tree_id = self[b'HEAD'].tree
+        # no head mean no commit in the repo
+        except KeyError:
+            remove(self, paths=[os.path.join(self.path, path.decode()) for path in relpaths], cached=True)
```
jelmer, line 1323: porcelain is low level code, please don't add circular imports from it into dulwich.repo. It should be the other way around - dulwich.porcelain.remove() calls Repo.unstage.

dulwich/objects.py
DedSecer: this [comment](https://github.com/dulwich/dulwich/pull/898#discussion_r692582168)

dulwich/objectspec.py
DedSecer: I would like to match treeish in the `repo.refs` while treeish is not a **commit sha** . It seem it isn't a good idea. Do you have better solution? what about replace `repo.object_store.__iter__()` with `list(repo.object_store)`?

dulwich/repo.py
DedSecer: updated in the new commit

dulwich/porcelain.py
```diff
@@ -1753,6 +1755,89 @@ def update_head(repo, target, detached=False, new_branch=None):
...
+        for file in get_tree_changes(repo)['modify']:
+            if file in repo.open_index():
+                raise DirNotCleanError('trying to checkout when working directory not clean')
+        
+        index = repo.open_index()
+        normalizer = repo.get_blob_normalizer()
+        filter_callback = normalizer.checkin_normalize
+        unstaged_changes = list(get_unstaged_changes(index, repo.path, filter_callback))
+        for file in unstaged_changes:
+            if file in repo.open_index():
```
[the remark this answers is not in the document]
DedSecer: updated in the new commit

dulwich/objects.py
jelmer: You're passing a file system path to a function that expects a canonical git path - the caller there should be converting from a filesystem path to a git path and then calling lookup_path().

dulwich/objectspec.py
jelmer: You should be able to rely on \_\_contains\_\_, or better yet - rely on just accessing the treeish and handling the exception if that doesn't work.

\_\_iter\_\_ and list() both traverse all the objects in the object store, which can be millions of objects for larger projects like the linux kernel.

dulwich/porcelain.py
jelmer: You'd have to pass index to reset_file() in that case though, and you may want to make it private - or provide a _reset_file() that works on an index and is used by both checkout() and a public reset_file() wrapper that takes a repo argument.

jelmer (commented):
Another round of review comments. I'll see if I can cherry-pick at least the unstage changes.

dulwich/objects.py
jelmer: E.g. by calling dulwich.index._fs_to_tree_path,

dulwich/objectspec.py
DedSecer: Thanks, updated in the new commit

dulwich/objects.py
DedSecer: call dulwich.index._fs_to_tree_path now

dulwich/porcelain.py
DedSecer: why should I pass index to reset_file()? I needn't index in reseet_file().

jelmer:
I've merged some of the components.

DedSecer:
@jelmer Hey,are you still merging my pull requests into dulwich?

jelmer:
Sorry for the delay - I'll take another look this week.

jelmer:
Closing this now that #1148 has landed.

PR #901, opened 2021-09-14
Fixes `Tag.signature` field

sobolevn:
I am opening this PR, because new version of `mypy` identified a bug in this definition.
`self.signature` is not in `__slots__`. I looked like a typo.
Unit tests in `tests/` seem to ignore this method.

Please, feel free to close if this is not a bug and there's some extra magic I've missed.
Related https://github.com/python/mypy/pull/10864


sobolevn:
Ok, I see that `signature` is a `serializable_property` and this is a false positive. https://github.com/dulwich/dulwich/blob/ede437fa736597a57c3fab7ef69fc281184a80f8/dulwich/objects.py#L868

Sorry for the noise!

PR #902, opened 2021-09-15
client: add support for GIT_SSH_COMMAND

pmrowla:
- Adds `ssh_command` parameter to `SSHGitClient` and `SSHVendor.run_command`
- If specified, `ssh_command` will be used instead of the default `ssh` or `plink[.exe]` executables when we run SSH via subprocess (in both `SubprocessSSHVendor` and `PLinkSSHVendor`)
    - Existing options for host/port/user/password/etc will still be appended to the command and should not be included in `ssh_command`
    - `ssh_command` is ignored for the paramiko contrib vendor
- `SSHGitClient` will use `GIT_SSH` or `GIT_SSH_COMMAND` environment variables if they are set (with precedence given to `ssh_command` > `GIT_SSH_COMMAND`, `GIT_SSH`)
    - We do not autodetect the vendor variant when `GIT_SSH` or `GIT_SSH_COMMAND` is set, currently we assume that the user has selected the appropriate dulwich SSH vendor and that it matches `GIT_SSH`/`GIT_SSH_COMMAND`

As far as I can tell this should be consistent with the cgit docs for [GIT_SSH_COMMAND](https://git-scm.com/docs/git#Documentation/git.txt-codeGITSSHcode) and [ssh.variant](https://git-scm.com/docs/git-config#Documentation/git-config.txt-sshvariant)

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1467,6 +1467,7 @@ class SSHVendor(object):
         port=None,
         password=None,
         key_filename=None,
+        ssh_command=None,
     ):
         """Connect to an SSH server.
 
@@ -1480,6 +1481,7 @@ class SSHVendor(object):
           port: Optional SSH port to use
           password: Optional ssh password for login or private key
           key_filename: Optional path to private keyfile
+          ssh_command: Optional SSH command
 
         Returns:
 
@@ -1505,6 +1507,7 @@ class SubprocessSSHVendor(SSHVendor):
         port=None,
         password=None,
         key_filename=None,
+        ssh_command=None,
     ):
 
         if password is not None:
@@ -1512,7 +1515,7 @@ class SubprocessSSHVendor(SSHVendor):
                 "Setting password not supported by SubprocessSSHVendor."
             )
 
-        args = ["ssh", "-x"]
+        args = [ssh_command or "ssh", "-x"]
 
         if port:
             args.extend(["-p", str(port)])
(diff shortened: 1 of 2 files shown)


pmrowla:
> - `SSHGitClient` will use `GIT_SSH` or `GIT_SSH_COMMAND` environment variables if they are set (with precedence given to `ssh_command` > `GIT_SSH_COMMAND`, `GIT_SSH`)
    - We do not autodetect the vendor variant when `GIT_SSH` or `GIT_SSH_COMMAND` is set, currently we assume that the user has selected the appropriate dulwich SSH vendor and that it matches `GIT_SSH`/`GIT_SSH_COMMAND`

Not sure if this is the desired behavior for dulwich, we could either ignore the env vars entirely (and rely on the user just setting `ssh_command`) or do the same autodetection that cgit does (check the basename of the first arg in `GIT_SSH_COMMAND` to see if its ssh or plink)

dulwich/client.py
```diff
@@ -1505,14 +1507,18 @@ def run_command(
...
     ):
 
         if password is not None:
             raise NotImplementedError(
                 "Setting password not supported by SubprocessSSHVendor."
             )
 
-        args = ["ssh", "-x"]
+        if ssh_command:
+            args = ssh_command.split() + ["-x"]
```
jelmer, line 1519: shlex.split() rather than plain split?

jelmer (approved):

PR #918, opened 2021-12-20
fix handle of gecos in case of null

hardyvince:
e.g. termux doesn't fill the user's full name which than lead gives back null to pwd's full name request what wasn't handled by repo.py

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -146,7 +146,8 @@ def _get_default_identity() -> Tuple[str, str]:
         except KeyError:
             fullname = None
         else:
-            fullname = gecos.split(",")[0]
+            if not gecos:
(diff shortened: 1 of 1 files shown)


dulwich/repo.py
```diff
@@ -146,7 +146,8 @@ def _get_default_identity() -> Tuple[str, str]:
         except KeyError:
             fullname = None
         else:
-            fullname = gecos.split(",")[0]
+            if not gecos:
```
jelmer, line 149: Is this tested? It seems like this is the wrong way around? Also, you'd probably want to assign fullname even if gecos is empty.

jelmer (changes requested):

jelmer:
Closed in favor of a fix in master.

PR #927, opened 2022-01-24
Add support for parsing author lines with no angle brackets

progval:
This commit adds support for parsing author lines with no angle brackets.
It uses a regexp to avoid hand-writing some backtracking logic.

This also adds tests for other types of brokenness that I found in existing
git repos, and that were already supported. I tried to use the original example
as test data when possible.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -27,6 +27,7 @@ from collections import namedtuple
 import os
 import posixpath
 import stat
+import re
 from typing import (
     Optional,
     Dict,
@@ -69,6 +70,11 @@ _TAGGER_HEADER = b"tagger"
 
 
 S_IFGITLINK = 0o160000
+
+# Intentionally flexible to support various types of brokenness
+_TIME_ENTRY_RE = re.compile(
+    b"^(?P<person>.*) (?P<time>-?[0-9]+) (?P<timezone>[+-]{0,2}[0-9]+)$"
+)
 
 
 MAX_TIME = 9223372036854775807  # (2**63) - 1 - signed long int max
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the PR. It looks like some of these will break roundtripping, i.e. parsing some bytes into a Commit object  and then getting the exact same set of bytes back out when calling Commit.as_raw_bytes() - which is a property I'd like to maintain.

Have you actually seen all of these in the wild? We could probably offer some sort of alternative for cases like this, if we don't want to add attributes for each one.

progval:
> It looks like some of these will break roundtripping

> which is a property I'd like to maintain

There were already some breakages on commits with weird timezones. For example, missing leading zeros in timezones (I don't know how this happened, I assume home-made Git implementations) are fixed by Dulwich:

```py
>>> b = b'tree aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa \nparent bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\nauthor John Doe <jdoe@example.com> 1614159930 +100\ncommitter John Doe <jdoe@example.com> 1614159930 +100\n\nfoo'
>>> c = dulwich.objects.Commit.from_string(b)
>>> c.author = c.author  # reset t._needs_serialization
>>> c.as_raw_string()
b'tree aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa \nparent bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb\nauthor John Doe <jdoe@example.com> 1614159930 +0100\ncommitter John Doe <jdoe@example.com> 1614159930 +0100\n\nfoo'
```

The solution we decided to use at @SoftwareHeritage was to keep the original timezone bytes somewhere. I could do it for Dulwich (I actually have a commit ready to do this, based on the one in this PR)

> Have you actually seen all of these in the wild?

Yes. I have seen about 14k commits with missing leading zeros as above. 232 with the timezone `+0575` (which gets rewritten as `+0615`), and 726 with various types of brokenness (the `0000` above, huge timezones with a leading zero, such as `+091800` which gets rewritten as `+91800`, etc.)

> We could probably offer some sort of alternative for cases like this, if we don't want to add attributes for each one.

Dulwich could store the raw bytes of the original timezone. This requires a new attribute, but it allows the `*_timezone_neg_utc` attribute to be removed.

> We could probably offer some sort of alternative for cases like this, if we don't want to add attributes for each one.

jelmer:
FWIW Rather than changing the default parser, I'm planning to lazily load these event lines (user identity + timestamp + timezone). That way, you can get access to the raw data and parse that with a less strict parser if you like. The added minor benefit could be that we get to avoid parsing them when we don't need them.

My PoC is in https://github.com/jelmer/dulwich/pull/1135; that approach works, but adds two extra object instances per commit which is not ideal.

PR #929, opened 2022-01-26
clone: re-add progress support

pmrowla:
Re-enable progress support in clone methods for reporting fetch progress

- Adds `progress=` in `Repo.clone`, `client.clone`
- Uses `progress=errstream.write` in `porcelain.clone`


jelmer (approved):

pmrowla:
@jelmer not sure what happened after the PR got merged but it looks like this commit is no longer available in master?

jelmer:
race condition - fixed now. I've also turned on branch protection, so this shouldn't happen again.

Issue #932, opened 2022-02-01
Add support for no_proxy environment variable 


epopcop:
I'm using dulwich in a special environment where the proxy settings are out of my control.
One of the features there supports "whitelisting" a set of addresses from the proxy, and these are written to the "no_proxy" environment variable, which is not standardized, but widely used, for example by [urllib](https://github.com/python/cpython/blob/030a713183084594659aefd77b76fe30178e23c8/Lib/urllib/request.py#L2519-L2551) or [requests](https://github.com/psf/requests/blob/99b3b492418d0751ca960178d274f89805095e4c/requests/utils.py#L833-L858), but not by urllib3.

Will it be possible to add this check? 
I will try doing so myself and will submit a PR if succeeded. 
The main issue I see now is that `default_urllib3_manager()` doesn't get the repo url or domain. Can I get it from the `config` parameter?

Thank you!

jelmer:
Is this something git itself supports somehow?

jelmer:
I'd be open to supporting git config variables that aren't currently supported.

Alternatively, you could construct a custom implementation of urllib3 manager that has the behaviour you need?

epopcop:
I'm not that good at c - but I believe git takes the variable into consideration [here](https://github.com/git/git/blob/69a9c10c95e28df457e33b3c7400b16caf2e2962/http.c#L970)
Looks like it's using curl to make the calls - and it's passing the no_proxy variable to curl behind the scenes.

I actually have a couple of ways to work around this because I barely use porcelain, so I think of overriding `HttpGitClient` in [client.py](https://github.com/dulwich/dulwich/blob/72377321d3107ab3b51121790d86968caa0efa7a/dulwich/client.py#L2214) with a custom client that checks the environment variables and passes the right pool manager. This is similar to how the other tools do it. There are ups and downs in this, but I do want to see that in dulwich eventually (if possible of course).

Also, I actually have an example of a similar workaround I did for adding support for paramiko private keys (Supported in `ParamikoSSHVendor`, not in `SSHGitClient`, the parameter pkey is not passed to the vendor)

```
from dulwich import client as _mod_client

class CustomParamikoSSHVendor:
    def run_command(
            self,
            host,
            command,
            username=None,
            port=None,
            password=None,
            key_filename=None,
            **kwargs
    ):

        ...
        # Using the password parameter to pass the key to pkey
        if password:
            connection_kwargs["pkey"] = password
        ...

# dulwich patch to add pkey support
[...]

PR #933, opened 2022-02-02
Add requests_vendor as higher level http client alternative

epopcop:
Workaround for using no_proxy environment variable as discussed in #932 by using requests as a higher-level alternative to urllib3. It's based on dulwich/cotrib/paramiko_vendor.py

### DIFF
--- a/dulwich/contrib/requests_vendor.py
+++ b/dulwich/contrib/requests_vendor.py
@@ -1,0 +1,142 @@
+# requests_vendor.py -- requests implementation of the AbstractHttpGitClient interface
+# Copyright (C) 2022 Eden Shalit <epopcop@gmail.com>
+#
+# Dulwich is dual-licensed under the Apache License, Version 2.0 and the GNU
+# General Public License as public by the Free Software Foundation; version 2.0
+# or (at your option) any later version. You can redistribute it and/or
+# modify it under the terms of either of these two licenses.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# You should have received a copy of the licenses; if not, see
+# <http://www.gnu.org/licenses/> for a copy of the GNU General Public License
+# and <http://www.apache.org/licenses/LICENSE-2.0> for a copy of the Apache
+# License, Version 2.0.
+
+
+"""Requests HTTP client support for Dulwich.
+
+To use this implementation as the HTTP implementation in Dulwich, override
+the dulwich.client.HttpGitClient attribute:
+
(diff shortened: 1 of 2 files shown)


dulwich/contrib/requests_vendor.py
```diff
@@ -0,0 +1,142 @@
...
+
+        if data:
+            resp = self.session.post(url, headers=req_headers, data=data)
+        else:
+            resp = self.session.get(url, headers=req_headers)
+
+        # Add required fields as stated in AbstractHttpGitClient._http_request
+        # Since requests follows redirects by default - redirect_location is always empty
+        resp.content_type = resp.headers.get("Content-Type")
+        resp.redirect_location = ""
```
jelmer, line 79: This is going to introduce extra requests since subsequent requests are going to have to follow redirects again.

dulwich/contrib/requests_vendor.py
```diff
@@ -0,0 +1,142 @@
...
+    session.headers.update({"User-agent": user_agent})
+
+    # Verify is True by default
+    if ca_certs:
+        session.verify = ca_certs
+    elif not ssl_verify:
+        session.verify = ssl_verify
+
+    if proxy_server:
+        # Assuming git config is referring to both http and https
```
jelmer, line 137: there's a separate setting for HTTPS I believe, though I don't think the standard Dulwich client gets this right either.

jelmer (approved):

dulwich/contrib/requests_vendor.py
jelmer: Happy to merge this with that in place, but perhaps add a note to the comment at the top?

jelmer:
The tests are failing because of the type checks; perhaps either add types-requests to the GitHub action or exclude contrib from type checks altogether.

dulwich/contrib/requests_vendor.py
epopcop: Would you recommend disabling redirects and resp.redirect_location?

jelmer:
On Fri, Feb 04, 2022 at 02:24:44PM -0800, epopcop wrote:
> @epopcop commented on this pull request.
> 
> 
> 
> > +            req_headers.update(headers)
> +
> +        if allow_compression:
> +            req_headers["Accept-Encoding"] = "gzip"
> +        else:
> +            req_headers["Accept-Encoding"] = "identity"
> +
> +        if data:
> +            resp = self.session.post(url, headers=req_headers, data=data)
> +        else:
> +            resp = self.session.get(url, headers=req_headers)
> +
> +        # Add required fields as stated in AbstractHttpGitClient._http_request
> +        # Since requests follows redirects by default - redirect_location is always empty
> +        resp.content_type = resp.headers.get("Content-Type")
> +        resp.redirect_location = ""
> 
> Would you recommend disabling redirects and resp.redirect_location?
No, you probably want to follow redirects but then also set
resp.redirect_location to where you ended up after the redirects.

Jelmer

jelmer (approved):

PR #951, opened 2022-03-11
Fix dulwich.porcelain.diff_tree default outstream value

### LINKED ISSUE #950: Import Error on notebook environment
Hello, when trying to import latest dulwich version (0.22.33), the import fails with an AttributeError:

```
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_756805/2729115134.py in <module>
----> 1 import dulwich.porcelain

~/.virtualenv/lib/python3.7/site-packages/dulwich/porcelain.py in <module>
    832 
    833 
--> 834 def diff_tree(repo, old_tree, new_tree, outstream=sys.stdout.buffer):
    835     """Compares the content and mode of blobs found via two tree objects.
    836 

AttributeError: 'OutStream' object has no attribute 'buffer'
```

It looks like `sys.stdout` has no 'buffer' attribute.

Would changing the outstream default argument in here https://github.com/jelmer/dulwich/blame/master/dulwich/porcelain.py#L835 to default_bytes_out_stream (https://github.com/jelmer/dulwich/blob/master/dulwich/porcelain.py#L173) would be the right approach to fix it? I would be happy to send a PR if that is the case.

Cheers and thanks again for your good work! :raised_hands:

Lothiraldan:
sys.stdout.buffer is failing on Jupyter notebook environments. Fix #950

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -832,7 +832,7 @@ def show(
             show_object(r, o, decode, outstream)
 
 
-def diff_tree(repo, old_tree, new_tree, outstream=sys.stdout.buffer):
+def diff_tree(repo, old_tree, new_tree, outstream=default_bytes_out_stream):
     """Compares the content and mode of blobs found via two tree objects.
 
     Args:


jelmer:
Hmm, typing errors on windows - but unrelated.

PR #953, opened 2022-04-02
WIP: Git attributes parser

andrewshadura:
Hi,

I’ve just found this uncommitted in my local Dulwich checkout. I wrote this for another project initially, as Dulwich only had `# TODO Parse the git attributes files`, so I decided to contribute it back, but apparently completely forgot about it. I’m not sure how I should proceed integrating it, since a lot of time has passed and there’s a lot of other stuff going on, so I’m not sure how much time I can dedicate to it at the moment, but I just wanted this to be around in case it’s useful.


jelmer:
Thanks! I appreciate you sharing this code even if it isn't finished. I've added it to my list of things to follow up on; might take a couple of weeks.

jelmer:
Rebased and further improved, now merged into master.

PR #956, opened 2022-04-29
add untracked_files kwarg to porcelain.status

dtrifiro:
porcelain.status: add untracked_files kwarg

Status calls can be quite slow when large untracked directories are present because they are being walked.
Using `untracked_files="normal"` returns untracked files and directories (behaving like `git status -unormal`,
which is git's default), using `untracked_files="all"` (dulwich default) also returns untracked files in
untracked directories, `untracked_files="no"` returns no untracked files (like `git status -uno`).

Note that old behaviour is preserved by having `untracked_files` default to "all".

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -1155,12 +1155,20 @@ def pull(
             _import_remote_refs(r.refs, remote_name, fetch_result.refs)
 
 
-def status(repo=".", ignored=False):
+def status(repo=".", ignored=False, untracked_files="all"):
     """Returns staged, unstaged, and untracked changes relative to the HEAD.
 
     Args:
       repo: Path to repository or repository object
       ignored: Whether to include ignored files in untracked
+      untracked_files: How to handle untracked files:
+            - "no": Show no untracked directories
+                  (like `git status -uno`).
+            - "all": show all files in untracked directories
+                  (like `git status -uall`, dulwich default).
+            - "normal": shows untracked files and directories.
+                  (like `git status -unormal`, git default).
+
     Returns: GitStatus tuple,
         staged -  dict with lists of staged paths (diff index/HEAD)
         unstaged -  list of unstaged paths (diff index/working-tree)
@@ -1175,10 +1183,19 @@ def status(repo=".", ignored=False):
         filter_callback = normalizer.checkin_normalize
         unstaged_changes = list(get_unstaged_changes(index, r.path, filter_callback))
 
-        untracked_paths = get_untracked_paths(
-            r.path, r.path, index, exclude_ignored=not ignored
-        )
-        untracked_changes = list(untracked_paths)
+        if untracked_files == "no":
+            untracked_changes = []
+        else:
+            if untracked_files == "normal":
+                walk = False
+            elif untracked_files == "all":
+                walk = True
+            else:
+                raise ValueError("untracked_files must be one of no, normal or all.")
+            untracked_paths = get_untracked_paths(
+                r.path, r.path, index, exclude_ignored=not ignored, walk=walk
+            )
+            untracked_changes = list(untracked_paths)
 
         return GitStatus(tracked_changes, unstaged_changes, untracked_changes)
 
@@ -1215,7 +1232,13 @@ def _walk_working_dir_paths(frompath, basepath, prune_dirnames=None):
             dirnames[:] = prune_dirnames(dirpath, dirnames)
 
 
-def get_untracked_paths(frompath, basepath, index, exclude_ignored=False):
+def get_untracked_paths(
+    frompath,
+    basepath,
+    index,
+    exclude_ignored=False,
+    walk=True,
+):
     """Get untracked paths.
 
     Args:
@@ -1223,6 +1246,7 @@ def get_untracked_paths(frompath, basepath, index, exclude_ignored=False):
       basepath: Path to compare to
       index: Index to check against
       exclude_ignored: Whether to exclude ignored paths
+      walk: Whether to walk untracked paths
 
     Note: ignored directories will never be walked for performance reasons.
       If exclude_ignored is False, only the path to an ignored directory will
@@ -1232,6 +1256,7 @@ def get_untracked_paths(frompath, basepath, index, exclude_ignored=False):
         ignore_manager = IgnoreFilterManager.from_repo(r)
 
     ignored_dirs = []
+    untracked_dirs = []
 
     def prune_dirnames(dirpath, dirnames):
         for i in range(len(dirnames) - 1, -1, -1):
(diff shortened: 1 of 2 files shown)


dulwich/tests/test_porcelain.py
```diff
@@ -2167,7 +2178,32 @@ def test_get_untracked_paths_subdir(self):
                     self.repo.open_index(),
                     exclude_ignored=True,
                 )
-            )
+            ),
+        )
+
+    def test_get_untracked_paths_walk(self):
+        os.mkdir(os.path.join(self.repo.path, "dir"))
+        os.mkdir(os.path.join(self.repo.path, "dir/subdir"))
```
efiop, line 2186:
```suggestion
        os.mkdir(os.path.join(self.repo.path, "dir", "subdir"))
```

efiop: Seems like there is inconsitency in this test. Sometimes use use separate parts and sometimes you hardcode `/`.

dulwich/porcelain.py
```diff
@@ -1155,12 +1155,16 @@ def determine_wants(remote_refs, **kwargs):
             _import_remote_refs(r.refs, remote_name, fetch_result.refs)
 
 
-def status(repo=".", ignored=False):
+def status(repo=".", ignored=False, walk_untracked=True):
```
efiop, line 1158: Should it be `untracked_files` to match the cli flag?

dulwich/tests/test_porcelain.py
```diff
@@ -1842,13 +1842,18 @@ def test_status_all(self):
         mod_path = os.path.join(self.repo.path, "bar")
         add_path = os.path.join(self.repo.path, "baz")
         us_path = os.path.join(self.repo.path, "blye")
-        ut_path = os.path.join(self.repo.path, "blyat")
+        ut_path = os.path.join(self.repo.path, "untracked_file")
+        os.mkdir(os.path.join(self.repo.path, "untracked_dir"))
+        ut_subfile_path = os.path.join(self.repo.path, "untracked_dir/file")
```
efiop, line 1847:
```suggestion
        ut_subfile_path = os.path.join(self.repo.path, "untracked_dir", "file")
```
hence the failing windows test

dtrifiro: Fixed everywhere

dulwich/porcelain.py
```diff
@@ -1215,14 +1232,21 @@ def _walk_working_dir_paths(frompath, basepath, prune_dirnames=None):
...
+    walk=True,
+):
     """Get untracked paths.
 
     Args:
       frompath: Path to walk
       basepath: Path to compare to
       index: Index to check against
       exclude_ignored: Whether to exclude ignored paths
+      walk: Whether to walk untracked paths.
```
efiop, line 1249:
```suggestion
      walk: Whether to walk untracked paths
```

dulwich/porcelain.py
```diff
@@ -1251,15 +1285,13 @@ def prune_dirnames(dirpath, dirnames):
         if not is_dir:
             ip = path_to_tree_path(basepath, ap)
             if ip not in index:
-                if (
-                    not exclude_ignored
-                    or not ignore_manager.is_ignored(
-                        os.path.relpath(ap, basepath)
-                    )
+                if not exclude_ignored or not ignore_manager.is_ignored(
+                    os.path.relpath(ap, basepath)
```
efiop, line 1289: Unrelated change?

dulwich/porcelain.py
```diff
@@ -1243,6 +1268,15 @@ def prune_dirnames(dirpath, dirnames):
                         os.path.join(os.path.relpath(path, frompath), "")
                     )
                 del dirnames[i]
+            elif not walk:
```
efiop, line 1271: `walk` seems like a very generic term that doesn't really describe what is going on. Not that `normal` vs `all` are very descriptive themselves, but maybe let's just pass the policy down here and decide what to do in one place instead of introducing another translation layer with `walk`?

dulwich/tests/test_porcelain.py
```diff
@@ -2167,7 +2184,32 @@ def test_get_untracked_paths_subdir(self):
...
+    def test_get_untracked_paths_walk(self):
+        os.mkdir(os.path.join(self.repo.path, "dir"))
+        os.mkdir(os.path.join(self.repo.path, "dir", "subdir"))
+        with open(os.path.join(self.repo.path, "dir", "file"), "w") as f:
+            f.write("foo")
+        with open(os.path.join(self.repo.path, "dir", "subdir", "subfile"), "w") as f:
+            f.write("foo")
+
+        self.assertEqual(
+            {"dir/"},
```
pmrowla, line 2199: need to account for windows paths here with `os.path.join("dir", "")`

dulwich/porcelain.py
```diff
@@ -1155,12 +1155,20 @@ def determine_wants(remote_refs, **kwargs):
...
       repo: Path to repository or repository object
       ignored: Whether to include ignored files in untracked
+      untracked_files: How to handle untracked files:
+            - "no": Show no untracked directories
+                  (like `git status -uno`).
+            - "all": show all files in untracked directories
+                  (like `git status -uall`, dulwich default).
+            - "normal": shows untracked files and directories.
+                  (like `git status -unormal`, git default).
+
```
pmrowla, line 1171: More of a question for @jelmer, but I'm not sure whether we actually need to support `all` vs `normal` here. It could just be `untracked_files=True` instead, where true=all and false=no. 

(The same way that existing `ignored` option just supports true/false rather than the full set of git's `traditional/no/matching` modes for ignored files)

jelmer: Ideally we'd match the git command-line here. 

For convenience, we could support both strings and booleans as value for untracked_files?

jelmer:
Any news on this; anything I can do to help unblock?

dtrifiro:
Hi Jelmer,
I ended up noticing that this patch would yield slightly different results compared to `git status`, and I ended up deciding dropping it to avoid having mismatching behaviour between dulwich and git.

Just for the record, for a dummy repo:
```console
$ tree
├── tracked_with_ignored
│   ├── tracked_file
│   └── untracked
│       └── ignored
├── untracked_empty
│   ├── ignored
│   └── untracked_empty
│       └── untracked_empty
├── untracked_empty_dir
└── untracked_nonempty_dir
    └── file

7 directories, 4 files

$ git status --porcelain                                                                                                                                                                                  
?? .gitignore
?? untracked_nonempty_dir/
$ python -c 'from dulwich.porcelain import status; _, _, u = status(walk_untracked=False); [print(el) for el in u]'                                                                                           
untracked_empty/
untracked_nonempty_dir/
untracked_empty_dir/
tracked_with_ignored/untracked/
```
It seemed too much of an effort to replicate's git's [exact behaviour](https://github.com/git/git/blob/277cf0bc36094f6dc4297d8c9cef79df045b735d/dir.c#L1851) with dubious advantages.

Closing this and opening https://github.com/jelmer/dulwich/pull/964.

PR #959, opened 2022-05-15
gha: publish: don't forget to install paramiko

efiop:
Was missed in https://github.com/jelmer/dulwich/commit/94a68a899b6139c35c8a10e683218ee2ba755105, which resulted in failed wheel builds https://github.com/jelmer/dulwich/runs/6442441328?check_suite_focus=true.


efiop:
Could we please trigger a new version release after this? Related to #7743

Also, looks like flake8 has been broken for a bit 🙁

jelmer:
Thanks!

PR #960, opened 2022-05-15
client: remove unused write_pack_data

efiop:
Was caught by flake8.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -105,7 +105,6 @@ from dulwich.protocol import (
     pkt_line,
 )
 from dulwich.pack import (
-    write_pack_data,
     write_pack_objects,
     PackChunkGenerator,
 )


efiop:
Ok, looks like there is more issues in dulwich/client.py. Taking a look...

jelmer:
Thanks; already resolved in 2f60f5bd89a37ad225a8f50e754b51b1bf4166d7

efiop:
@jelmer Thank you! There is also another mypy error.

Btw, looks like we all would really benefit from a pre-commit hook. I suppose you wouldn't mind if we contribute it?

jelmer:
The remaining test error appears to be some sort of timing issue on Mac OS X.

> Btw, looks like we all would really benefit from a pre-commit hook. I suppose you wouldn't mind if we contribute it?
I'd be open to a reasonable PR to add pre-commit hooks. That said, I don't necessarily work on Dulwich in a location with a ".git" control directory, so pre-commit hooks may not always get run. I originally planned to use PRs for all of my personal changes, but in practice that has been quite heavyweight especially for things like merging other people's changes and making small incremental changes on top.

jelmer:
On Mon, May 16, 2022 at 07:14:29AM -0700, Ruslan Kuprieiev wrote:
> > That said, I don't necessarily work on Dulwich in a location with a ".git" control directory, so pre-commit hooks may not always get run.
> 
> Hm, not sure what you mean. Do you mean that you are not using git to work on dulwich?
I'm using both dulwich itself and breezy, which interacts with git through dulwich but uses a different on-disk format.

> > I originally planned to use PRs for all of my personal changes, but in practice that has been quite heavyweight especially for things like merging other people's changes and making small incremental changes on top.
> 
> But it seems like master and releases are breaking from time to time because of that, which affects the stability of the project. Maybe we could offer some help maintaining it?
Help in general would be great, especially addressing some of the issues you've been submitting PRs for.

I do run "make check" before pushing, and that hasn't regressed recently as far
as I can tell. Happy to see "make check" extended to include typing and/or flake8. I'd
argue that both flake8 and typing warnings don't have an impact on the
stability of the project though.

Jelmer

PR #962, opened 2022-05-16
Explicitly reenable paramiko server on Gentoo for tests

mgorny:
Gentoo patches out server support from paramiko for security reasons.
Since dulwich uses it only for testing, use the knob to explicitly
reenable it.


mgorny:
Thank you!

PR #971, opened 2022-05-23
gha: don't forget to upload wheels

efiop:
Unfortunately wheel publishing got broken in https://github.com/jelmer/dulwich/commit/978cd0115bf16eaeca510e55028d12fc5d9344ba

We now upload built wheels into one `dist` artifact and then publish them all at once in one job, saving on workers.

Related to #963


efiop:
Here is a built artifact https://github.com/jelmer/dulwich/actions/runs/2373351950 . Seems like all wheels (20, same as in the latest dulwich release that had wheels) are in place and `publishing` step can see them.

jelmer (approved):

jelmer:
Thanks!

efiop:
Thank you for a quick release 🙏

PR #974, opened 2022-06-04
type-annotate the ConfigDict

dimbleby:


dulwich/config.py
```diff
@@ -252,35 +252,52 @@ def has_section(self, name: Tuple[bytes, ...]) -> bool:
...
+
+
+class ConfigDict(Config, MutableMapping[Key, MutableMapping[BytesLike, Value]]):
     """Git configuration stored in a dictionary."""
 
-    def __init__(self, values=None, encoding=None):
+    def __init__(
+        self,
+        values: Union[
+            MutableMapping[KeyLike, MutableMapping[BytesLike, Value]], None
```
dimbleby, line 268: I suspect it's desirable for this to be `ValueLike` rather than `Value`; but I don't see anything that uses `self.encoding` to convert the values here to `bytes` so I've typed it as I see it.

dulwich/config.py
```diff
@@ -252,35 +252,52 @@ def has_section(self, name: Tuple[bytes, ...]) -> bool:
         return name in self.sections()
 
 
-class ConfigDict(Config, MutableMapping):
+BytesLike = Union[bytes, str]
+Key = Tuple[bytes, ...]
+KeyLike = Union[bytes, str, Tuple[BytesLike, ...]]
+Value = Union[bytes, bool]
```
dimbleby, line 258: not sure whether a `bool` is really intended to be allowed here, but `set()` goes
```
if not isinstance(value, (bool, bytes)):
  ....
```
which suggests that it is.

Perhaps only bytes are intended, though, and that code in `set()` is redundant?

dulwich/config.py
```diff
@@ -335,18 +356,26 @@ def get(  # type: ignore[override]
...
+    def set(
+        self,
+        section: KeyLike,
+        name: BytesLike,
+        value: BytesLike,
+    ) -> None:
         section, name = self._check_section_and_name(section, name)
 
-        if type(value) not in (bool, bytes):
+        if not isinstance(value, bytes):
```
dimbleby, line 367: I decided that this was an accident and that the dictionary is not intended to hold booleans, say if that's wrong.

(There'd be some modest tweaks on the typing in this MR, and a couple of callers that should assert that they haven't got a surprise bool)

dimbleby:
Values being either `bytes` or `bool` is a bit of an oddity IMO but I've annotated the code as it is rather than as I might prefer it to be.

(There are a couple of callers in repo.py that set boolean values on a `ConfigFile`, converted into `"true"` and `"false"` in `write_to_file()`.  Maybe it would be simpler just to use `"true"` and `"false"` all along, that would save everyone from having to check what type they have when they get a value from a `ConfigDict`)

Workflow failure doesn't seem related to this change.

PR #976, opened 2022-06-07
add support for git credential helpers

### LINKED ISSUE #873: Support for credential helpers
It would be really nice to support credential loading from helpers like in the standalone `git` command‑line tool. Many users rely on HTTPS remotes and their system keychain for passwordless authentication, especially on macOS. 

### Minimal example

```python3
>>> from dulwich.porcelain import ls_remote
>>> ls_remote("https://github.com/user/repository")
```

#### Expected output

```console
{b'HEAD': b'···', b'refs/heads/main': b'···'}
```

#### Actual output

```console
  File "/···/site-packages/dulwich/porcelain.py", line 1618, in ls_remote
    return client.get_refs(host_path)
  File "/···/site-packages/dulwich/client.py", line 2089, in get_refs
    refs, _, _ = self._discover_references(b"git-upload-pack", url)
  File "/···/site-packages/dulwich/client.py", line 1906, in _discover_references
    resp, read = self._http_request(url, headers, allow_compression=True)
  File "/···/site-packages/dulwich/client.py", line 1875, in _http_request
    raise HTTPUnauthorized(resp.getheader("WWW-Authenticate"), url)
dulwich.client.HTTPUnauthorized: No valid credentials provided
```

### System configuration

```console
$ uname -a
Darwin ··· 19.6.0 Darwin Kernel Version 19.6.0: Tue Nov 10 00:10:30 PST 2020; root:xnu-6153.141.10~1/RELEASE_X86_64 x86_64
```
```console
$ python --version
Python 3.9.4
```
```console
$ pip show dulwich
Name: dulwich
Version: 0.20.23
Summary: Python Git Library
Home-page: https://www.dulwich.io/
Author: Jelmer Vernooij
Author-email: jelmer@jelmer.uk
License: Apachev2 or later or GPLv2
Location: /usr/local/lib/python3.9/site-packages
Requires: certifi, urllib3
Required-by: dvc
```
```console
$ cat /Applications/Xcode.app/Contents/Developer/usr/share/git-core/gitconfig
[credential]
        helper = osxkeychain
```

***

By the way: I love this project! ❤️

dtrifiro:
- add `CredentialHelper` class to run credential helper commands and retrieve values
- add `get_credentials_from_helper` to retrieve credentials from helpers defined in provided config
- moved (unused) `get_credentials_from_store` to `credentials.py`, refactor it to use `CredentialHelper` with `store` helper.

Also see https://www.git-scm.com/docs/gitcredentials

fixes #873


dtrifiro:
Would appreciate if you could take a look @pmrowla

dulwich/credentials.py
```diff
@@ -0,0 +1,122 @@
...
+        return self.command.split(" ")
+
+    def get(self, url: str) -> Dict[str, str]:
+        cmd = self._prepare_cmd()
+        if isinstance(cmd, str):
+            cmd = f"{cmd} get"
+        else:
+            cmd.append("get")
+
+        helper_input = f"url={url}\n".encode("UTF-8")
```
pmrowla, line 77: Should this be `ascii` and not UTF-8?

dulwich/credentials.py
```diff
@@ -0,0 +1,122 @@
...
+        except FileNotFoundError as exc:
+            raise CredentialNotFoundError("Helper not found") from exc
+
+        credentials = {}
+        for line in res.stdout.decode("UTF-8").strip().split("\n"):
+            try:
+                key, value = line.split("=")
+                credentials[key] = value
+            except ValueError:
+                continue
```
pmrowla, line 99: I don't think we should be making assumptions about text encodings here, and just handle credentials as `Dict[bytes, bytes]` rather than `[str, str]`.

Text encoding for HTTP basic auth username/password is technically undefined (https://www.rfc-editor.org/rfc/rfc7617#appendix-B.3), and depending on the credential manager/system locale settings/etc, I think it's possible that the returned credentials here may not be utf-8 or ascii.

Then later we can just do `b64encode(username + b':' + password)` where the username/pw are already bytes objects.

dtrifiro: Agreed and corrected.

Note though that when username/password kwargs are provided to `Urllib3HttpGitClient()`, we use `urllib3.utils.make_headers`:
https://github.com/jelmer/dulwich/blob/bb83067f5d021b80d208dd1b93eda8d02578eeb2/dulwich/client.py#L2187

which encodes the `basic_auth` string as `latin-1`:
https://github.com/urllib3/urllib3/blob/972e9f02cd219892701fa0a1129dda7f81dac535/src/urllib3/util/request.py#L101-L104

In practice though, Imost of the time user/pass will be provided as kwargs when using `from_parsedurl` and in this case username/password are already decoded through `urlparse`.

dulwich/credentials.py
```diff
@@ -0,0 +1,125 @@
+# credentials.py -- support for git credential helpers
+
+# Copyright (C) 2022 Daniele Trifirò <daniele@iterative.ai>
```
dtrifiro, line 3: @jelmer Please let me know if I need to update this header with your name :slightly_smiling_face:

dulwich/credentials.py
pmrowla: When using creds from a credential manager we should be generating the basic auth header ourselves. (We should be following the RFC, not urllib3)

dulwich/client.py
```diff
@@ -2236,7 +2238,60 @@ def _http_request(self, url, headers=None, data=None):
...
+                    credentials = next(self.get_credentials(url))
+                except StopIteration:
+                    raise exc
+
+                import base64
+
+                basic_auth = credentials[b"username"] + b":" + credentials[b"password"]
+                self.pool_manager.headers.update(
+                    {
+                        "authorization": f"Basic {base64.b64encode(basic_auth).decode('UTF-8')}"
```
pmrowla, line 2270:
```suggestion
                        "authorization": f"Basic {base64.b64encode(basic_auth).decode('ascii')}"
```

dulwich/tests/test_client.py
```diff
@@ -1130,6 +1131,74 @@ def request(self, method, url, fields=None, headers=None, redirect=True, preload
...
+
+        class PoolManagerMock:
+            def __init__(self):
+                self.headers = {}
+
+            def request(self, method, url, headers, **kwargs):
+                if (
+                    "authorization" in headers
+                    and headers["authorization"]
+                    == f"Basic {encoded_credentials.decode('UTF-8')}"
```
pmrowla, line 1152:
```suggestion
                    == f"Basic {encoded_credentials.decode('ascii')}"
```

dtrifiro:
Added some more tests with the latest push

dulwich/credentials.py
jelmer: This is fine.

jelmer (commented):
.

dulwich/credentials.py
```diff
@@ -0,0 +1,125 @@
...
+            # On Windows this will only work in git-bash and/or WSL2
+            self._run_kwargs["shell"] = True
+
+    def _prepare_cmd(self) -> Union[str, List[str]]:
+        if self.command[0] == "!":
+            return self.command[1:]
+
+        argv = self.command.split(" ")
+        if not os.path.isabs(argv[0]):
+            cmd = ["git", f"credential-{argv[0]}"]
```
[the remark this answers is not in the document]
pmrowla: To clarify the intention here isn't to introduce any dependency on C Git, the expected behavior for 3rd party credential helpers is to follow the executable naming conventions set by C Git:

> Generally speaking, rule (3) above is the simplest for users to specify. Authors of credential helpers should make an effort to assist their users by naming their program "git-credential-$NAME", and putting it in the $PATH or $GIT_EXEC_PATH during installation, which will allow a user to enable it with git config credential.helper $NAME.

Although that does mean we probably should just use `git-credential-{name}` instead of relying on `git credential-{name}` @dtrifiro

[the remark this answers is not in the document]
dtrifiro: I agree.

The rationale for using `git credential-foo` instead of `git-credential-foo` here was that (I'm assuming) most users will be using cgit provided credential helpers (such as `git-credential-store`, `git-credential-osxkeychain`, `git-credential-gnome-keyring`, etc), which are in `GIT_EXEC_PATH` (which is not in `PATH`). Using `git` made sure that `GIT_EXEC_PATH` was also checked.

Pushed a patch which fixes the issue: if `git-credential-foo` is not in `PATH` and `git` cli is available, `GIT_EXEC_PATH` is retrieved using `git --exec-path` and we check if `git-credential-foo` is available in `GIT_EXEC_PATH`. If `git` cli is not available, nothing is done, and calling `subprocess.run` will just raise `FileNotFoundError` (which will result in `CredentialNotFound` exception)

dulwich/client.py
```diff
@@ -2236,7 +2238,63 @@ def _http_request(self, url, headers=None, data=None):
...
+                return super()._http_request(url, *args, **kwargs)
+            except (
+                HTTPUnauthorized,
+                NotGitRepository,  # some git servers returns 404 instead of 401 (github)
+            ) as exc:
+                if self.credentials is None:
+                    self.credentials = self.get_credentials(url)
+                try:
+                    credentials = next(self.credentials)
+                except StopIteration:
```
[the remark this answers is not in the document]
dtrifiro: It should not, there's a test for that here:
https://github.com/dtrifiro/dulwich/blob/f3ed0104e17c3b9c74952fe62f401008cb590579/dulwich/tests/test_client.py#L1205

dtrifiro:
Hey @jelmer, any updates?

jelmer:
Sorry for the delay, I'll try to have another look at it this week.

dtrifiro:
Temporarily marking this as draft while I work on a patch for a bug in this implementation (as well as making this more closely match cli git's behaviour)

dtrifiro:
While fixing a bug, I spent some time doing cleanups/refactors. I dropped `GitCredentialsHttpClient` and moved git credentials management in `Urllib3HttpGitClient`, as well as getting rid of of the retry logic and add some url-matching functions to more that try to match cli git behaviour. Overall I feel like the code is a bit cleaner and behavior more predictable and consistent with cli git.

bakaleks:
Hi @jelmer I would really appreciate if this PR gets merged. Thanks.

d-miketa:
Same, without it I can't reach private Bitbucket repos. :(

jelmer:
Sorry, I'm currently traveling but will be able to take a look in a couple of days.

oslobowl:
Hi guys :) Do you know when this branch will be merged, approximately? Thanks a lot :)

jelmer (changes requested):
Sorry for sitting on this for so long; I'm a bit concerned about how this makes the interface for authentication more complicated. Ideally there should just be one pythonic interface for authentication, with perhaps one implementation of that that can interact with C git.

dulwich/credentials.py
```diff
@@ -0,0 +1,295 @@
...
+                    ):
+                        return parsed_line.username, parsed_line.password
+        except FileNotFoundError:
+            # If the file doesn't exist, try the next one.
+            continue
+
+
+def get_credentials_from_helper(base_url: str, config) -> Tuple[bytes, bytes]:
+    """Retrieves credentials for the given url from git credential helpers"""
+    if isinstance(config, StackedConfig):
```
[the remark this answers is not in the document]
dtrifiro: Given an URL, git loops though the available configs and matches urls in credentials sections with the provided URL, which is something we cannot do with the config API.

For example with a config like this:
```
[credential "https://github.com/jelmer/dulwich"]
    helper = githubhelper

[credential]
    helper = generichelper
```
Running 
```python
config.get((b"credential", b"https://github.com))
```

would return `generichelper` instead of `githubhelper`, which is not what we want.

I refactored this a bit so that we use `urlmatch_credential_sections` to cycle though config sections that match the URL.
This could be extracted to become a config API method, although I'm not sure if there's any other use cases where one needs to match URLs in sections, apart from the credential helpers scenario.

dulwich/credentials.py
```diff
@@ -0,0 +1,295 @@
...
+        # falling back to the generic section if there's no match or
+        # if url is None
+        section = urlmatch_credential_section(config, url)
+
+        command = config.get(section, b"helper")
+        assert command and isinstance(command, bytes)
+        return cls(command.decode(encoding))
+
+
+def get_credentials_from_store(
```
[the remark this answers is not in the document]
dtrifiro: I moved `get_credentials_from_store` from `client` to here since it seemed the right place for it, but it is dead code. I would remove this if that's ok with you.

dulwich/credentials.py
```diff
@@ -0,0 +1,295 @@
...
+            return False
+
+    if any(
+        (
+            (parsed.hostname and valid_url.hostname != parsed.hostname),
+            (parsed.username and valid_url.username != parsed.username),
+            (parsed.port and valid_url.port != parsed.port),
+            (
+                parsed.path
+                and not valid_url.path.rstrip("/").startswith(parsed.path.rstrip("/"))
```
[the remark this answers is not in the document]
dtrifiro: Indeed, fixed.

dtrifiro:
> Sorry for sitting on this for so long

Hi @jelmer, thanks for taking the time to review this :slightly_smiling_face: 

> I'm a bit concerned about how this makes the interface for authentication more complicated. Ideally there should just be one pythonic interface for authentication, with perhaps one implementation of that that can interact with C git.

I'm not sure I understand what you're proposing here, but below you'll find my thoughts. Please let me know if I misunderstood and/or if there's any other parts you find complicated you can point me to, so I can have another look and take action.

With my last push I streamlined `get_credentials_from_helpers` to explicitly how the configs sections and values are being used. The only other method to get credentials is `get_credentials_from_store` which, as far as I can tell, is unused. We could either refactor this or get rid of it (as I proposed in another comment).

jelmer:
sorry about that, looking at it again now

jelmer:
> sorry about that, looking at it again now

Haven't forgotten about this, will probably put forward a PR with a subset of your changes later this week.

jelmer:
Sorry again for the long delay. As a means of trying to move this forward, I've at least cherry-pick the config-related bits of this PR that are non-controversial.

jelmer:
> Hey @jelmer , thanks for taking a look! 🙏  FYI: @dtrifiro broke his clavicle/collarbone and won't be able to reply for a few weeks. If there is anything he needs to address, he'll get back to it once he's recovered.

Ah, thanks - good to know. I hope his recovery goes well!

PR #979, opened 2022-06-13
Fix pushing to a repository served by Dulwich itself

### LINKED ISSUE #977: Push fails with unsupported capabilities
Hi

I have a test in hg-git that checks interacting with a repository, served by Dulwich itself. Unfortunately, [that test fails](https://foss.heptapod.net/mercurial/hg-git/-/jobs/557263#L68) with recent releases; somehow, pushing triggers this error:

```
dulwich.errors.GitProtocolError: Client asked for capability b'side-band' that was not advertised.
```

As best I can tell, there's no logic related to capabilities in hg-git, so the error seems somewhat surprising. Bisecting recent history reveals that the bug is caused by commit e9faf83ac437, and undoing its changes fixes the test.

What are your thoughts on this? Is this a regression in Dulwich, or is there something we're doing wrong?

Thanks

— Dan

danchr:
This contains a backout/revert of commit e9faf83ac437, which broke pushing to a Dulwich-served repository using Dulwich itself. The hg-git testsuite was affected by this; it broke one of our tests.

Fixes #977


jelmer:
Thanks for the PR - I've pushed a proper fix for this rather than a revert; thanks for tests - please do let us know if you run into other issues.

PR #984, opened 2022-06-30
more typechecking

dimbleby:
Continuing to add type annotations to config-related classes.

That there are places in the code that put booleans into the configuration remains a wart, and its influence grows as the types pass that fact along.  It might be a worthwhile clean up to have [this code](https://github.com/jelmer/dulwich/blob/bbf100b63fb00ebb4a1d0ca698b1abe337611bae/dulwich/repo.py#L368-L374) use the strings "true" and "false" and remove `bool` from `Value` and `ValueLike`...

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -2285,6 +2285,7 @@ def iter_instead_of(config: Config, push: bool = False) -> Iterable[Tuple[str, s
             except KeyError:
                 pass
         for needle in needles:
+            assert isinstance(needle, bytes)
             yield needle.decode('utf-8'), replacement.decode('utf-8')
 
 
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -34,11 +34,13 @@ from typing import (
     BinaryIO,
     Iterable,
     Iterator,
+    List,
     KeysView,
     MutableMapping,
     Optional,
     Tuple,
     Union,
+    overload,
 )
 
 from dulwich.file import GitFile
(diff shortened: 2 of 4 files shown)


jelmer:
> That there are places in the code that put booleans into the configuration remains a wart, and its influence grows as the types pass that fact along. It might be a worthwhile clean up to have [this code](https://github.com/jelmer/dulwich/blob/bbf100b63fb00ebb4a1d0ca698b1abe337611bae/dulwich/repo.py#L368-L374) use the strings "true" and "false" and remove `bool` from `Value` and `ValueLike`...

That would mean that each caller has to do its own interpretation of boolean values though - the possible values aren't just "true" and "false".

dimbleby:
`Config` has a `get_boolean()` which callers can use to read back boolean values, it could also gain a `set_boolean()`.  

That would allow the internals to deal with `bytes` only

Then callers who don't explicitly `get_boolean()` would not have to check every time whether they got `bytes` or `bool`.

jelmer:
On Thu, Jun 30, 2022 at 09:22:27AM -0700, David Hotham wrote:
> `Config` has a `get_boolean()` which callers can use to read back boolean values, it could also gain a `set_boolean()`.  
> 
> That would allow the internals to deal with `bytes` only
> 
> Then callers who don't explicitly `get_boolean()` would not have to check every time whether they got `bytes` or `bool`.

Yeah, that makes sense - my main concern is that callers shouldn't be
involved in parsing/formatting config files.

Jelmer

-- 
Jelmer Vernooĳ ***@***.***>

PR #996, opened 2022-08-01
Allow missing gpg package during tests

kulikjak:
`gpg` is one of the extra optional packages, but when running tests, it's currently required (otherwise, several tests fail with `ImportError`). This PR skips those tests if `import gpg` fails.

### DIFF
--- a/dulwich/tests/test_porcelain.py
+++ b/dulwich/tests/test_porcelain.py
@@ -66,6 +66,11 @@ from dulwich.web import (
     make_wsgi_chain,
 )
 
+try:
+    import gpg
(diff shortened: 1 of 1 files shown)


kulikjak:
Uh, weird. I can reproduce this locally, but none of these warning/errors seem to be related to my change as they are in different files - and I see the same issue on current master with no changes.

jelmer:
This is due to the new flake8 version that adds more checks - I've just merged a PR that works around it by pinning an older version.

kulikjak:
Thanks!

PR #997, opened 2022-08-02
gha: temporarily pin flake8 to 4.0.1

dtrifiro:
5.x started reporting new issues which were making CI fail.


jelmer:
Thanks for the PR!

I'd prefer to just fix the new issues that are now reported and unpin again. Whether we should use black is a separate conversation IMHO.

PR #1001, opened 2022-08-22
Add 3.11 support

skshetry:
Python 3.11.0rc1 is now released, see https://pythoninsider.blogspot.com/2022/08/python-3110rc1-is-now-available.html, which is guaranteed to be ABI compatible with the final release.


jelmer:
Thanks! Can you update setup.py as well, to list 3.11 as being supported?

PR #1017, opened 2022-09-01
Document authentication with porcelain.clone

TuringTux:
Currently, the porcelain documentation does not mention that `porcelain.clone` supports authentication using the `username` and `password` parameters.

I think it would be nice to mention this especially for new users of the library like me, who go straight to the porcelain section of the documentation because they just want to clone a repository in Python. It took me a while to figure out that authentication is possible, so I'd like to ease that burden for all future new users of Dulwich by explicitly stating that possibility :)


jelmer:
Makes sense; thanks for the patch!

TuringTux:
Thanks for the quick response! I just noted you forget an "a", I have added in back in, but probably disabled the auto merge by doing so.

PR #1019, opened 2022-09-01
Prettify md/yaml files

kianmeng:
Resolved via `prettier -w .`


jelmer:
Not convinced by the value of this, but I guess.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #1026, opened 2022-09-03
Implement timezone offset detection

springheeledjack0:
This PR adds functionality to detect local timezone offsets to porcelain.
[date-formats.txt](https://raw.githubusercontent.com/git/git/v2.3.0/Documentation/date-formats.txt).
~~The implementation optionally supports [```python-dateutil```](https://github.com/dateutil/dateutil) to parse ISO 8601 formatted strings.~~

RFC on some details:
- move ```parse_timezone_format()``` out of porcelain?
- different ordering of formats?
- use author date instead of committer date in ```tag_create()```? (currently works like C git)

IMO, only porcelain should support timezone detection. All "internal/low-level" functions should just default to 0.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -266,7 +266,7 @@ def parse_timezone_format(tz_str):
     raise TimezoneFormatError(tz_str)
 
 
-def get_user_timezone():
+def get_user_timezones():
     """Retrieve local timezone as described in
     https://raw.githubusercontent.com/git/git/v2.3.0/Documentation/date-formats.txt
     Returns: A tuple containing author timezone, committer timezone
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -401,7 +401,9 @@ def commit(
     repo=".",
     message=None,
     author=None,
+    author_timezone=None,
     committer=None,
+    commit_timezone=None,
     encoding=None,
     no_verify=False,
     signoff=False,
@@ -412,7 +414,9 @@ def commit(
       repo: Path to repository
       message: Optional commit message
       author: Optional author name and email
+      author_timezone: Author timestamp timezone
       committer: Optional committer name and email
+      commit_timezone: Commit timestamp timezone
       no_verify: Skip pre-commit and commit-msg hooks
       signoff: GPG Sign the commit (bool, defaults to False,
         pass True to use default GPG key,
@@ -426,7 +430,11 @@ def commit(
         author = author.encode(encoding or DEFAULT_ENCODING)
     if getattr(committer, "encode", None):
         committer = committer.encode(encoding or DEFAULT_ENCODING)
-    author_timezone, commit_timezone = get_user_timezones()
+    local_timezone = get_user_timezones()
+    if author_timezone is None:
+        author_timezone = local_timezone[0]
+    if commit_timezone is None:
+        commit_timezone = local_timezone[1]
     with open_repo_closing(repo) as r:
         return r.do_commit(
             message=message,
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -448,7 +448,7 @@ def commit(
         author = author.encode(encoding or DEFAULT_ENCODING)
     if getattr(committer, "encode", None):
         committer = committer.encode(encoding or DEFAULT_ENCODING)
-    author_timezone, commit_timezone = get_user_timezone()
+    author_timezone, commit_timezone = get_user_timezones()
     with open_repo_closing(repo) as r:
         return r.do_commit(
             message=message,
@@ -1056,7 +1056,7 @@ def tag_create(
                 tag_time = int(time.time())
             tag_obj.tag_time = tag_time
             if tag_timezone is None:
-                tag_timezone = get_user_timezone()[1]
+                tag_timezone = get_user_timezones()[1]
             elif isinstance(tag_timezone, str):
                 tag_timezone = parse_timezone(tag_timezone)
             tag_obj.tag_timezone = tag_timezone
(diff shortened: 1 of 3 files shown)


dulwich/porcelain.py
```diff
@@ -191,6 +191,87 @@ class RemoteExists(Error):
...
+                # `The time zone offset in seconds, or (since version 2.6.0, represented as a datetime.timedelta object).`
+                if isinstance(delta, int):
+                    return delta
+                return int(delta.total_seconds())
+        except (AttributeError, TypeError, ValueError):
+            pass
+    except (ModuleNotFoundError, ImportError):
+        # Supported offsets:
+        # sHHMM, sHH:MM, sHH
+        iso_8601_pattern = re.compile("[0-9]([+-])([0-9]{2})(?::(?=[0-9]{2}))?([0-9]{2})?$")
```
jelmer, line 240: this is actually just a small subset of iso8601. Does git support "proper" iso8601 or just rfc3339 ?

https://stackoverflow.com/questions/522251/whats-the-difference-between-iso-8601-and-rfc-3339-date-formats

dulwich/porcelain.py
```diff
@@ -191,6 +191,87 @@ class RemoteExists(Error):
     """Raised when the remote already exists."""
 
 
+def _parse_timezone_format(tz_str):
```
jelmer, line 194: Let's make this public? This should probably be parse_timezone rather than parse_timezone_format, since it's not parsing the format but the timezone itself?

dulwich/porcelain.py
```diff
@@ -191,6 +191,87 @@ class RemoteExists(Error):
...
+            sign, hours, minutes = match.groups()
+            total_secs += int(hours) * 3600
+            if minutes:
+                total_secs += int(minutes) * 60
+            total_secs = -total_secs if sign == "-" else total_secs
+            return total_secs
+
+    # YYYY.MM.DD, MM/DD/YYYY, DD.MM.YYYY contain no timezone information
+
+    return 0
```
jelmer, line 253: If the timezone is unparseable, then please throw an exception rather than return 0.

dulwich/porcelain.py
```diff
@@ -191,6 +191,87 @@ class RemoteExists(Error):
...
+
+
+def get_user_timezone():
```
jelmer, line 269: get_user_timezones(), since it returns two?

dulwich/porcelain.py
```diff
@@ -354,11 +435,14 @@ def commit(
         author = author.encode(encoding or DEFAULT_ENCODING)
     if getattr(committer, "encode", None):
         committer = committer.encode(encoding or DEFAULT_ENCODING)
+    author_timezone, commit_timezone = get_user_timezone()
```
jelmer, line 451: this should probably also allow specifying the two timezones as argument

dulwich/porcelain.py
```diff
@@ -191,6 +191,87 @@ class RemoteExists(Error):
...
+        except ValueError:
+            pass
+
+    # RFC 2822
+    import email.utils
+    rfc_2822 = email.utils.parsedate_tz(tz_str)
+    if rfc_2822:
+        return rfc_2822[9]
+
+    # ISO 8601
```
jelmer, line 228: Not a big fan of supporting two codepaths here. At the very least, could you make sure that both paths are covered by the testsuite?

jelmer:
Thanks for the PR! Overall, looks good. I've left some smaller comments (especially around the iso8601 handling) inline.

dulwich/porcelain.py
springheeledjack0: Haven't really dug into the code of git itself.
I was mainly using [this StackOverflow answer](https://stackoverflow.com/questions/19742345/what-is-the-format-for-date-parameter-of-git-commit/19742762#19742762) and [this test](https://github.com/git/git/blob/96b2d4fa927c5055adc5b1d08f10a5d7352e2989/t/t6300-for-each-ref.sh#L128) as guidance.

dulwich/porcelain.py
springheeledjack0: Should we throw a custom Exception for this?

dulwich/porcelain.py
springheeledjack0: I named it ```parse_timezone_format``` because there already is a ```parse_timezone``` in ```objects```, that this very function is using.

dulwich/porcelain.py
springheeledjack0: Resolved in 33d9f65866cfffe212bec969f1a5afd79a8ebc7a

jelmer (commented):
sorry, forgot to hit send on pending comments :-/

dulwich/porcelain.py
springheeledjack0: Makes sense. ae14d3ebf28d25f76468efec137e917b8a416970

dulwich/porcelain.py
springheeledjack0: All kinds of cases I found so far are supported without needing dateutil. I was anticipating it to maybe catch some edge cases, but I guess we can also drop it?

jelmer: Yeah, let's just do that. If we find that there are other cases where we do want to rely on dateutil because handling them with custom code here is too complicated, we can always still add it later.

springheeledjack0: Alright, done in 322ad9bc910b238f0e3d60ef9841fb3914a22aaa.

dulwich/porcelain.py
springheeledjack0: Resolved in 4ca95f5d56ff50b31c3c465cac1c256984a66b08

dulwich/porcelain.py
springheeledjack0: Made public in 33d9f65866cfffe212bec969f1a5afd79a8ebc7a

PR #1029, opened 2022-09-04
Drop use of certifi to find ca certificate path

### LINKED ISSUE #1025: System SSL certificates aren't used
My company uses its own certificate, which I have installed on my system. Cloning a repository from our company gitlab instance works with `git`, but not with `dulwich`.

Log of the following script is [here](https://gist.github.com/Kuba314/4c3a40bc445008e87f340644c383763a).
```py
from dulwich import porcelain

porcelain.clone("https://company.gitlab.instance.com/namespace/project.git", "test-repo")
```
Note: I've replaced the url, project name and project namespace with placeholder values.

I've found this issue through poetry. Here's the original issue: https://github.com/python-poetry/poetry/issues/6340

jelmer:
Drop use of certifi to find ca certificate path.
  
This seems to be having adverse affects where we specify an incorrect path in some cases. Fixes #1025
  
Increment minimum urllib3 to >= 1.25 just to be sure; not strictly necessary -
we're already trying to be explicit about whether we want verification.
  
This was originally introduced in 48e2ef8c2f2c65a33ad3276458791c19ef208c1a, but
urllib3 now verifies certificates by default.
See also https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification
  
thanks to @dimbleby  for help with the archeology on this one :)

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1863,7 +1863,9 @@ def default_urllib3_manager(   # noqa: C901
 
     headers = {"User-agent": user_agent}
 
-    kwargs = {}
+    kwargs = {
+        "ca_certs" : ca_certs,
+    }
     if ssl_verify is True:
         kwargs["cert_reqs"] = "CERT_REQUIRED"
     elif ssl_verify is False:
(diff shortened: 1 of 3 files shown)


dulwich/client.py
```diff
@@ -1854,9 +1854,9 @@ def default_urllib3_manager(   # noqa: C901
             ssl_verify = True
 
         try:
-            ca_certs = config.get(b"http", b"sslCAInfo")
+            kwargs["ca_certs"] = config.get(b"http", b"sslCAInfo")
```
dimbleby, line 1857: kwargs is not defined yet

PR #1042, opened 2022-09-14
Catch SymrefLoop in set_if_equals

domdfcoding:
Previously `self.follow(name)` raised `KeyError` which was handled, but now it raises `SymrefLoop` the exception is passed up to the caller. With this patch the old behaviour is restored.


jelmer:
Thanks! Any change you could add a test for this case as well?

PR #1046, opened 2022-09-19
display stderr output from git programs run by the test suite

stspdotname:
This makes it easier to determine why a Git command failed. In my case, while exploring Git-protocol-v2 support, it is very helpful to see what git-http-backend complains about when my WIP implementation sends an invalid request.


stspdotname:
There was test fallout in the first version of this patch. I missed that the git --version tests override the run_git() function. This has been fixed in the forced-pushed update; tests are now passing locally for me.

PR #1058, opened 2022-10-05
fix for windows usage as "username getpass.getuser" is not available

nokeechia:
fix for windows usage as "username getpass.getuser" is not available
Added os.getlogin() for windows usage

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -139,11 +139,11 @@ class InvalidUserIdentity(Exception):
 def _get_default_identity() -> Tuple[str, str]:
     import getpass
     import socket
-
-    username = getpass.getuser()
     try:
+        username = getpass.getuser()
         import pwd
     except ImportError:
+        username = os.getlogin()
         fullname = None
     else:
         try:


dulwich/repo.py
```diff
@@ -140,10 +140,12 @@ def _get_default_identity() -> Tuple[str, str]:
     import getpass
     import socket
 
-    username = getpass.getuser()
```
jelmer, line 143: I believe the USER environment variable should be present on Windows - see https://pureinfotech.com/list-environment-variables-windows-10/. Where are you finding that it's missing?

jelmer: getpass should always be available IIUC

jelmer:
I believe the USER environment variable should be present on Windows - see https://pureinfotech.com/list-environment-variables-windows-10/. getpass.getuser should always be present.Where are you finding that it's missing?

jelmer:
I've refactored this code a little, but am still curious in what situation you end up without the appropriate environment variables on Windows. Since I haven't heard back, I'll close this PR but please feel free to open a new one if you still need improvements here.

PR #1084, opened 2022-11-02
Allow passing abbrev to describe

nanonyme:
This provides ability for user to request full commit through API


jelmer:
Thanks! Can you also add a test in dulwich/tests/test_porcelain.py ? (around line 2947)

jelmer (changes requested):
See other comment

nanonyme:
Force pushed to fix flake8 nit about extra newline.

nanonyme:
It looks like docs is failing because it's running on Python 3.11 which is incompatible with pydoctor used.

PR #1085, opened 2022-11-05
Add Python 3.9 into tox config

nanonyme:
Latest Debian stable has this and adding it makes development easier. Also looks like this no longer supports Python 3.5 so removing


jelmer:
Thanks, merging. 

I don't really use tox though, so this is likely to regress again.

PR #1086, opened 2022-11-05
Simplify tox setup

nanonyme:
Since upstream doesn't predominantly use tox, just remove all extra boilerplate so tox will work out of the box with whatever Python is.


nanonyme:
@jelmer all the boilerplate in the file seems to me mostly be for running tests against multiple Python versions on same machine. But if this is not in reality done, I'm not sure what the point of having the boilerplate is. Having the boilerplate at all means it needs to be kept up to date.

PR #1089, opened 2022-11-08
empty proxy environment variables are now ignored

afaul:
This small change would allow the usage of
`http_proxy='' command`
as for example in
`http_proxy='' dvc list ...`
in bash scripts to temporary overwrite a proxy configuration by environment variables.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1833,7 +1833,7 @@ def default_urllib3_manager(   # noqa: C901
     if proxy_server is None:
         for proxyname in ("https_proxy", "http_proxy", "all_proxy"):
             proxy_server = os.environ.get(proxyname)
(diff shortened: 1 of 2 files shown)


dulwich/client.py
```diff
@@ -1833,7 +1833,7 @@ def default_urllib3_manager(   # noqa: C901
     if proxy_server is None:
         for proxyname in ("https_proxy", "http_proxy", "all_proxy"):
             proxy_server = os.environ.get(proxyname)
-            if proxy_server is not None:
+            if proxy_server is not None and proxy_server != "":
```
jelmer, line 1836: perhaps just "if not proxy_server"?

afaul: This condition is much clearer. I just negated it to break the loop if a non empty value is set and added it to the pull request.

PR #1096, opened 2022-11-11
Cope with missing content type

jelmer:
Cope with missing content type

Unfortunately our testing story for dumb remotes isn't great, so no updates there :-/

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1943,7 +1943,9 @@ class AbstractHttpGitClient(GitClient):
             base_url = resp.redirect_location[: -len(tail)]
 
         try:
-            self.dumb = not resp.content_type.startswith("application/x-git-")
+            self.dumb = (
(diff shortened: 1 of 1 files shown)


dimbleby:
https://github.com/python-poetry/poetry/issues/6999 presumably.

Not sufficient to make that case work, though, now it hits https://github.com/jelmer/dulwich/blob/0b78b674478f1e56480617617ac973a0917fdd97/dulwich/client.py#L2092-L2093

jelmer:
> https://github.com/python-poetry/poetry/issues/6999 presumably.
> 
> Not sufficient to make that case work, though, now it hits https://github.com/jelmer/dulwich/blob/0b78b674478f1e56480617617ac973a0917fdd97/dulwich/client.py#L2092-L2093

Yup, see https://github.com/jelmer/dulwich/issues/1097 for that.

PR #1098, opened 2022-11-12
Implementation to honor no_proxy environment variable

afaul:
This implementation will honor a no_proxy environmental variable.
That will resolve the issue #932.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1811,7 +1811,7 @@ def default_user_agent_string():
 
 
 def default_urllib3_manager(   # noqa: C901
-    config, pool_manager_cls=None, proxy_manager_cls=None, **override_kwargs
+    config, pool_manager_cls=None, proxy_manager_cls=None, base_url=None, **override_kwargs
 ) -> Union["urllib3.ProxyManager", "urllib3.PoolManager"]:
     """Return urllib3 connection pool manager.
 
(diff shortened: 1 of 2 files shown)


dulwich/client.py
```diff
@@ -1836,6 +1837,53 @@ def default_urllib3_manager(   # noqa: C901
             if proxy_server:
                 break
 
+    if base_url and proxy_server:  # only check for no_proxy environment variable if a proxy is set and a base_url is provided
+        no_proxy_str = os.environ.get("no_proxy")
+        if no_proxy_str:
```
jelmer, line 1842: please factor this out into a separate function

dulwich/client.py
```diff
@@ -1836,6 +1837,53 @@ def default_urllib3_manager(   # noqa: C901
             if proxy_server:
                 break
 
+    if base_url and proxy_server:  # only check for no_proxy environment variable if a proxy is set and a base_url is provided
+        no_proxy_str = os.environ.get("no_proxy")
```
jelmer, line 1841: please factor this out into a separate function

jelmer (changes requested):

dulwich/client.py
afaul: done

jelmer (approved):

PR #1105, opened 2022-12-12
Fix pack objects cli

stspdotname:
This pull request fixes a couple of issues I found while trying to test packfile generation with the 'dulwich pack-objects' command.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -135,7 +135,7 @@ def hex_to_filename(path, hex):
     # os.path.join accepts bytes or unicode, but all args must be of the same
     # type. Make sure that hex which is expected to be bytes, is the same type
     # as path.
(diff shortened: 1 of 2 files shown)


dulwich/objects.py
```diff
@@ -135,7 +135,7 @@ def hex_to_filename(path, hex):
     # os.path.join accepts bytes or unicode, but all args must be of the same
     # type. Make sure that hex which is expected to be bytes, is the same type
     # as path.
-    if getattr(path, "encode", None) is not None:
+    if type(path) != type(hex) and getattr(path, "encode", None) is not None:
```
jelmer, line 138: This looks like just a typo to me; shouldn't it be checking for "decode" rather than "encode" here, in which case we don't need the type check?

PR #1106, opened 2022-12-13
allow for reuse of existing deltas while creating pack files

stspdotname:
This feature allows the ``dulwich pack-objects`` command to complete in a somewhat reasonable amount of time on small repositories, provided a pack file which can provide deltas already exists.

Reusing deltas from an existing pack file avoids deltification overhead while creating a new pack file. This optimization is especially important when serving clones and fetches from a server.

For testing purposes, add two options to ``dulwich pack-objects``:
  --deltify: enables deltification and delta-reuse
  --no-reuse-deltas: disables delta-reuse during deltification

The write_pack_objects() now accepts an optional Pack file object argument which enables reuse of deltas.

If deltification is enabled and delta-reuse is not explicitly disabled, the pack_objects() routine in porcelain.py picks a suitable pack file for reuse, which is the pack file which contains the largest amount of objects. A delta found in this pack file can be reused if its delta-base will also be part of the generated pack file.

There are some limitations which should be lifted eventually:

Reused deltas are stored as ref-deltas only. There is no logic yet to figure out where the base of a reused offset-delta has been placed in the new pack file.

We decompress deltas for reuse and compress them again when writing them to the new pack file. More code refactoring would be needed to support direct copying from the source pack file in order to avoid re-compression overhead.

### DIFF
--- a/dulwich/cli.py
+++ b/dulwich/cli.py
@@ -523,12 +523,18 @@ class cmd_ls_tree(Command):
 
 class cmd_pack_objects(Command):
     def run(self, args):
-        opts, args = getopt(args, "", ["stdout"])
+        deltify = False
+        reuse_deltas = True
+        opts, args = getopt(args, "", ["stdout", "deltify", "no-reuse-deltas"])
         opts = dict(opts)
         if len(args) < 1 and "--stdout" not in opts.keys():
             print("Usage: dulwich pack-objects basename")
             sys.exit(1)
         object_ids = [line.strip() for line in sys.stdin.readlines()]
+        if "--deltify" in opts.keys():
+            deltify = True
+        if "--no-reuse-deltas" in opts.keys():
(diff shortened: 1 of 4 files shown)


dulwich/cli.py
```diff
@@ -523,12 +523,18 @@ def run(self, args):
 
 class cmd_pack_objects(Command):
     def run(self, args):
-        opts, args = getopt(args, "", ["stdout"])
+        deltify = False
```
jelmer, line 526: Ideally we should migrate this to argparse, like some of the other commands in this file.

jelmer (approved):
I'm not entirely sure about the API - in particular, shouldn't we ideally be comparing against all packs rather than just the largest?

It's a great improvement though. So let's do that, and then we can iterate further.

Thanks!

PR #1108, opened 2022-12-31
Add 'pack-refs' command

### LINKED ISSUE #1107: Add ability to pack refs
With how hg-git manages refs — only updating them on pull or push, and not on commit — I believe it makes sense to me to always pack them. Within the context of hg-git, refs are only updated on push or pull, but read _a lot_ — most commands that display a commit will always decorate it, and that involves reading all refs. In addition, users are also likely to notice after the fact that some branch or tag name cannot be stored on disk, such as a [tag name containing double quotes on Windows](https://foss.heptapod.net/mercurial/hg-git/-/issues/397). With regular Git, that would typically trigger an error on creation or pull, and the user might simply elect to skip that ref — with hg-git, on the other hand the name might already be enshrined in the version control history, and not easily skipped.

So, it makes sense to always pack them, I believe. This should make reading all refs slightly faster, while making updates somewhat slower. That shouldn't be much of an issue in hg-git, though, as it already has to manage a full mapping of commits anyway.

I managed to get this working with the current version of Dulwich, but it seems to me that this code really doesn't belong in hg-git, except possibly as a compatibility workaround. What do you think of adding something like this to Dulwich?

```python
def set_refs(git, new_refs):
    if new_refs:
        f = file.GitFile(os.path.join(git.refs.path, b"packed-refs"), "wb")

        try:
            # reread cached refs from disk, while holding the lock
            packed_refs = git.refs.get_packed_refs()

            for ref in new_refs:
                if ref not in packed_refs and ref_exists(ref, git.refs):
                    del git.refs[ref]

            packed_refs.update(new_refs)

            refs.write_packed_refs(f, packed_refs, git.refs._peeled_refs)
            f.close()

            git.refs._packed_refs = packed_refs
        finally:
            f.abort()
```

danchr:
This contains the following changes:

* [Handle more errors when trying to read a ref](https://github.com/jelmer/dulwich/commit/2339d6a0a8e76573ad1828cd33b3ecfdd859e87b)
   Some files aren't just not there, but actually invalid. Different operating systems handle that differently, but rather than getting cute and specific, treat any error when reading a ref as indicating that it's not there.
  
  (As an example, double quotes raise one type of error on Windows; non-UTF-8 paths another on macOS, and e.g. ZFS on Linux triggers a third if UTF-8 is active.)

  See: https://foss.heptapod.net/mercurial/hg-git/-/issues/397
* [Make the GitFile context manager discard writes on exceptions](https://github.com/jelmer/dulwich/commit/199d7d6d7d9f43c90484a1d2c390602cf706e353)
* [Add 'pack-refs' command](https://github.com/jelmer/dulwich/commit/ae14e4ee8249f2b54c388e9bb42d3a3d7d3918f3)
  Generally speaking, packed refs are much faster to read than file-based refs, when present in large numbers.

Fixes #1107

### DIFF
--- a/dulwich/refs.py
+++ b/dulwich/refs.py
@@ -155,6 +155,15 @@ class RefsContainer(object):
         """
         raise NotImplementedError(self.get_packed_refs)
 
+    def add_packed_refs(self, new_refs: Dict[Ref, Optional[ObjectID]]):
+        """Add the given refs as packed refs.
+
+        Args:
+          new_refs: A mapping of ref names to targets; if a target is None that
+            means remove the ref
+        """
+        raise NotImplementedError(self.add_packed_refs)
+
     def get_peeled(self, name):
         """Return the cached peeled value of a ref, if available.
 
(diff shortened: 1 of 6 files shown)


dulwich/refs.py
```diff
@@ -748,7 +795,10 @@ def read_loose_ref(self, name):
                 else:
                     # Read only the first 40 bytes
                     return header + f.read(40 - len(SYMREF))
-        except (FileNotFoundError, IsADirectoryError, NotADirectoryError):
+        except (OSError, UnicodeError):
```
jelmer, line 798: Since everything here is just dealing with bytes, what situations would you expect UnicodeError to be raised here?

jelmer:
LGTM, thanks! One small inline question

dulwich/refs.py
danchr: I just added it out of an abundance of caution — I'm not sure how it could happen in practice. In particular, I'm not sure how this behaves on Windows…

PR #1146, opened 2023-02-14
client: preserve original url during clone

### LINKED ISSUE #1145: fetch fails on a repo cloned using `porcelain.clone()` when using credentials in the url
```python
from dulwich.porcelain import clone, fetch

url = "https://username:password@www.example.com/repository.git"
repo = clone(url, "test")
fetch(repo)
```
```console
dulwich.client.HTTPUnauthorized: No valid credentials provided
```

This happens because the `AbstractHTTPClient.from_parsedurl()` strips the password from the url,
and set `https://username@www.example.com/repository.git` as remote url. Which is why next `fetch` fails.

skshetry:
Fixes #1145

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -2149,12 +2149,6 @@ class AbstractHttpGitClient(GitClient):
         username = parsedurl.username
         if username is not None:
             kwargs["username"] = urlunquote(username)
-        netloc = parsedurl.hostname
-        if parsedurl.port:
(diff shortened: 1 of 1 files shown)


jelmer:
Does this match what C Git does? This was originally done to prevent accidentally leaking credentials.

skshetry:
yes, when you do `git clone https://username:password@server.com`, the whole url is saved to the config.

skshetry:
Thank you @jelmer for a quick review. 🙂

PR #1171, opened 2023-04-27
Support core.symlink=false

### LINKED ISSUE #1169: Allow custom symlink handling on Windows
This is a follow up on #1005.

On poetry, I'm not able to add a git repository containing symlinks (https://github.com/python-poetry/poetry/issues/7798).

When I clone the git repository with the git client, it simply creates text files containing the path of the symlink. There should be a way to get similar behavior in dulwich.

jelmer:
Fixes #1169

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -328,7 +328,14 @@ class BaseRepo:
         """
         raise NotImplementedError(self._determine_file_mode)
 
-    def _init_files(self, bare: bool) -> None:
+    def _determine_symlinks(self) -> bool:
+        """Probe the filesystem to determine whether symlinks can be created.
+
+        Returns: True if symlinks can be created, False otherwise.
+        """
+        raise NotImplementedError(self._determine_symlinks)
+
+    def _init_files(self, *, bare: bool, symlinks: Optional[bool] = None) -> None:
         """Initialize a default set of named files."""
         from .config import ConfigFile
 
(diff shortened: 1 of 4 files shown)


dulwich/repo.py
```diff
@@ -328,7 +328,13 @@ def _determine_file_mode(self) -> bool:
         """
         raise NotImplementedError(self._determine_file_mode)
 
-    def _init_files(self, bare: bool) -> None:
+    def _determine_symlinks(self) -> bool:
```
bruchar1, line 331: The function is defined at line 1766.  Is this stub needed?

jelmer: This is just documentation for any subclasses, which will need to provide an implementation.

dulwich/repo.py
```diff
@@ -1234,6 +1246,14 @@ def _determine_file_mode(self):
 
         return mode_differs and st2_has_exec
 
+    def _determine_symlinks(self):
+        """Probe the filesystem to determine whether symlinks can be created.
+
+        Returns: True if symlinks can be created, False otherwise.
+        """
+        # TODO(jelmer): Actually probe disk / look at filesystem
```
bruchar1, line 1254:
```
with TemporaryDirectory() as tmpdir:
    source = Path(tmpdir) / 'file.txt'
    source.touch()

    link = Path(tmpdir) / 'link.txt'

    try:
        link.symlink_to(source)
        return True
    except OSError:
        return False
```

bruchar1: Good point.

However, on windows, it depends on setted privileges:
> On newer versions of Windows 10, unprivileged accounts can create symlinks if Developer Mode is enabled. When Developer Mode is not available/enabled, the SeCreateSymbolicLinkPrivilege privilege is required, or the process must be run as an administrator.

I'm not sure it is possible to probe without trying to create a real file.

But deactivating symlink on windows unless specifically required is already more that enough. Thank you!

PR #1198, opened 2023-07-24
for use with graph.py new pr

kevinhendricks:
Following the logic of actual git, this pr modifies graph.py to use a priority queue and commit time stamps to greatly speed up "can fast forward" calculations by using commit time to know which parent to walk up first and when to stop the search.

The use of commit time stamps necessitates a tweak to the test_graph.py.

### DIFF
--- a/dulwich/graph.py
+++ b/dulwich/graph.py
@@ -66,7 +66,7 @@ class CommitCache(MutableMapping):
 # priority queue using builtin python minheap tools
 # why they do not have a builtin maxheap is simply ridiculous but
 # liveable with integer time stamps using negation
-class WrkLst(object):
+class WorkList(object):
     def __init__(self):
         self.pq = []
 
@@ -105,7 +105,7 @@ def _find_lcas(lookup_parents, c1, c2s, lookup_stamp, min_stamp=0):
 
     # initialize the working list states with ancestry info
     # note possibility of c1 being one of c2s should be handled
-    wlst = WrkLst()
+    wlst = WorkList()
     cstates[c1] = _ANC_OF_1
     wlst.add((lookup_stamp(c1), c1))
     for c2 in c2s:
--- a/dulwich/graph.py
+++ b/dulwich/graph.py
@@ -20,11 +20,73 @@
 
 """Implementation of merge-base following the approach of git."""
 
-from collections import deque
-from typing import Deque
+from collections.abc import MutableMapping
+from collections import OrderedDict
+
+from heapq import heappush, heappop
+
+# Because opening an sha file from the object_store, reading it,
+# then decoding its will be a slow operation, we want to create
+# a cache of Commit objects so that repeat parent and commit time
+# lookup (which are inherent when looking for a common ancestor)
+# will be much faster, so we need to implement a simple Commit cache
+# using an OrderedDict
 
 
-def _find_lcas(lookup_parents, c1, c2s):
+class CommitCache(MutableMapping):
+
+    def __init__(self, maxsize):
+        self.maxsize = maxsize
+        self.d = OrderedDict()
+
+    def __getitem__(self, key):
+        self.d.move_to_end(key)
+        return self.d[key]
+
+    def __setitem__(self, key, value):
+        if key in self.d:
+            self.d.move_to_end(key)
+        elif len(self.d) == self.maxsize:
+            self.d.popitem(last=False)
+        self.d[key] = value
+
+    def __delitem__(self, key):
+        del self.d[key]
+
+    def __iter__(self):
+        return self.d__iter__()
+
+    def __len__(self):
+        return len(self.d)
+
+    def __contains__(self, key):
+        return key in self.d
+
+
+# priority queue using builtin python minheap tools
+# why they do not have a bultin maxheap is simply ridiculous but
+# liveable with integer time stamps using negation
+class WrkLst(object):
+    def __init__(self):
+        self.pq = []
+
+    def add(self, item):
+        dt, cmt = item
+        heappush(self.pq, (-dt, cmt))
+
+    def get(self):
+        item = heappop(self.pq)
+        if item:
+            pr, cmt = item
+            return -pr, cmt
+        return None
+
+    def iter(self):
+        for (pr, cmt) in self.pq:
+            yield (-pr, cmt)
+
+
+def _find_lcas(lookup_parents, c1, c2s, lookup_stamp, min_stamp=0):
     cands = []
     cstates = {}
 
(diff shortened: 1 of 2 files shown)


dulwich/graph.py
```diff
@@ -20,11 +20,73 @@
...
-def _find_lcas(lookup_parents, c1, c2s):
+# Because opening an sha file from the object_store, reading it,
+# then decoding its will be a slow operation, we want to create
+# a cache of Commit objects so that repeat parent and commit time
+# lookup (which are inherent when looking for a common ancestor)
+# will be much faster, so we need to implement a simple Commit cache
+# using an OrderedDict
+
+
+class CommitCache(MutableMapping):
```
jelmer, line 36: Won't this grow with the size of the graph? In other places we use LRUCache for this.

dulwich/graph.py
```diff
@@ -20,11 +20,73 @@
...
+        return len(self.d)
+
+    def __contains__(self, key):
+        return key in self.d
+
+
+# priority queue using builtin python minheap tools
+# why they do not have a builtin maxheap is simply ridiculous but
+# liveable with integer time stamps using negation
+class WrkLst(object):
```
jelmer, line 69: Please just spellout the name? Is this WorkList?

dulwich/tests/test_graph.py
```diff
@@ -32,9 +32,12 @@ def run_test(dag, inputs):
         def lookup_parents(commit_id):
             return dag[commit_id]
 
+        def lookup_stamp(commit_id):
+            return 100
```
jelmer, line 36: Why 100?

jelmer (changes requested):
Please also add a test for the new behaviour to ensure we don't regress.

dulwich/tests/test_graph.py
kevinhendricks: Your graph as used in tests have no commit time info, just parentage.  Since the more efficient version needs commit time info, it must be supplied by the test itself.  A constant commit time is returned for all points causing the test to be equivalent to its initial form.  Any constant value will work.

dulwich/graph.py
kevinhendricks: It has a maxsize parameter which can be quite small as once the most recent commit works itself back to be nearer in time to the commit time of the older commit, the size of the Working List is quite small.

dulwich/graph.py
kevinhendricks: Yes I will make that change.

dulwich/tests/test_graph.py
kevinhendricks: It acts as the regression test you asked for above as it reverts the test to exactly the behaviour previously tested.

dulwich/graph.py
kevinhendricks: Looked at the LRUCache implementation.

It could be used but is way overkill when a simple OrderedDict is all you need for a finite sized commit cache.  The version in this PR will perform much faster because it internally is using the built-in C implementation of OrderedDict and so does not need separate python based linked lists and their overhead.

But it is your call.  If you want this routine to use your LRUCache instead of being self-contained, faster/simpler, I will make that change.

dulwich/graph.py
```diff
@@ -20,11 +20,73 @@
...
+        return len(self.d)
+
+    def __contains__(self, key):
+        return key in self.d
+
+
+# priority queue using builtin python minheap tools
+# why they do not have a builtin maxheap is simply ridiculous but
+# liveable with integer time stamps using negation
+class WorkList(object):
```
jelmer, line 69: Please add tests for WorkList specifically.

dulwich/tests/test_graph.py
jelmer: please add a comment to that effect

dulwich/graph.py
```diff
@@ -74,17 +136,24 @@ def _has_candidates(wlst, cstates):
...
-    return results
+    for dt, cmt in cands:
+        if not ((cstates[cmt] & _DNC) == _DNC) and not (dt, cmt) in results:
+            results.append((dt, cmt))
+    results.sort(key=lambda x: x[0])
+    lcas = [cmt for dt, cmt in results]
+    return lcas
 
 
+# actual git sorts these based on commit times
```
jelmer, line 156: not sure what this comment is trying to do? IIRC git supports both topological sort and sort based on commit times

dulwich/graph.py
jelmer: It's an existing and tested piece of infrastructure though. Why is an OrderedDict sufficient? What if the breadth is such that parents scroll out of view?

Either way, if you're adding a new class, please add unit tests specifically for that class.

dulwich/graph.py
kevinhendricks: That comment is meant to explain the change to use commit time for this routine.
Git'ss  lca routine the commit time is now used almost exclusively or it is approximated by "adjusted" commit times, but that requires a full tree walk.

dulwich/graph.py
kevinhendricks: Will do.   FYI worklist has existed in some form or other since the beginning and this version uses built-in python minheap.  The test would be to add 3 nodes out of time order and prove they are ordered properly upon retrieval.

dulwich/tests/test_graph.py
kevinhendricks: Will do.

dulwich/graph.py
kevinhendricks: "What if the breadth is such that parents scroll out of view?" ... all fixed size LRU implementations has the same scope problem.  If you want LRUCache and slower that is fine by me.

jelmer: (the criteria for what scrolls out of view is different for this implementation vs the LRUCache though)

Yeah, I'd prefer that - I'd rather be conservative here and use whatever we already have. If it turns out there's a performance win to be had, let's measure implement that separately - and use it everywhere in Dulwich, rather than just here.

kevinhendricks: The "criteria" is to add any new data to the end of the ordered (fifo) dict on writes, and on reads you move the most recently accessed to the just added end.  When cap is reached drop the first (oldest).  In other words an LRU algo with all linked lists and lookups handled internally by python's Ordered dict.  A pretty standard way to implement a LRU cache is with a  Ordered dict but will do as you ask since you will be the one to have to maintain it.

kevinhendricks: fixed in my pr

dulwich/tests/test_graph.py
kevinhendricks: fixed in the updated PR

dulwich/graph.py
kevinhendricks: Test for WorkList added to test_graph.py in the updated PR

jelmer (approved):

PR #1202, opened 2023-08-21
Raise GitProtocolError on unexpected pkt

nanonyme:
There are some packets (like NAK) which have no sha. This does not change them to be handled but you get in exception always the command that failed to be unpacked instead of strange exception about there not being enough values to unpack.


nanonyme:
It's not clear yet whether this is a bug on dulwich or our side on shallow repo handling but in any case dulwich should always report the unexpected packet from remote repo.

jelmer (approved):
NAK shouldn't actually appear in this location. So while this seems fine as a means of improving the error message, it still seems like there's something else going wrong.

PR #1204, opened 2023-08-29
Convert _objects.c to _objects.rs

jelmer:


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=61.2"]
+requires = ["setuptools>=61.2", "setuptools-rust"]
```
Secrus, line 2: Maybe you should consider using [`maturin`](https://github.com/pyo3/maturin)?

jelmer: I don't think I need any of its features for the moment, and would rather avoid the extra (mandatory) dependency.

nanonyme:
If you end up actually integrating this, can you please have Cargo.lock for at least release tags?

jelmer:
> If you end up actually integrating this, can you please have Cargo.lock for at least release tags?

I'm not sure I follow - this branch already includes a Cargo.lock file. Do you mean something else?

nanonyme:
> I'm not sure I follow - this branch already includes a Cargo.lock file. Do you mean something else?

Sorry, didn't spot it from the diff. Looks great to me then.

jelmer:
> > I'm not sure I follow - this branch already includes a Cargo.lock file. Do you mean something else?
> 
> Sorry, didn't spot it from the diff. Looks great to me then.

No worries - was just a bit confused by your comment :)

PR #1214, opened 2023-09-08
fix bug after index merge support rework, add back way to set conflicts

kevinhendricks:
When you reworked the PR to add merge conflict to index.py, you removed the interface to actually set conflicts, and created a bug that will trigger if conflicts ever existed.  Your choice of class for Enum made it hard to effectively cast back to integers as needed to work with conflict levels.

This PR, fixes the bug, bases the enums on IntEnum and adds back the interface to actually set conflicts.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -25,7 +25,7 @@ import stat
 import struct
 import sys
 from dataclasses import dataclass
-from enum import Enum
+from enum import IntEnum
 from typing import (
     Any,
     BinaryIO,
@@ -72,7 +72,7 @@ EXTENDED_FLAG_INTEND_TO_ADD = 0x2000
 DEFAULT_VERSION = 2
 
 
-class Stage(Enum):
+class Stage(IntEnum):
     NORMAL = 0
     MERGE_CONFLICT_ANCESTOR = 1
     MERGE_CONFLICT_THIS = 2
@@ -495,6 +495,27 @@ class Index:
                 return True
         return False
 
+    def set_merge_conflict(self, base_entry, base_time, this_entry, this_time, other_entry, other_time):
+        path, base_index_entry, this_index_entry, other_index_entry = None, None, None, None
+        if this_entry:
+            path = this_entry.path
+            this_index_entry = IndexEntry(this_time, this_time, 0, 0, this_entry.mode,
+                                          0, 0, 0, this_entry.sha)
+        if base_entry:
+            if not path:
+                path = base_entry.path
+            base_index_entry = IndexEntry(base_time, base_time, 0, 0, base_entry.mode,
+                                          0, 0, 0, base_entry.sha)
+        if other_entry:
+            if not path:
+                path = other_entry.path
+            other_index_entry = IndexEntry(other_time, other_time, 0, 0, other_entry.mode,
+                                           0, 0, 0, other_entry.sha)
+        if path:
+            self._byname[path] = ConflictedIndexEntry(ancestor=base_index_entry,
+                                                      this=this_index_entry,
+                                                      other=other_index_entry)
+
     def clear(self):
         """Remove all contents from this index."""
         self._byname = {}
@@ -533,6 +554,8 @@ class Index:
 
         def lookup_entry(path):
             entry = self[path]
+            if isinstance(entry, ConflictedIndexEntry):
+                entry = entry.this
             return entry.sha, cleanup_mode(entry.mode)
 
         yield from changes_from_tree(


dulwich/index.py
```diff
@@ -533,6 +554,8 @@ def changes_from_tree(
 
         def lookup_entry(path):
             entry = self[path]
+            if isinstance(entry, ConflictedIndexEntry):
```
jelmer, line 557: I think this should raise an exception rather than silently defaulting to .this (which may not actually be set)

dulwich/index.py
```diff
@@ -495,6 +495,27 @@ def has_conflicts(self) -> bool:
                 return True
         return False
 
+    def set_merge_conflict(self, base_entry, base_time, this_entry, this_time, other_entry, other_time):
```
jelmer, line 498: please add types

dulwich/index.py
jelmer: please also add a test for this behaviour

dulwich/index.py
jelmer: why does this take a separate time argument? Couldn't the caller just set that if they needed to?

jelmer: please add tests for this new method

dulwich/index.py
```diff
@@ -495,6 +495,27 @@ def has_conflicts(self) -> bool:
...
+        path, base_index_entry, this_index_entry, other_index_entry = None, None, None, None
+        if this_entry:
+            path = this_entry.path
```
jelmer, line 501: It seems a bit odd to derive the path from one of the entries passed in. What if the entries are for different paths? Perhaps pass in the path and then verify that each of the entries matches the path?

dulwich/index.py
```diff
@@ -495,6 +495,27 @@ def has_conflicts(self) -> bool:
...
+                path = base_entry.path
+            base_index_entry = IndexEntry(base_time, base_time, 0, 0, base_entry.mode,
+                                          0, 0, 0, base_entry.sha)
+        if other_entry:
+            if not path:
+                path = other_entry.path
+            other_index_entry = IndexEntry(other_time, other_time, 0, 0, other_entry.mode,
+                                           0, 0, 0, other_entry.sha)
+        if path:
+            self._byname[path] = ConflictedIndexEntry(ancestor=base_index_entry,
```
jelmer, line 515:
```suggestion
            self[path] = ConflictedIndexEntry(ancestor=base_index_entry,
```

jelmer (changes requested):

dulwich/index.py
jelmer: Also, IndexEntry doesn't have a path associated with it - just the internal type, SerializedIndexEntry

dulwich/index.py
kevinhendricks: Okay.

FWIW, all of these changes were part of my earlier PR that you accepted and then reworked completely.

I will work on the requested test additions.

dulwich/index.py
kevinhendricks: It is deriving the path because that is how you stored it in the current index dict (accessed by path).  If path and stage were used as I had originally submitted in the first PR then you would not have this issue as multiple paths and stages would allowed.  Either way this index is for the current brach/commit/tree and not for "other" or "base" trees.

jelmer: IndexEntry never had a path attribute, not even in your PR as far as I can tell.

dulwich/index.py
jelmer: I intentionally changed the code back to preserve the original API that uses names as keys,  requiring people to explicitly specify which stage they're wanting to deal with for conflicted entries.

dulwich/index.py
kevinhendricks: Huh?  The name was the path and that had to be unique to access any index entry?  I guessI am not sure what you are asking then?  To store apan index entry in the index dict you need a name which is the path.  So a path must exist.  That path should be based on the 'this' currently checked out but in merge conflicts, that file may no longer exist (think of case where exists in base tree, exists and is modified in "other" tree, but removed in "this" tree forexample). In this case to report the conflicted entry, the path is extracted from other in order to have a unique name to store in index dict. 

So exactly what is the issue that you want me to address?

jelmer: My point was that this particular line is "this_entry.path", but IndexEntry does not have a path attribute. If that doesn't make sense feel free to ignore it - it should come up as you write the unit tests for this function.

dulwich/index.py
kevinhendricks: That  could also be done in the original PR you accepted, and then you changed it almost completely to the point of even dropping the interface needed to set a conflict!  Your changes created all of these issues.  I am just trying to get back to the functionality it had when you first accepted it.

jelmer: The original PR had an ambiguous interface (allowing both names and (name, stage) tuples). It implicitly defaulted to "this" for anything that was conflicted.

kevinhendricks: How on earth is allowing a key that is a path or tuple or any more ambiguous than allowing a value to be either of two very different classes?   And then breaking things and dropping the interface?

After 40+ years of being a paid and opensource developer who has made contributions to many many projects, I can see this is just not going to work.  You place more hurdles to getting working code going than anyone I have ever met.  Please close this.

Sigil will fork and go our own way.

jelmer: Sorry :-( That was not my intent; I made that change specifically so that at least the PR itself could land, and perhaps that was the wrong call.

Maybe to give a expand on that justification: I feel that Index with current implementation is more pythonic. You could replace "Index" with "dict", and it would mostly work. It's not uncommon for different types to be used for the values of a dict, but seems far more surprising that changing one key deletes another.

Issue #868, opened 2021-05-18
win32api.FindExecutable('git') not working?


ngbrown:
Hello, I'm using dulwich through hg-git, and got the following error:

```
...
  File "C:\Python37x64\lib\site-packages\dulwich\client.py", line 798, in send_pack
    proto, unused_can_read, stderr = self._connect(b'receive-pack', path)
  File "C:\Python37x64\lib\site-packages\dulwich\client.py", line 1080, in _connect
    git_command = find_git_command()
  File "C:\Python37x64\lib\site-packages\dulwich\client.py", line 1059, in find_git_command
    status, git = win32api.FindExecutable('git')
pywintypes.error: FindExecutable: There is no association for the file

D:\dev\H-D\VTS\hg-to-git-project\vts-protocolbuffers-hg>where git
C:\Program Files\Git\cmd\git.exe
```

'git' is on the path:

```
D:\>where git
C:\Program Files\Git\cmd\git.exe
```

The behavior does seem to be different between Python 2.7 and 3.7:

```
C:\>c:\Python27\python.exe
Python 2.7.12 (v2.7.12:d33e0cf91556, Jun 27 2016, 15:24:40) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import win32api
>>> win32api.FindExecutable('git')
(42L, 'C:\\Program Files\\Git\\cmd\\git.exe')
```

and

```
C:\>python
Python 3.7.3 (v3.7.3:ef4ec6ed12, Mar 25 2019, 22:22:05) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import win32api
>>> win32api.FindExecutable('git')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pywintypes.error: FindExecutable: There is no association for the file
>>> win32api.FindExecutable('git.exe')
(42, 'C:\\Program Files\\Git\\cmd\\git.exe')
```

I changed:

https://github.com/dulwich/dulwich/blob/c6993451a745d313f61e5d080b9f9d8611a8d7f4/dulwich/client.py#L1059

to

```py
            status, git = win32api.FindExecutable('git.exe')
```

which worked.

I tried both the latest (v0.20.21) and the one I modified (v0.19.16).

Maybe catch the thrown error and fall back to the `cmd /c git` option or try other extensions?

jelmer:
[...]

PR #1215, opened 2023-09-20
Issue #868 - Add fallback if win32api.FindExecutable cannot find git

mbellgardt:
See issue #868. Since OP did not react, I opened this requested PR.


jelmer (approved):
thanks!

PR #1234, opened 2023-11-21
Update cibuildwheel in CI workflow

### LINKED ISSUE #1233: No wheel built by CI
All is in the title, the fix would be to replace cibuildwheel==2.9.0 by cibuildwheel==2.16.2.
I tested the build with this. You could look the action result here https://github.com/Starmania/dulwich/actions/runs/6930985984

Starmania:
Upgrade cibuildwheel so python could build wheel

Close #1233


jelmer (approved):

jelmer:
Thanks!

PR #1235, opened 2023-11-24
repo: add pattern argument to get_refs

dtrifiro:
This matches `git for-each-ref <pattern>` behaviour, where `<pattern>` is a `glob (7)` style pattern.


jelmer (changes requested):
Could you perhaps move this into dulwich.porcelain?

dtrifiro:
@jelmer thanks for reviewing this. I moved the code to `porcelain.for_each_ref` and also added a corresponding cli entrypoint `dulwich.cli:for-each-ref()`

PR #1244, opened 2023-12-12
add initial support for Git protocol v2

stspdotname:
Implement Git protocol version negotiation and use Git protocol v2 for fetches if supported. For now, the observable behaviour of Dulwich is equivalent regardless of protocol version, except that two new features may be used if the server supports Git protocol v2.

The first feature is a reference prefix list which may be used to filter refs server-side. This can be used to reduce the size of the initial reference listing announced by the server. Reducing the size of this announcement was a major design goal for protocol v2 to avoid sending of very large announcements when a repository contains a lot of references. This feature is intended as an optimization which servers are free to ignore depending on available server-side resources. Therefore, users of Dulwich should still be prepared to filter redundant refs manually (this limitation also applies to Git itself).
A new --refspec porcelain option is provided in order to test this feature on the command line.

The second feature is an object filter specification, which corresponds to the --filter option of 'git clone'. This can be used to omit objects while cloning repositories. For instance, the following command will clone a given repsitory without fetching any blob objects:

   dulwich clone --filter blob:none --bare REPO_URL

(In this example the --bare option is used because creation of a work tree would fail without any blobs present.)

The test suite now enables protocol v2 and keeps passing for me.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -486,6 +486,26 @@ def init(path=".", *, bare=False, symlinks: Optional[bool] = None):
         return Repo.init(path, symlinks=symlinks)
 
 
+def encode_refspecs(refspecs, refspec_encoding):
+    if refspecs is None:
+        return [b"HEAD"]
+
+    def encode_refspec(ref):
+        if isinstance(ref, bytes):
+            return ref
+        else:
+            return ref.encode(refspec_encoding)
+
+    encoded_refs = []
+    if isinstance(refspecs, bytes) or isinstance(refspecs, str):
+        encoded_refs.append(encode_refspec(refspecs))
+    else:
+        for ref in refspecs:
+            encoded_refs.append(encode_refspec(ref))
+
+    return encoded_refs
+
+
 def clone(
     source,
     target=None,
@@ -497,6 +517,9 @@ def clone(
     depth: Optional[int] = None,
     branch: Optional[Union[str, bytes]] = None,
     config: Optional[Config] = None,
+    refspecs=None,
+    refspec_encoding=DEFAULT_ENCODING,
+    filter_spec=None,
     **kwargs,
 ):
     """Clone a local or remote git repository.
@@ -513,6 +536,13 @@ def clone(
       branch: Optional branch or tag to be used as HEAD in the new repository
         instead of the cloned repository's HEAD.
       config: Configuration to use
+      refspecs: refspecs to fetch. Can be a bytestring, a string, or a list of
+        bytestring/string.
+      refspec_encoding: Character encoding of bytestrings provided in the refspecs parameter.
+        If not specified, the internal default encoding will be used.
+      filter_spec: A git-rev-list-style object filter spec, as an ASCII string.
+        Only used if the server supports the Git protocol-v2 'filter'
+        feature, and ignored otherwise.
     Returns: The new repository
     """
     if outstream is not None:
@@ -533,6 +563,8 @@ def clone(
     if checkout and bare:
         raise Error("checkout and bare are incompatible")
 
+    encoded_refs = encode_refspecs(refspecs, refspec_encoding)
+
     if target is None:
         target = source.split("/")[-1]
 
@@ -542,6 +574,9 @@ def clone(
     mkdir = not os.path.exists(target)
 
     (client, path) = get_transport_and_path(source, config=config, **kwargs)
+
+    if filter_spec:
+        filter_spec = filter_spec.encode("ascii")
 
     return client.clone(
         path,
@@ -553,6 +588,8 @@ def clone(
         branch=branch,
         progress=errstream.write,
         depth=depth,
+        ref_prefix=encoded_refs,
+        filter_spec=filter_spec,
     )
 
 
@@ -1238,6 +1275,8 @@ def pull(
     errstream=default_bytes_err_stream,
     fast_forward=True,
     force=False,
+    refspec_encoding=DEFAULT_ENCODING,
+    filter_spec=None,
     **kwargs,
 ):
     """Pull from remote via dulwich.client.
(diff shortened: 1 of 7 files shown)


stspdotname:
Not sure what the "unexpected success" failures on Windows are about. Are the more detailed test logs available? I don't have a Windows machine to run tests on myself.

dulwich/porcelain.py
```diff
@@ -543,6 +573,9 @@ def clone(
 
     (client, path) = get_transport_and_path(source, config=config, **kwargs)
 
+    if filter_spec:
+        filter_spec = filter_spec.encode('ascii'),
```
jelmer, line 577: is the trailing comma here intentional?

jelmer (changes requested):
thanks! LGTM overall; I left some nits inline, but nothing major.

It would be good to have at least some basic tests; maybe there's a way of forcing the git server to force a particular protocol version and rely on the compat tests?

jelmer:
> Not sure what the "unexpected success" failures on Windows are about. Are there more detailed test logs available? I don't have a Windows machine to run tests on myself.

These are spurious - I've been meaning to look into them and possibly mark them as flapping.

dulwich/porcelain.py
```diff
@@ -1239,28 +1274,35 @@ def pull(
...
       repo: Path to repository
       remote_location: Location of the remote
-      refspecs: refspecs to fetch
+      refspecs: refspecs to fetch. Can be a bytestring, a string, or a list of
+        bytestring/string.
       outstream: A stream file to write to output
       errstream: A stream file to write to errors
+      refspec_encoding: Character encoding of bytestrings provided in the refspecs parameter.
+        If not specified, the internal default encoding will be used.
+      filter_spec: A git-rev-list-style object filter spec, as an ASCII string.
```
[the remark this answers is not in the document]
stspdotname: > What does C Git do if --filter is specified and the remote server doesn't support filtering?

The client has no way to tell whether the server will do any filtering. The first message sent by the server already includes the list of references, and the decision whether to filter this list happens server-side.

This parameter is more of a hint for the server, to avoid unnecessary work when the client doesn't need the full list of references. However, the Git docs explicitly state that the server is free to ignore this parameter and just provide all the refs anyway.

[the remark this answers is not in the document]
jelmer: ack, thanks!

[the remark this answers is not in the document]
stspdotname: Actually you may have been asking about the object filter rather than the reference filter?

A lack of the object filtering capability could be treated as an error if the user has requested object filtering.

However, Git does not treat this as a fatal error, it prints a warning and proceeds with a full clone:

Cloning into bare repository 'test-clone'...                                                   
warning: filtering not recognized by server, ignoring

I am not sure if dulwich is supposed to print such a warning?

I suppose we could provide some way for API users to detect the lack of server-side object filtering, e..g. via a callback, and let them decide whether the transfer should be aborted early. I would keep such an enhancement for later though. For now, matching Git's fallback behaviour seems good enough to me.

stspdotname:
> It would be good to have at least some basic tests; maybe there's a way of forcing the git server to force a particular protocol version and rely on the compat tests?

Reading Git's server code suggests that the server always uses the highest mutually supported version . See determine_protocol_version_server() in Git's protocol.c. 

It doesn't make much sense to run dulwich compat tests against ancient versions of Git so in practice the protocol version used during dulwich tests depends entirely on what the client is requesting.

Therefore we should add a configuration knob to dulwich that controls the protocol version the client will request. Git clients already support such an option: protocol.version in git-config.

I will try to add such an option to dulwich as a command-line parameter, and/or environment variable, and/or via the existing git config option -- any preference?

Once this parameter exists, ideally the entire compat test suite would always run twice, once with the old protocol (v0/v1) and again with protocol v2. Any tests which only make sense in one case could be skipped in the other.

stspdotname:
I am not quite sure yet how to adjust the test suite to run compat tests across both protocol versions. I will look into this next. Meanwhile, feel free to merge this PR if you are happy with the changes so far. I can always open a new PR for the test suite changes later.

tests/__init__.py
```diff
@@ -213,11 +214,21 @@ def nocompat_test_suite():
     return result
 
 
+@patch("dulwich.protocol.DEFAULT_GIT_PROTOCOL_VERSION_FETCH", new=0)
+def compat_test_suite_git_protocol_v0():
```
jelmer, line 218: Is this running all compat tests twice, not just the client ones?

This should really just be specific to the client code.

dulwich/protocol.py
```diff
@@ -30,6 +30,10 @@
 
 TCP_GIT_PORT = 9418
 
+GIT_PROTOCOL_VERSIONS = [0, 1, 2]
```
jelmer, line 33: Is there actually a version 1? Can you add a comment about these versions.

tests/compat/test_client.py
```diff
@@ -429,16 +433,22 @@ def run_command(
         port=None,
         password=None,
         key_filename=None,
+        protocol_version=None,
     ):
         cmd, path = command.split(" ")
         cmd = cmd.split("-", 1)
         path = path.replace("'", "")
+        env = copy.deepcopy(os.environ)
+        if protocol_version is None or protocol_version == 2:
```
jelmer, line 442: Can we do the inverse here, i.e. 

If protocol version is not set, set it to 2.

If protocol version != 0 then set this header.

That prepares us a bit better to future versions.

tests/compat/test_client.py
```diff
@@ -429,16 +433,22 @@ def run_command(
         port=None,
         password=None,
         key_filename=None,
+        protocol_version=None,
     ):
         cmd, path = command.split(" ")
         cmd = cmd.split("-", 1)
         path = path.replace("'", "")
+        env = copy.deepcopy(os.environ)
```
jelmer, line 441: Does this need a deep copy rather than just dict(os.environ)?

tests/__init__.py
```diff
@@ -213,11 +214,21 @@ def nocompat_test_suite():
     return result
 
 
+@patch("dulwich.protocol.DEFAULT_GIT_PROTOCOL_VERSION_FETCH", new=0)
+def compat_test_suite_client_git_protocol_v0():
```
jelmer, line 218: This should all be down in compat testsuite, rather than complicating things here.

tests/__init__.py
```diff
@@ -213,11 +214,21 @@ def nocompat_test_suite():
     return result
 
 
+@patch("dulwich.protocol.DEFAULT_GIT_PROTOCOL_VERSION_FETCH", new=0)
```
jelmer, line 217: I guess this works and is okay for now but ideally this would just be a parameter on the test classes.

jelmer (changes requested):

dulwich/protocol.py
stspdotname: Yes, I am adding comments in the next version.

They will appear in an earlier commit which adds this list.
I have rewritten history a bit, also to get rid of the 'ruff format' commit in the middle.

tests/compat/test_client.py
stspdotname: Yes, we can do that. I don't anticipate a new protocol version in the foreseeable future, but it doesn't hurt to handle things that way.

tests/compat/test_client.py
stspdotname: It works with just dict(), indeed.  I don't recall specifics for why I made it use deepcopy() . I recall seeing some problems with existing environment being modified at some point somehow.

The HTTP server tests also use a deepcopy() function. Perhaps I just copied from there.

tests/__init__.py
stspdotname: I have found a better way to do this. We can subclass all relevant client test classes and mock the default protocol  value on those subclasses. This achieves the same with much less code (and for some reason the code in the version you reviewed didn't even work correctly anymore -- all test runs were using protocol version 2).

tests/__init__.py
stspdotname: Please check the new version, it should be better.

jelmer (changes requested):
Please take a look at the test failures.

jelmer (approved):

jelmer (approved):

Issue #1225, opened 2023-10-14
Dulwich passes through IncompleteRead from urllib3


nanonyme:
We hit this in CI, most likely due to network issues
```
          File "/usr/lib/python3.11/site-packages/dulwich/client.py", line 816, in fetch
            result = self.fetch_pack(
                     ^^^^^^^^^^^^^^^^
          File "/usr/lib/python3.11/site-packages/dulwich/client.py", line 2146, in fetch_pack
            _handle_upload_pack_tail(
          File "/usr/lib/python3.11/site-packages/dulwich/client.py", line 606, in _handle_upload_pack_tail
            for chan, data in _read_side_band64k_data(proto.read_pkt_seq()):
          File "/usr/lib/python3.11/site-packages/dulwich/client.py", line 497, in _read_side_band64k_data
            for pkt in pkt_seq:
          File "/usr/lib/python3.11/site-packages/dulwich/protocol.py", line 275, in read_pkt_seq
            pkt = self.read_pkt_line()
                  ^^^^^^^^^^^^^^^^^^^^
          File "/usr/lib/python3.11/site-packages/dulwich/protocol.py", line 223, in read_pkt_line
            pkt_contents = read(size - 4)
                           ^^^^^^^^^^^^^^
          File "/usr/lib/python3.11/site-packages/urllib3/response.py", line 879, in read
            data = self._raw_read(amt)
                   ^^^^^^^^^^^^^^^^^^^
          File "/usr/lib/python3.11/site-packages/urllib3/response.py", line 813, in _raw_read
            with self._error_catcher():
          File "/usr/lib/python3.11/contextlib.py", line 155, in __exit__
            self.gen.throw(typ, value, traceback)
          File "/usr/lib/python3.11/site-packages/urllib3/response.py", line 727, in _error_catcher
            raise ProtocolError(f"Connection broken: {e!r}", e) from e
        urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(7577 bytes read)', IncompleteRead(7577 bytes read))
```
We would expect this to result in GitProtocolError.

jelmer:
Happy to merge PRs that wrap urllib3.exceptions.ProtocolError in a GitProtocolError.

nanonyme:
I now finally looked into this. I'm not sure if I'm understanding how the git client dependency graph works but I will submit a PR to hopefully get this forward.

PR #1261, opened 2024-03-05
Wrap all urllib3 read ProtocolErrors as GitProtocolError

nanonyme:
Fixes https://github.com/jelmer/dulwich/issues/1225

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -2229,6 +2229,17 @@ class AbstractHttpGitClient(GitClient):
         return f"{type(self).__name__}({self._base_url!r}, dumb={self.dumb!r})"
 
 
+def _wrap_urllib3_exceptions(func):
+
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except urllib3.exceptions.ProtocolError as error:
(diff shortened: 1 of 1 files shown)


nanonyme:
@jelmer not sure if this is overkill. It seemed simpler to just add an exception to protocol.py but that would have leaked HTTP client implementation details there (it looked like this has been insulated into client.py).

dulwich/client.py
```diff
@@ -2229,6 +2229,17 @@ def __repr__(self) -> str:
         return f"{type(self).__name__}({self._base_url!r}, dumb={self.dumb!r})"
 
 
+def wrap_exceptions(func):
```
jelmer, line 2232: This should probably be private and include urllib3 in the name.

jelmer (changes requested):
Looks good to me overall, just one nit about the function name.

nanonyme:
Looks like CI is failing all over due to ruff

jelmer (approved):

nanonyme:
Is there any plan for next release?

jelmer:
Not currently but I can probably do one in the next couple of weeks

Issue #1285, opened 2024-04-22
worktreeConfig capitalization problem


wshanks:
dulwich only tests for `extensions.worktreeconfig` but in the [git documentation](https://git-scm.com/docs/git-worktree#_configuration_file) it is written as `extensions.worktreeConfig`. Recently, we started seeing [dulwich error in GitHub Actions](https://github.com/Qiskit-Extensions/qiskit-experiments/actions/runs/8786231049/job/24108610355#step:6:152) with `dulwich.repo.UnsupportedExtension: b'worktreeConfig'`. I tested locally and confirmed that `extensions.worktreeconfig` does not lead to an error from dulwich but `extensions.worktreeConfig` does.

I am not sure what is going on with GitHub Actions. I see that someone was affected by this in https://github.com/jelmer/dulwich/pull/1262 but then closed it as "fixed by git". Also, some other users have been having the opposite problem with GitHub Actions it seems: https://github.com/GitTools/actions/issues/1115. It seems like this setting is not consistent across GitHub Actions runs recently.

0x2b3bfa0:
> _[The variable names are case-insensitive, allow only alphanumeric characters and `-`, and must start with an alphabetic character.](https://www.git-scm.com/docs/git-config/2.14.6#_configuration_file)_

wshanks:
After some more testing, I found that git itself is insensitive to the capitalization of config entries. So I think the issue is that dulwich is treating the config entry as it is in the file rather without lower casing it first and that #1262 would be the right solution.

jelmer:
Couli you create a new PR, ideally with a unit test that reproduces the issue?

shcheklein:
I think the easiest, less intrusive fix is to change this:


```python
        for extension, _value in config.items((b"extensions",)):
            if extension not in (b"worktreeconfig",):
                raise UnsupportedExtension(extension)
```

to something like:

```python
        for extension, _value in config.items((b"extensions",)):
            if extension.lower() not in (b"worktreeconfig",):
                raise UnsupportedExtension(extension)
```

`dulwich` is actually using `CaseInsensitiveOrderedMultiDict` internally, but `items` is iterating on an actual (real) keys.

It would be great to see if `items()` can be changed to return a nomalized list, but most likely there was a good reason for this.

0x2b3bfa0:
> _I think the easiest, less intrusive fix is to change this_ [...] _to something like_ [...]

[...]

PR #1286, opened 2024-04-22
Handle alternate case for worktreeconfig setting

wshanks:
`extensions.worktreeconfig` was checked in a way that was not case-insensitive, but git config settings should be case insensitive. This behavior led to problems since the documentation describes the setting as `worktreeConfig` and GitHub Actions started setting the option with this casing.

Fixes https://github.com/jelmer/dulwich/issues/1285

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -1143,7 +1143,7 @@ class BuildRepoRootTests(TestCase):
         self.assertEqual(cs.get(("user",), "name"), b"Jelmer")
 
     def test_worktreeconfig_extension_case(self):
-        """Test that worktree code does not error for alternate case format"""
+        """Test that worktree code does not error for alternate case format."""
         r = self._repo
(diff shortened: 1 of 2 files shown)


jelmer (approved):

jelmer:
Please take a look at the test failure.

tests/test_repository.py
```diff
@@ -1142,6 +1142,22 @@ def test_worktreeconfig_extension(self):
         cs = r.get_config_stack()
         self.assertEqual(cs.get(("user",), "name"), b"Jelmer")
 
+    def test_worktreeconfig_extension_case(self):
+        """Test that worktree code does not error for alternate case format"""
```
wshanks, line 1146:
```suggestion
        """Test that worktree code does not error for alternate case format."""
```

jelmer (approved):

PR #1288, opened 2024-04-23
Lowercase `CaseInsensitiveOrderedMultiDict.items()` keys

0x2b3bfa0:
> _The variable names are case-insensitive._ [...] _Section names are case-insensitive._ ([Git documentation](https://www.git-scm.com/docs/git-config/2.14.6#_configuration_file))

Could close https://github.com/jelmer/dulwich/issues/1285.

### DIFF
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -89,7 +89,7 @@ class CaseInsensitiveOrderedMultiDict(MutableMapping):
         return self._keyed.keys()
 
     def items(self):
-        return iter(self._real)
+        return ((lower_key(name), value) for name, value in self._real)
 
     def __iter__(self):
         return self._keyed.__iter__()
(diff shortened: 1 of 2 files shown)


jelmer (changes requested):
The git config file is indeed case insensitive for lookups but also case preserving. This change disregards the case altogether.

0x2b3bfa0:
> _This change disregards the case altogether._

Certainly, but please note that the `keys` and `__iter__` methods of `CaseInsensitiveOrderedMultiDict` already disregard the case, and `items` is the only method still preserving it.

jelmer:
> > _This change disregards the case altogether._
> 
> Note that the `keys` and `__iter__` methods of `CaseInsensitiveOrderedMultiDict` were already converting keys to lowercase, and `items` was the only method still preserving the original case.

I'd argue that that is a bug in those methods, rather than a bug in .items().

0x2b3bfa0:
Closing as per https://github.com/jelmer/dulwich/pull/1288#issuecomment-2071953393 & #1286

PR #1289, opened 2024-04-23
Make `CaseInsensitiveOrderedMultiDict.items()` keys case insensitive

### LINKED ISSUE #1285: worktreeConfig capitalization problem

0x2b3bfa0:
Variant of https://github.com/jelmer/dulwich/issues/1288 addressing https://github.com/jelmer/dulwich/pull/1288#pullrequestreview-2016417146.

Could close #1285.

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -1142,6 +1142,22 @@ class BuildRepoRootTests(TestCase):
         cs = r.get_config_stack()
         self.assertEqual(cs.get(("user",), "name"), b"Jelmer")
 
+def test_worktreeconfig_extension_case(self):
+        """Test that worktree code does not error for alternate case format."""
+        r = self._repo
+        c = r.get_config()
+        c.set(("core",), "repositoryformatversion", "1")
+        # Capitalize "Config"
+        c.set(("extensions",), "worktreeConfig", True)
+        c.write_to_path()
+        c = r.get_worktree_config()
+        c.set(("user",), "repositoryformatversion", "1")
+        c.set((b"user",), b"name", b"Jelmer")
+        c.write_to_path()
+        # The following line errored before
+        # https://github.com/jelmer/dulwich/issues/1285 was addressed
+        Repo(self._repo_dir)
+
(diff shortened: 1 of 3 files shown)


tests/test_repository.py
```diff
@@ -1142,6 +1142,22 @@ def test_worktreeconfig_extension(self):
...
+        # Capitalize "Config"
+        c.set(("extensions",), "worktreeConfig", True)
+        c.write_to_path()
+        c = r.get_worktree_config()
+        c.set(("user",), "repositoryformatversion", "1")
+        c.set((b"user",), b"name", b"Jelmer")
+        c.write_to_path()
+        # The following line errored before
+        # https://github.com/jelmer/dulwich/issues/1285 was addressed
+        Repo(self._repo_dir)
```
0x2b3bfa0, line 1159: Copied from #1286

jelmer:
Thanks for taking the time to create a PR. 

Wrapping str is a bit too much magic for my taste, it creates an extra class and instance for each yielded item. This PR also doesn't implement some other methods from str such as __lt__ and __le__ which could lead to other issues further down the line.

Issue #1279, opened 2024-04-21
test_init_connector fails with geventhttpclient>=2.2.1


mweinelt:
The `geventhttpclient.HTTPClient.request.get_base_url` function has been removed in https://github.com/geventhttpclient/geventhttpclient/commit/d673ea93d5fd55ea4b8b8f62b6caa3e730b80403.

This breaks `TestSwiftConnector.test_init_connector`.

```
dulwich> ____________________ TestSwiftConnector.test_init_connector ____________________
dulwich> [gw2] linux -- Python 3.11.9 /nix/store/lpi16513bai8kg2bd841745vzk72475x-python3-3.11.9/bin/python3.11
dulwich> 
dulwich> self = <dulwich.contrib.test_swift.TestSwiftConnector testMethod=test_init_connector>
dulwich> 
dulwich>     def test_init_connector(self):
dulwich>         self.assertEqual(self.conn.auth_ver, "1")
dulwich>         self.assertEqual(self.conn.auth_url, "http://127.0.0.1:8080/auth/v1.0")
dulwich>         self.assertEqual(self.conn.user, "test:tester")
dulwich>         self.assertEqual(self.conn.password, "testing")
dulwich>         self.assertEqual(self.conn.root, "fakerepo")
dulwich>         self.assertEqual(
dulwich>             self.conn.storage_url, "http://127.0.0.1:8080/v1.0/AUTH_fakeuser"
dulwich>         )
dulwich>         self.assertEqual(self.conn.token, "12" * 10)
dulwich>         self.assertEqual(self.conn.http_timeout, 1)
dulwich>         self.assertEqual(self.conn.http_pool_length, 1)
dulwich>         self.assertEqual(self.conn.concurrency, 1)
dulwich>         self.conf.set("swift", "auth_ver", "2")
dulwich>         self.conf.set("swift", "auth_url", "http://127.0.0.1:8080/auth/v2.0")
dulwich>         with patch("geventhttpclient.HTTPClient.request", fake_auth_request_v2):
dulwich>             conn = swift.SwiftConnector("fakerepo", conf=self.conf)
dulwich>         self.assertEqual(conn.user, "tester")
dulwich>         self.assertEqual(conn.tenant, "test")
dulwich>         self.conf.set("swift", "auth_ver", "1")
dulwich>         self.conf.set("swift", "auth_url", "http://127.0.0.1:8080/auth/v1.0")
[...]

PR #1299, opened 2024-04-28
Keep up with changes in geventhttpclient

PerchunPak:
Fixes https://github.com/jelmer/dulwich/issues/1279

### DIFF
--- a/dulwich/contrib/swift.py
+++ b/dulwich/contrib/swift.py
@@ -257,7 +257,7 @@ class SwiftConnector:
         if ret.status_code < 200 or ret.status_code >= 300:
             raise SwiftException(
                 "AUTH v1.0 request failed on "
-                + f"{str(auth_httpclient.get_base_url()) + path} with error code {ret.status_code} ({ret.items()!s})"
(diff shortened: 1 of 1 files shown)


dulwich/contrib/swift.py
```diff
@@ -257,7 +257,7 @@ def swift_auth_v1(self):
         if ret.status_code < 200 or ret.status_code >= 300:
             raise SwiftException(
                 "AUTH v1.0 request failed on "
-                + f"{str(auth_httpclient.get_base_url()) + path} with error code {ret.status_code} ({ret.items()!s})"
+                + f"{self.auth_url} with error code {ret.status_code} ({ret.items()!s})"
```
PerchunPak, line 260: I am not sure if this information is safe to print, I am also unsure why the author of this code used `auth_httpclient.get_base_url` instead of `self.auth_url` and if they are equal

jelmer (approved):

dulwich/contrib/swift.py
jelmer: Let's optimistically assume this is okay. If we're wrong, we can revert and add a comment (and add a comment)

jelmer (approved):

PR #1313, opened 2024-05-14
Reformat Source Code & Add `ruff format --check` to the Makefile

DaveLak:
This is a follow-up to PR #1312 which missed a few files. This PR:

- Reformats the remaining unformatted files with `ruff format`.
- Adds `ruff format --check .` to code style check in the Makefile.
- Updates the `fix` and `reformat` Makefile targets to use `$(RUFF)` instead of `ruff` directly, making it consistent with the `style` make target, and the other targets in
the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PYTHON = python3
-RUFF ?= $(PYTHON) -m ruff 
+RUFF ?= $(PYTHON) -m ruff
 SETUP = $(PYTHON) setup.py
 TESTRUNNER ?= unittest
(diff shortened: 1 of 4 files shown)


DaveLak:
Hmm, never mind. It looks like CI is using `ruff 0.4.4` but I had `ruff 0.2.2` locally.

With `ruff 0.4.4`, the reformatting I proposed here is unnecessary and results in `python3 -m ruff check .` failing.

DaveLak:
I created issue #1314 with details related to the misunderstanding that led me to open this PR.

Issue #1352, opened 2024-08-06
Support Python 3.13


edgarrmondragon:
Python 3.13.0rc1 is out[^1] so it'd be nice to start publishing the corresponding wheels for this package.

[^1]: https://discuss.python.org/t/python-3-13-0-release-candidate-1-released/59703

jelmer:
It should just be a matter of updating the files in .github/workflows/ to include 3.13. can you submit a PR?

edgarrmondragon:
> can you submit a PR?

Yes, will do!

jelmer:
Reopening this since the actual tests are failing. This appears to be blocked on https://github.com/PyO3/pyo3/pull/4420

jelmer:
We're also hitting https://github.com/python/cpython/issues/122728:

```
ERROR: test_default_key (tests.test_porcelain.CommitSignTests.test_default_key)
----------------------------------------------------------------------
SystemError: Objects/dictobject.c:3774: bad argument to internal function

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/runner/work/dulwich/dulwich/tests/test_porcelain.py", line 469, in test_default_key
    sha = porcelain.commit(
        self.repo.path,
    ...<3 lines>...
        signoff=True,
    )
  File "/home/runner/work/dulwich/dulwich/dulwich/porcelain.py", line 444, in commit
    return r.do_commit(
           ~~~~~~~~~~~^
        message=message,
        ^^^^^^^^^^^^^^^^
    ...<6 lines>...
        sign=signoff if isinstance(signoff, (str, bool)) else None,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/runner/work/dulwich/dulwich/dulwich/repo.py", line 1045, in do_commit
    c.sign(keyid)
    ~~~~~~^^^^^^^
  File "/home/runner/work/dulwich/dulwich/dulwich/objects.py", line 1508, in sign
    self.gpgsig, unused_result = c.sign(
                                 ~~~~~~^
        self.as_raw_string(), mode=gpg.constants.sig.mode.DETACH
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #1353, opened 2024-08-06
Support Python 3.13

edgarrmondragon:
Closes https://github.com/jelmer/dulwich/issues/1352

### DIFF
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version:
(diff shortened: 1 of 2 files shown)


jelmer:
You'll also need to update python-distributions.yml - that's where the actual wheels are built and uploaded

jelmer (changes requested):
You'll also need to update python-distributions.yml - that's where the actual wheels are built and uploaded

jelmer (approved):

edgarrmondragon:
I think the error is in the 3.13 workflow is https://github.com/python/cpython/issues/122728. The `3.13.0rc2` release will probably be alright.

Issue #1346, opened 2024-07-13
dulwich.io Website and Documentation Down


maybephilipp:
dulwich.io is down again

BlazarKnight:
It got archived!!!!!!! [https://web.archive.org/web/20250114030653/https://www.dulwich.io/](https://web.archive.org/web/20250114030653/https://www.dulwich.io/)

kaycebasques:
Website seems to be down again 

![Image](https://github.com/user-attachments/assets/8ae44767-9ea7-4e2e-a90a-cd5e98defa9c)

https://www.isitdownrightnow.com/dulwich.io.html

PR #1377, opened 2024-10-11
remove broken website link

matkoniecz:
see #1346 (I cannot make PR to remove it from repository description)


jelmer:
The website is just temporarily down, no point in removing it from the readme.

matkoniecz:
If I understood #1346 right it is down since July (though maybe it was up between time of making that PR and opening of that issue?)

PR #1403, opened 2024-10-30
Test symrefs handling for git protocol v2

stspdotname:
This adds missing compat tests for symref handling in git protocol v2 and fixes a bug in the HTTP client implementation uncovered by the tests.


jelmer (changes requested):
Thanks! Changes look good, but please take a look at the ruff failures.

stspdotname:
Forgot to run ruff format locally. Fixed now.

jelmer (approved):

Issue #1410, opened 2024-11-02
`GitClient.get_refs()` doesn't return peeled refs anymore (git protocol v2)


abderrahim:
The following code snipped prints 198 when ran against dulwich 0.22.1 and 0 when ran against latest 0.22.4.

```
from dulwich.client import get_transport_and_path
from dulwich.refs import ANNOTATED_TAG_SUFFIX

client, path = get_transport_and_path("https://github.com/jelmer/dulwich.git")
refs = client.get_refs(path)

print(len([k for k in refs.keys() if k.endswith(ANNOTATED_TAG_SUFFIX)]))
```

The issue seems to be git protocol version 2. In the newer protocol, [`ls-refs`](https://git-scm.com/docs/protocol-v2#_ls_refs) command needs a special argument `peel` to return peeled tags.

While I think ultimately it would be a good idea to add more flexibility to getting refs afforded by protocol version 2 (either by adding optional arguments to the current method, or having a new method), the default behaviour should closely mirror what you get with protocol version 1.

nanonyme:
Right, this is fairly large functional change and if desired, should probably happen in 0.23.0, not 0.22.4. @jelmer any thoughts?

jelmer:
cc @stspdotname, as this relates to v2.

I'd argue that actually this is regression and that we should fix this in 0.22. I don't think it's necessarily a big change.

stspdotname:
Agreed. See the PR #1413

PR #1413, opened 2024-11-02
return peeled refs from GitClient.get_refs() if protocol-v2 is used

stspdotname:
This aligns protocol-v2 behaviour with earlier protocol versions. Add test coverage for this behaviour, too.

Fixes issue #1410

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -1378,6 +1378,7 @@ class TraditionalGitClient(GitClient):
                 proto.write_pkt_line(b"command=ls-refs\n")
                 proto.write(b"0001")  # delim-pkt
                 proto.write_pkt_line(b"symrefs")
+                proto.write_pkt_line(b"peel")
                 for prefix in ref_prefix:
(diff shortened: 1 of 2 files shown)


dulwich/client.py
```diff
@@ -285,14 +284,14 @@ def read_pkt_refs_v2(
         ref = parts[1]
         for part in parts[2:]:
             if part.startswith(b"peeled:"):
-                peeled[ref] = part[7:]
+                refs[ref + PEELED_TAG_SUFFIX] = part[7:]
```
jelmer, line 287: please avoid deserializing and then reserializing; this also seems unnecessary.

Aren't the four lines around line 2680 all that we need?

stspdotname: Something like those four lines, indeed. However there are two different implementations of get_refs() and we need to satisfy both of them. I have now force-pushed a smaller diff which doesn't have unnecessary changes.

jelmer (approved):

Issue #1389, opened 2024-10-16
Porcelain Clone 0.22.3 not Cloning GitHub Branches Correctly in at least one case


erl-hpe:
Using dulwich.porcelain clone() at the 0.22.3 release level to clone

```
git@github.com:Cray-HPE/vtds-configs.git
```

I wind up with refs that contain a single branch named 'master'. Unfortunately, the remote repository no 'master' branch, but it does have a 'main' branch. In other words, 'master' in the clone should be 'main' but is not.  When I do the same thing with 0.22.1 I get what I expect.

Here is a transcript of the (correct) 0.22.1 behavior:

```
$ python3
Python 3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dulwich.porcelain import clone
>>> url='git@github.com:Cray-HPE/vtds-configs.git'
>>> target='foo'
>>> clone(url, target)
<Repo at 'foo'>
Enumerating objects: 248, done.
Counting objects: 100% (248/248), done.
Compressing objects: 100% (100/100), done.
Total 248 (delta 109), reused 247 (delta 109), pack-reused 0 (from 0)
copied 247 pack entries47/248
>>> rating index: 247/248
```
and the list of branches in the resulting repo:

```
$ cd foo
$ git branch -l
* main
```

Here is what I see with 0.22.3:
```
$ python3
Python 3.11.6 (v3.11.6:8b6ee5ba3b, Oct  2 2023, 11:18:21) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from dulwich.porcelain import clone
>>> url='git@github.com:Cray-HPE/vtds-configs.git'
>>> target='foo'
>>> clone(url, target)
<Repo at 'foo'>
Enumerating objects: 207, done.
Counting objects: 100% (207/207), done.
Compressing objects: 100% (78/78), done.
Total 207 (delta 94), reused 204 (delta 93), pack-reused 0 (from 0)
copied 206 pack entries06/207
>>> rating index: 206/207
```

and the branches:

```
(venv) HPE-L91WD9JY2R:tmp erl$ cd foo
(venv) HPE-L91WD9JY2R:foo erl$ git branch -l
* master
```

jelmer:
[...]

PR #1416, opened 2024-11-05
fix fallback ref-prefix used by porcelain.clone()

stspdotname:
The porcelain layer fell back on just "HEAD" as the default ref-prefix, exluding all other refs from the cloned repository.

Fixes issue #1389

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -488,7 +488,7 @@ def init(path=".", *, bare=False, symlinks: Optional[bool] = None):
 
 def encode_refspecs(refspecs, refspec_encoding):
     if refspecs is None:
-        return [b"HEAD"]
+        return [b"HEAD", b"refs/"]
 
     def encode_refspec(ref):
         if isinstance(ref, bytes):


dulwich/porcelain.py
```diff
@@ -488,7 +488,7 @@ def init(path=".", *, bare=False, symlinks: Optional[bool] = None):
 
 def encode_refspecs(refspecs, refspec_encoding):
     if refspecs is None:
-        return [b"HEAD"]
+        return [b"HEAD", b"refs/"]
```
jelmer, line 491: This is significantly changing the behaviour, as it means not just discovering all refs but also fetching them and populating the local repository with them.

jelmer (changes requested):

dulwich/porcelain.py
jelmer, line 491: This is a significant change in behaviour, as it doesn't just change what refs we discover but forces us to fetch all refs rather than just HEAD.

stspdotname: With protocol v1, server-side ref filtering is non-existent.
With protocol v2, the client can ask the server to filter refs, but the server is free to return a non-filtered result. So in any case, the client needs to be prepared to filter refs on its own.

But indeed, I used this parameter to test the protocol v2 functionality. It was never intended to be used with protocol v1. I agree that having this parameter work consistently across all protocol versions would be much better.

The problem where Github does not return all refs unless we ask for a filter only affects fetches with protocol v2 over SSH. Protocol v2 over HTTP works fine.  I consider this to a bug in their git/ssh server. We're just trying to work around that.

Issue #1405, opened 2024-10-30
Is it possible not to force dependency on Rust?


barracuda156:
Recent update pulled in Rust and a lot of related stuff as dependencies. This is undesirable both due to just being heavy and due to Rust being broken on some non-mainstream platforms (which makes it impossible to install this at all).

barracuda156:
UPD. I apologize, I should have checked README first. Apparently Rust can be disabled, and it is MacPorts fault not to consider that. I close this and will make a PR to MacPorts.

barracuda156:
@jelmer Sorry, I have to reopen this: Rust is unconditionally pulled in here: https://github.com/jelmer/dulwich/blob/e2ccec69613a10d8e0abc5a480b66ca55b9a7589/setup.py#L9

Or do I miss something? I think I passed `--pure` arg, but it still wants `setuptools_rust`.

barracuda156:
```
--->  Building py312-dulwich
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_python_py-dulwich/py312-dulwich/work/dulwich-0.22.3" && /opt/local/Library/Frameworks/Python.framework/Versions/3.12/bin/python3.12 -m build --no-isolation --config-setting build-option=--pure --wheel --outdir /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_python_py-dulwich/py312-dulwich/work 
* Getting build dependencies for wheel...
Traceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
    main()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 373, in main
    json_out["return_val"] = hook(**hook_input["kwargs"])
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py", line 143, in get_requires_for_build_wheel
    return hook(config_settings)
           ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1417, opened 2024-11-05
build: respect the "pure" argument at metadata generation time

eli-schwartz:
When a pure build is requested, no compiled extensions are produced. The setuptools_rust dependency ends up imported-but-not-used. Teach setuptools to avoid build-requiring it in this case.

Closes: https://github.com/jelmer/dulwich/issues/1405

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@ import os
 import sys
 
 from setuptools import setup
-from setuptools_rust import Binding, RustExtension
 
 if sys.platform == "darwin" and os.path.exists("/usr/bin/xcodebuild"):
     # XCode 4.0 dropped support for ppc architecture, which is hardcoded in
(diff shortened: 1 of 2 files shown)


jelmer (approved):
Neat! I stand corrected - thanks for the fix.

setup.py
```diff
@@ -35,38 +34,45 @@
...
-    sys.argv.remove("--pure")
+if "PURE" in os.environ or "--pure" in sys.argv:
+    if "--pure" in sys.argv:
+        sys.argv.remove("--pure")
+    setup_requires = []
     rust_extensions = []
+else:
+    setup_requires = ["setuptools_rust"]
+    # We check for egg_info since that indicates we are running prepare_metadata_for_build_*
+    if "egg_info" in sys.argv:
```
jelmer, line 46: how future-proof is this?

eli-schwartz: It is heavily baked into the setuptools design so if it ever did change I would expect it to be *extremely* long-term... and there are a bunch of packages that work like this already.

@jaraco / @abravalheri could confirm if there's any plans to change it -- but I think it would break far too many projects to be worth changing, especially if no replacement is available for the existing comment:

```
# Ideally, setuptools would just provide a way to do this
```

Either setuptools needs to provide a full-fledged API for defining build options, or it needs to allow people to execute build options in setup.py as pure python code hooking into phases such as `egg_info` and conditionally defining build metadata.

jaraco: Setuptools is driving users toward standards-based tooling. Setuptools' PEP 517 implementation passes `setup_requires` through as `get_requires_for_build_*`. However, I'd only expect this setting to have effect when building the sdist or the wheel. I'd not expect it to occur when running `egg-info`.

Regardless, Setuptools has deprecated use of `setup.py` invocations, so if this behavior is needed, I'd work to investigate alternative reasons to support `setup.py egg_info`.

eli-schwartz: We aren't invoking setup.py here, the expected use of dulwich is to run `python -m build` and as of today, this PR works fine.

When `python -m build` runs the official `get_requires_for_build_*` hook provided by `setuptools.build_meta`, setuptools internally runs egg_info, that is why we check `sys.argv` -- because we only want to define Rust extensions  when running `build_wheel`, not when running `get_requires_for_*` (setup.py will error out if the dependencies aren't installed).

It is necessary because setup.py is being used to execute code as part of `setuptools.build_meta` -- no one is running `python setup.py ....` at any point.

eli-schwartz: I'd be interested to hear if there's a better way to tell, when using setup.py as a configuration file for:
- configuring extension modules
- defining additional metadata for `get_requires_for_build_*`

which PEP 517 hook is currently running.

My understanding is that currently this means checking `sys.argv`. If it contains "egg_info" then the PEP 517 hook for `get_requires_for_build_*` is running, but if it contains "build" or "build_py" or "build_ext" then one of those 3 `setuptools.setup(cmdclass={...}` are running (possibly overridden but in this case we don't override them).

PR #1420, opened 2024-11-05
Drop broken refspecs argument for porcelain.clone

jelmer:
This argument currently only works for v2 and is untested.

when encoding refspecs, handle the absence case (thanks to @stspdotname in https://github.com/jelmer/dulwich/pull/1416 for finding this)

Also, don't provide separate encoding argument for refspecs. This is consistent with other arguments. Callers that really need a non-standard encoding can do their own encoding and pass in a bytestring.


stspdotname:
Fine with me for now. I can revisit this later.

PR #1432, opened 2024-11-14
intial unit test and impl for ssh-keygen simulator

castedo:
* unit test that check-novalidate subcmd rejects
* basic CLI impl for ssh-keygen simulator
* ~~trivial always fail impl of check-novalidate subcmd~~


castedo:
Probably not worth merging yet since it doesn't actually to any check. So I've put it in draft mode. But good preview of what's to come soon so you can point out any improvements and changes that make sense.

jelmer:
> @jelmer Pick your poison:
> 
> Soon I'm going to be adding a line of 
> ```
> from __future__ import annotations
> ```
> from https://gitlab.com/perm.pub/hidos/-/blob/923691a51921de59250e80f7fd82fe40fa895637/hidos/sshsiglib/ssh_keygen.py#L1
> so that a return type hint like:
> https://gitlab.com/perm.pub/hidos/-/blob/923691a51921de59250e80f7fd82fe40fa895637/hidos/sshsiglib/ssh_keygen.py#L48
> does not have to be quoted.
> 
> And I think there are few other benefits.
> 
> BUT if you think it's better to not `__future__`, just let me know and I'll quote some of these type hints and go back from the `__future__`.

Why do we need that? Dulwich is python >= 3.9, so annotations shouldn't require a "from __future__ import"

castedo:
> > @jelmer Pick your poison:
> > Soon I'm going to be adding a line of
> > ```
> > from __future__ import annotations
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > from https://gitlab.com/perm.pub/hidos/-/blob/923691a51921de59250e80f7fd82fe40fa895637/hidos/sshsiglib/ssh_keygen.py#L1
> > so that a return type hint like:
> > https://gitlab.com/perm.pub/hidos/-/blob/923691a51921de59250e80f7fd82fe40fa895637/hidos/sshsiglib/ssh_keygen.py#L48
> > does not have to be quoted.
> > And I think there are few other benefits.
> > BUT if you think it's better to not `__future__`, just let me know and I'll quote some of these type hints and go back from the `__future__`.
> 
> Why do we need that? Dulwich is python >= 3.9, so annotations shouldn't require a "from **future** import"

It is still needed for recursive/self-referential type hints. For instance, in 

https://gitlab.com/perm.pub/hidos/-/blob/923691a51921de59250e80f7fd82fe40fa895637/hidos/sshsiglib/ssh_keygen.py#L48

the _unquoted_ `AllowedSigner` return type hint is for a function of the class `AllowedSigner`. So even with 3.9, either the type hint needs to be quoted or the `__future__` is needed.

castedo:
I'm usually in a container with Python 3.12 and the choice still is needed in 3.12.

tests/test_ssh_keygen.py
castedo, line 1: @jelmer You probably have a better idea than I do on the best way to handle this with unittest. This DOUBLE_CHECK_SSH_KEYGEN_SIMULATION flag will enable also running the **real** `ssh-keygen`along with the main unit test, with the same test case input data, and make sure the exit return code is the same.

I'm thinking this double checking against the real ssh-keygen should be off by default. It's not really the point of the usual unittest. But it's also nice to sanity double check from time to time.

Makefile
castedo, line 1: @jelmer I wasn't sure the best way to handle this mypy strictness. Mypy seems to be pretty annoying in terms of selectively enable strict mode (it doesn't really seem to be possible). But I've written all the sshsiglib code passing under strict mode. So seems a pity to not try and keep it type strict.

castedo:
Interesting ... I just discovered another benefit to using `from __future__ import annotations`:

ruff requests using the `X | Y` type notation instead of Union and Option, which Python 3.9 barfs on (I just confirmed) ... but without `__future__`. With the `__future__` Python 3.9 will ignore the pipe `X | Y` type notation. I'll start following ruffs suggestion by using `__future__` to keep both ruff and Python 3.9 happy.

castedo:
@jelmer OK, this is an actual implementation of `ssh-keygen -Y check-novalidate`, the core crypto functionality at the base of SSH key signing. This could be a good check point to merge, even though there's more to come, and some more extra sshsig features layerd on top that git uses (namely checking the allowed_signers file for the key found in the signature from this core stage).

castedo:
I figured I would just throw all the code in to one file, to keep it simple within the bigger picture of dulwich. However, there are a number of natural splits and layers to this implementation. This single file could naturally be split into multiple and placed within a subdirectory/submodule for all this sshsig/ssh_keygen stuff.

tests/test_ssh_keygen.py
```diff
@@ -0,0 +1,98 @@
...
+pQWO8AJXrx7gcPAcCFGQ0=
+-----END SSH SIGNATURE-----
+""",
+)
+
+crazy_ascii = b"Nobody expects the Spanish ..."
+crazy_unicode = "Nobody expects 🥘💃🐂 ...".encode()
+
+
+class CLITests(TestCase):
```
jelmer, line 54: Please add tests for the actual lower level functions - I'm not expecting us to actually provide the cli to users, they can use ssh-keygen itself for that.

Compatibility tests (such as for ssh-keygen here) should probably live in tests/compat, and run when ssh-keygen is actually installed on the system - consistently rather than just when a flag is set.

Makefile
```diff
@@ -46,6 +46,7 @@ check-all: check check-pypy check-noextensions
 
 typing:
 	mypy dulwich
+	mypy --strict dulwich/ssh_keygen.py
```
jelmer, line 49: There's multiple places where we invoke mypy; I'd rather enable strictness on a per-project basis, separate of this PR. If you insist, please add this in pyproject.toml so all mypy invocations use it.

dulwich/ssh_keygen.py
```diff
@@ -0,0 +1,296 @@
+# (c) 2018 Mantas Mikulėnas <grawity@gmail.com>
+# (c) 2024 E. Castedo Ellerman <castedo@castedo.com>
+# Released under the MIT License (https://spdx.org/licenses/MIT)
```
jelmer, line 3: Please use the standard preamble (Apache-2/GPLv2+) as elsewhere in the project, or include the MIT license here verbatim since we don't explicitly ship a copy anywhere.

castedo: I like the option of including the MIT license. Short and less viral! :+1: Will do that.

Makefile
castedo: I think the problem is strict mode is either on or off per mypy invocation and can not be enabled or disabled on a per-submodule basis. I think closest one can get to "enabling strictness on a per-project basis" is to make separate invocation of mypy specifically scoped to a sub-project.

I don't think you actually want strict mode on for all mypy invocations because most invocation will also type check all other other Python files which is going to generate a bazillion errors.

Something I can cook up if you want to see it is turn on strict for all innovations and then put in a bazillion disable settings for all the submodules EXCEPT ssh_keygen (and whatever files will pass strict mode). It's kind of a pain in the neck though. But if the long-term plan is to get lots of files passing strict, it's the way to go.

tests/test_ssh_keygen.py
castedo: Nice! I had not looked at tests/compat. Looks like a good approach. I'm going to think about what should go in tests/compat and what stays in tests/test_????.py.

I was imagining that, similar to `python3 -m dulwich ...`, an new possibility would be
```
python3 -m dulwich.ssh_keygen -Y ...
```
mainly for troubleshooting and debugging reasons. But I can't really think of a compelling reason for this versus somebody just cloning the Dulwich GitHub repo and running a `ssh_keygen.py` testing/compat/util script that does not ship with the Dulwich package.

I'll look into having `ssh_keygen.py` be a testing/troubleshootin/debugging script that lives in the Dulwich repo but not in the code that gets packaged.

Then maybe `sshsig.py` is a good filename for the Python script that does get packaged. Or `sshsiglib.py`. Any thoughts or prefs?

dulwich/ssh_keygen.py
castedo: I just educated myself on the Apache license and now understand it is not copyleft. So I'll just use the standard preamble to keep things simple. My desire is to be contribute to the public domain and avoid copylefting unless really needed.

castedo:
> Interesting ... I just discovered another benefit to using `from __future__ import annotations`:
> 
> ruff requests using the `X | Y` type notation instead of Union and Option, which Python 3.9 barfs on (I just confirmed) ... but without `__future__`. With the `__future__` Python 3.9 will ignore the pipe `X | Y` type notation. I'll start following ruffs suggestion by using `__future__` to keep both ruff and Python 3.9 happy.

Another option here is to stay consistent with the rest of the code and ignore this suggestion from ruff by adding `UP007` (Use `X | Y` for type annotations) to the pyproject ruff ignore list.

tests/test_ssh_keygen.py
jelmer: What's the purpose of having ssh_keygen.py over just running the ssh-keygen command? I'd rather have proper unit tests for the lower level code than to have to run another command and verify that it behaves the same as ssh-keygen.

castedo: Next commit is going to have all the command line option parsing cut out. I still have it working over at gitlab `dulwich-sshsiglib/sshsiglib` so I can add a utility back into dulwich later and separately if you'd like. You'll find the new commit to dulwich much more proper.

As far as reasoning behind having the CLI... the CLI is the unambiguous "ground truth", more so than the best most proper unit test. If some input is not working as expected with the low level code, a sensible troubleshooting step is 1st) see how it behaves via the CLI with the real ssh-keygen, then 2nd) confirm the ssh_keygen.py CLI matches or not. Otherwise there is ambiguity as to whether one is simply calling the low-level function improperly or in the wrong state etc... vs the low-level functions are actually not working properly.

Makefile
castedo: I can keep track of strictly typed files from my beautiful justfiles in the dulwich-sshsiglib union repo :wink:

castedo:
@jelmer OK, I believe all outstanding issues are resolved now. I've also rebased and squashed.

There are some minor trade-offs between using `__future__`, using quoted type names, using pipe symbol instead of Union/Optional, and a ruff ignore setting. I'm fine switching the code be older-style Python. But I went ahead and committed newer style.

Since the key two functions at the end of sshsig.py aren't getting called by anything in Dulwich yet, I'm not sure whether it makes sense to merge this into the master branch. But to me it looks fine to merge now.

From what I know now, there isn't much reason for this code to change. Additional SSH signing functionality is all built on top of this PR and shouldn't require much modification of this PR code.

castedo:
Closing this PR since all of the functionality and unittesting in this PR is now in https://github.com/castedo/sshsig.

PR #1437, opened 2024-11-15
Reject refcontainer values that are not 40 char sha or symref

abn:


jelmer:
In what situation are you encountering this? My understanding is that short refs can't appear in refs files.

abn:
We started seeing this in python-poetry/poetry#9560. When the user uses a a short sha we set it at https://github.com/python-poetry/poetry/blob/main/src/poetry/vcs/git/backend.py#L280C12-L280C56.

I suspect we might also need to add a call to `scan_for_short_id` on our end as well. Or fix how we resolve commit ids when it is not part of a branch.

jelmer:
> We started seeing this in [python-poetry/poetry#9560](https://github.com/python-poetry/poetry/issues/9560). When the user uses a a short sha we set it at https://github.com/python-poetry/poetry/blob/main/src/poetry/vcs/git/backend.py#L280C12-L280C56.
> 
> I suspect we might also need to add a call to `scan_for_short_id` on our end as well. Or fix how we resolve commit ids when it is not part of a branch.

Ah, I see. Yeah, pretty sure you're not meant to store short shas like that. I think the correct fix here is actually for dulwich to reject short shas (and only allow symrefs or liong shas) when assigned.

abn:
Happy either way. I am guessing that would mean we need to handle short to long conversion in our code?

jelmer:
> Happy either way. I am guessing that would mean we need to handle short to long conversion in our code?

Yeah, I think regardless of the question of whether it's valid to include a short sha in a ref file, it would be very unusual (I've never seen it) so I would try to stay consistent with other tools and resolve the short sha before you write out the refs file.

There is a ``ObjectStore.iter_prefix`` function in modern versions of dulwich that you can use to efficiently iterate over all shas with a particular prefix.

It would be great if the RefsContainer implementation in Dulwich coul;d reject shas that are not 40 characters *or* start with "ref: " (a symref). That would encourage users to do the right thing here.

abn:
I will attempt a patch here shortly to reject things that are not a long sha or a symref.

jelmer (approved):

jelmer:
Thanks!

PR #1438, opened 2024-11-16
add CGit compat test case for sign/verify commit

castedo:


tests/compat/test_porcelain.py
```diff
@@ -91,3 +91,54 @@ def test_verify(self):
         tag = self.repo[b"refs/tags/verifyme"]
         self.assertNotEqual(tag.signature, None)
         tag.verify()
+
+
+@skipIf(
+    platform.python_implementation() == "PyPy" or sys.platform == "win32",
```
jelmer, line 97: please make this dependent on gpgme actually being importable. Even on linux people might not have it installed and if we do manage to install gpgme on mac or win32 in the ci, we wouldn't have to manually update this code.

tests/compat/test_porcelain.py
```diff
@@ -91,3 +91,54 @@ def test_verify(self):
...
+    "gpgme not easily available or supported on Windows and PyPy",
+)
+class CommitCreateSignTestCase(PorcelainGpgTestCase, CompatTestCase):
+    def setUp(self):
```
jelmer, line 101: This setUp() is a noop, remove it?

jelmer (approved):
LGTM, two minor comments.

tests/compat/test_porcelain.py
castedo: I believe it is. CommitCreateSignTestCase inherits from PorcelainGpgTestCase which has a decorator
```
@skipIf(gpg is None, "gpg is not available")
```
in `dulwich/tests/test_porcelain.py`.

I don't 100% understand these `@skipIf` decorators but I suspect the current code is doing what you want.

tests/compat/test_porcelain.py
castedo: I'm not sure. I don't know Python multiple inheritance. I don't know what the default behavior is. I'm guessing `super()` only accesses the first parent class. Do you know?

You're thinking it's a noop because it just re-implements the default behavior?

castedo: BTW, this setUp is a copy of the tag test case above it. I'm just copying the pattern of the existing test case for a tag, but re-adapting it to a commit.

jelmer: See https://docs.python.org/3/library/functions.html#super, I'm pretty sure dropping this is a noop.

I've fixed both of these things in main for the existing test; see https://github.com/jelmer/dulwich/pull/1439

castedo:
@jelmer What the heck is up with ruff? The error it triggers looks bogus. The names ruff claims are not defined are defined. They are imported in the same file.

jelmer:
This is because I removed the imports in main when I changed the behaviour for the existing tag tests.

If you rebase, you should see it locally.

castedo:
> @castedo This is because I removed the imports in main when I changed the behaviour for the existing tag tests.
> 
> If you rebase, you should see it locally.

Ah, I see now. I was under the mistaken impression that the Git Action checks run on the branch of the PR, but in fact they are running on a merge of the PR branch and the main/master branch.

PR #1441, opened 2024-11-16
Fix empty message serialization

### LINKED ISSUE #1429: GPG signature verification fails incorrectly for commits with empty message
I've hit this problem in my messy dev environment. I figure I should make a PR for a nice clean repo case as a unit test. Do you have a recommendation on where commit signature verification test cases should go?

An example of a GPG signed commit that I am 95% sure will fail incorrectly is this one:

https://archive.softwareheritage.org/api/1/revision/92b3f55bcec30906a39796c90018475bc1cbabb7/

I think the problem is the calling of `_format_message` with an argument of `self._message` to the _body_ parameter:

https://github.com/jelmer/dulwich/blob/e7e83d6e97f306c5d2963be7e370f85fe9f6f263/dulwich/objects.py#L1576

When the commit message is empty, I believe this causes the newline on this line:

https://github.com/jelmer/dulwich/blob/e7e83d6e97f306c5d2963be7e370f85fe9f6f263/dulwich/objects.py#L730

that is inside the if statement (instead of above it) to not get output (when it should). But when a signature-less commit is generated on this line:

https://github.com/jelmer/dulwich/blob/e7e83d6e97f306c5d2963be7e370f85fe9f6f263/dulwich/objects.py#L1531

the GPG signature verification will fail when the commit has an empty commit message, because a newline will be missing at the end.

castedo:
Fixes #1429, incorrect signature verification for commits with an empty message, as encoded by CGit.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -728,8 +728,8 @@ def _format_message(headers, body):
         yield git_line(field, lines[0])
         for line in lines[1:]:
             yield b" " + line + b"\n"
+    yield b"\n"  # There must be a new line after the headers
     if body:
-        yield b"\n"  # There must be a new line after the headers
         yield body
 
 
--- a/tests/compat/test_porcelain.py
+++ b/tests/compat/test_porcelain.py
@@ -126,9 +126,32 @@ class CommitCreateSignTestCase(PorcelainGpgTestCase, CompatTestCase):
                 "commit",
                 "--allow-empty",
                 "-S" + PorcelainGpgTestCase.DEFAULT_KEY_ID,
-                "--allow-empty-message",
                 "-m",
                 "foo",
+            ],
+            env={
+                "GNUPGHOME": os.environ["GNUPGHOME"],
+                "GIT_COMMITTER_NAME": "Joe Example",
+                "GIT_COMMITTER_EMAIL": "joe@example.com",
+            },
+        )
+        commit = self.repo[b"HEAD"]
+        self.assertNotEqual(commit.gpgsig, None)
+        commit.verify()
+
+    def test_verify_with_empty_message(self):
+        # Test that CGit signatures can be verified by dulwich
+        self.import_default_key()
+
+        run_git_or_fail(
+            [
+                f"--git-dir={self.repo.controldir()}",
+                "commit",
+                "--allow-empty",
+                "-S" + PorcelainGpgTestCase.DEFAULT_KEY_ID,
+                "--allow-empty-message",
+                "-m",
+                "",
             ],
             env={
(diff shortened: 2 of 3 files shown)


castedo:
@jelmer I've made this a draft, because the change still seems terrifying to me. I haven't read and understood enough of the code to feel great about this change. I'm planning to read and understand some more of the code.

What do you think? You see any issues?

An addition improvement to consider is to strongly type the `message` attribute of commit to be a *non-optional* bytes string and only a string. In other words, it can not be None. I think the idea of a missing message in a commit is more error prone than a required message, but the message can be empty. But this re-typing of message in commit is probably not worth the time/effort.

jelmer:
> @jelmer I've made this a draft, because the change still seems terrifying to me. I haven't read and understood enough of the code to feel great about this change. I'm planning to read and understand some more of the code.
> 
> What do you think? You see any issues?
> 
> An addition improvement to consider is to strongly type the `message` attribute of commit to be a _non-optional_ bytes string and only a string. In other words, it can not be None. I think the idea of a missing message in a commit is more error prone than a required message, but the message can be empty. But this re-typing of message in commit is probably not worth the time/effort.

This doesn't look correct to me. If you ask C Git to create a commit message that is empty, you still end up with two lines after the last header.

jelmer:
> @jelmer I've made this a draft, because the change still seems terrifying to me. I haven't read and understood enough of the code to feel great about this change. I'm planning to read and understand some more of the code.
> 
> What do you think? You see any issues?
> 
> An addition improvement to consider is to strongly type the `message` attribute of commit to be a _non-optional_ bytes string and only a string. In other words, it can not be None. I think the idea of a missing message in a commit is more error prone than a required message, but the message can be empty. But this re-typing of message in commit is probably not worth the time/effort.

This looks correct to me; if you create a commit with an empty commit message in C git, you still end up with two newlines after the last header.

castedo:
Reading the code, I'm seeing that this change is affecting a code path when a tag lacks a non-empty message and lacks a signature. So this code change is also going to cause an extra newline in a tag corner case.

I'm going peek at some more unit tests to see if they can help confirm that a new newline is appropriate in the tag case.

castedo:
Doing `git cat-file -p ... | xxd` indicates that the logic is appropriate for tags too. When there is no message, cgit still places two newlines after the headers.

castedo:
OK, i've looked over the code/tests, looked at `git cat-file -d .. | xxd` with some tags and commits with empty messages and this looks good to me. I give a :+1: to merge!

jelmer:
Thanks for the careful checking!

PR #1447, opened 2024-11-19
expose Commit.raw_without_gpgsig method

castedo:


castedo:
@jelmer Now that I've got a fairly complete SSHSIG functionality set implemented for git verification purposes, the various possible future paths/approaches and trade-off is becoming much more clear.

This PR is an easy tweak that makes a number of possible future paths more elegant.

Here is the slightly inelegant solution that I have coded in `hidos` right now:

```
def verify(commit: Commit, allowed_signers: set[PublicKey]) -> bool:
    try:
        sshsig.verify(commit_without_sig(commit), commit.gpgsig.decode(), allowed_signers)
        return True
    except InvalidSignature:
        return False

def commit_without_sig(self: Commit) -> bytes:
    tmp = self.copy()
    assert isinstance(tmp, Commit)
    tmp._gpgsig = None
    tmp.gpgsig = None
    ret = tmp.as_raw_string()
    if self.message == b'' and not ret.endswith(b"\n\n"):
        # work around for https://github.com/jelmer/dulwich/issues/1429
        ret += b'\n'
    return ret
```
The actual code with a dumb function name is here:

https://gitlab.com/perm.pub/hidos/-/blob/5f8f578a79191db19b4ee6c9f2e7abe85be2ad19/hidos/dulwich.py#L131

With this PR change future code could be something more like
```
        sshsig.verify(commit.without_sig(), commit.gpgsig, allowed_signers)
```

There's a whole other can of worms regarding what method/exceptions interface Commit.verify should have. But I'll bring that up for discussion elsewhere.

For now these seems like an easy improvement. Not sure if `Commit.without_gpgsig` should be a property like `Commit.gpg` or a function like `Commit.as_raw_string()`. I figured since it's probably NOT pre-computed it should be a function. But either way seems reasonable to me.

dulwich/objects.py
```diff
@@ -1506,6 +1506,13 @@ def sign(self, keyid: Optional[str] = None):
                     self.as_raw_string(), mode=gpg.constants.sig.mode.DETACH
                 )
 
+    def without_gpgsig(self) -> bytes:
```
jelmer, line 1509: the name without_gpgsig suggests to me that this will return a Commit rather than bytes.

jelmer, line 1509: please add a docstring and a test

jelmer (changes requested):
the overall change looks good, but some minor comments inline

dulwich/objects.py
castedo: good point. Some brainstorming here:

1. `commit.bytes_without_gpgsig()`
2. `commit.as_raw_without_gpgsig()`

And `gpgsig` could be shortened to `sig` which has some logic to it because the signature can be with ssh instead of gpg.

castedo: I like the idea of `commit._____without_sig()` except that a user might be unsure whether the `sig` removed is the same contents as what the property `gpgsig` returns.

jelmer: does it need to return the bytes rather than the actual object? The caller could always call .as_raw_string() if they wanted to.

Dropping the "gpg" bit makes sense to me.

castedo: And there's `def __bytes__(self): -> bytes` too.

Let's image a user wants to check the commit signature and any public key is ok.
```
check_signature(commit.raw_without_sig(), commit.gpgsig)
```
v.s.
```
check_signature(bytes(commit.without_sig()), commit.gpgsig)
```
or equiv
```
check_signature(commit.without_sig().as_raw_string(), commit.gpgsig)
```

Ya, returning the object and doing `bytes(commit.without_sig())` looks like a winner to me! What do ya think?

jelmer: looks good to me too :)

castedo: One heads up on this return-a-Commit approach. If there is a Commit.without_sig() it makes sense there should also be a Tag.without_sig(). And the current plan is the equivalent signed cryptographic message for a tag is accessed by doing:
```
bytes(tag.without_sig())
```
which to a human reader is intuitive.

However that will be an inefficient way to accomplish what is currently done in the code as:
```
self.as_raw_string()[: -len(self._signature)]
```
where `self` is a tag object.

But I'm thinking this is OK since this is Python, not C or Rust. But my old brain can't help but cringe a bit at how much CPU and RAM is used by friendly readable Python :nerd_face:.

castedo:
I've rebased/squashed to the latest tip. doctring and unit testing added/enhanced. New typing looking nice!

dulwich/objects.py
```diff
@@ -1510,6 +1510,18 @@ def sign(self, keyid: Optional[str] = None) -> None:
                     self.as_raw_string(), mode=gpg.constants.sig.mode.DETACH
                 )
 
+    def without_sig(self) -> "Commit":
+        """Return a Commit object copy with the signature (GPG or SSH) removed.
+
+        self.gpgsig is a signature for the raw string serialization bytes of the
```
jelmer, line 1516: this sentence doesn't make sense to me

dulwich/objects.py
```diff
@@ -1510,6 +1510,18 @@ def sign(self, keyid: Optional[str] = None) -> None:
...
+        Commit object returned.
+        """
+        ret = self.copy()
+        assert isinstance(ret, Commit)
+        ret._gpgsig = None
+        ret.gpgsig = None
```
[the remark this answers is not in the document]
castedo: Are you sure you want that removed? I'm repeating the code that has been on the master master branch prior to this PR.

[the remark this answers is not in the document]
jelmer: in that case it's fine because the object isn't returned and discarded after verifying the signature. here, the caller might rely on Commit.gpgsig actually working later.

dulwich/objects.py
castedo: One challenge I have is there are multiple different computing concepts that use the words "message" or "signature". In an earlier version of this text I used the phrase "cryptographic message" to refer to the "message" from the perspective of cryptographic signing. Do you find the phrase "cryptographic message" confusing? I worry it will be confused with the git commit message.

jelmer: Actually, nevermind. I think I understand what is trying to say upon rereading it.

castedo: Here's a more wordy alternative:

"The raw byte string serialization of the returned Commit object is the cryptographic message for which self.gpgsig is the signature."

Do you think this alternative is more understandable?

castedo: This forced-pushed squashed [934b017](https://github.com/jelmer/dulwich/commit/934b017555e168eeff548af33d323aeeac5e2e8e) of this PR branch has a very tiny wording tweak.

The alternative wording above is in a further commit as branch https://github.com/castedo/dulwich/tree/commit_without_sig2 .

I can't tell which wording is better because I'm not sure how well "cryptographic message" is understood as intended.

jelmer:
> @jelmer In light of this regression, and a potential corresponding version of `without_sig()` for the Tag object, I can't help but wonder...
> 
> Is there really any important or common scenario where a dev wants a copy of the Commit object, just like the original, but as if it had never been signed, without the signature?
> 
> Isn't really the only important scenario merely to get the cryptographic message corresponding to the signature?
> 
> I don't fully understand how complicated and how many foot-guns there are in trying to return a proper copy. The two commit objects I'm guessing here are mutable, so they might now be sharing container objects for which both Commit objects have attributes point to shared mutable container objects?
> 
> In contrast, if this without_sig() function just returns `bytes` none of these potential foot-guns or shared mutable state is at risk. It could be named `.raw_without_sig()` to be more clearly raw bytes.

I don't see that the regression is related to the fact that we're returning an object rather than the serialized bytes.

Dulwich is used by tools like hg-git and bzr-git, which do a lot of object manipulation. It wouldn't be unreasonable for them to e.g. want to strip the signature and then do further manipulation on the object. If all we cared was fully serialized or deserialized objects, then all we'd need is one serialize/deserialize functions for each one of Commit, Blob, Tree.

castedo:
> > @jelmer In light of this regression, and a potential corresponding version of `without_sig()` for the Tag object, I can't help but wonder...
> > Is there really any important or common scenario where a dev wants a copy of the Commit object, just like the original, but as if it had never been signed, without the signature?
> > Isn't really the only important scenario merely to get the cryptographic message corresponding to the signature?
> > I don't fully understand how complicated and how many foot-guns there are in trying to return a proper copy. The two commit objects I'm guessing here are mutable, so they might now be sharing container objects for which both Commit objects have attributes point to shared mutable container objects?
> > In contrast, if this without_sig() function just returns `bytes` none of these potential foot-guns or shared mutable state is at risk. It could be named `.raw_without_sig()` to be more clearly raw bytes.
> 
> I don't see that the regression is related to the fact that we're returning an object rather than the serialized bytes.
> 
> Dulwich is used by tools like hg-git and bzr-git, which do a lot of object manipulation. It wouldn't be unreasonable for them to e.g. want to strip the signature and then do further manipulation on the object. If all we cared was fully serialized or deserialized objects, then all we'd need is one serialize/deserialize functions for each one of Commit, Blob, Tree.

I don't understand the deeper reason the regression is happening, but it is due to removing the `ret.gpgsig = None` assignment that you requested to be removed and said was not needed. 

I am neutral on whether a `def without_sig(self) -> Commit` method exists. I'm more than happy if you want to do a proper implementation.

I do know that a `def raw_without_sig(self) -> bytes` is easy and much less risky and I can do a PR for it and it will make calling Dulwich with crypto libraries significantly more elegant and less error prone. The digital bytes crypto message and the ascii armored signature are the two fundamental key bits of information needed for signature verification and can be used with many different independent libraries.

This PR is suffering from mission creep. I'm just trying to add a simple method that makes doing signature verification significantly more elegant and less error prone. And now this is turning into a PR of involved in proper copying and modification of mutable objects. I'll leave that challenge to you.

I'll rename the method to `.raw_without_sig()` and have it return `bytes`. If you want to implement a proper `.without_sig() -> Commit` by all means go ahead. After I modify the PR you can merge it or not. Either way is fine by me.

castedo:
@jelmer Nice! This enables calling Dulwich in a manner way more elegant and less error prone than the hacky code I have in hidos right now.

Do you want me to do corresponding `Tag.raw_without_sig()` method and similar unittest enhancements in a new PR? Then the fundamental underlying universal crypto (message, sig) pair interfaces become:

* `(Commit.raw_without_sig(), Commit.gpgsig)` and
*  `(Tag.raw_without_sig(), Tag.signature)`.

Those are actually the only differences between Commit and Tag in the `verify()` method implementations. The `verify()` methods of both Tag and Commit are basically copy and pasted with the only differences being how the fundamental underlying crypto (message, sig) pairs are extracted from Tags vs Commits.

jelmer:
> @jelmer Nice! This enables calling Dulwich in a manner way more elegant and less error prone than the hacky code I have in hidos right now.
> 
> Do you want me to do corresponding `Tag.raw_without_sig()` method and similar unittest enhancements in a new PR? Then the fundamental underlying universal crypto (message, sig) pair interfaces become:
> 
>     * `(Commit.raw_without_sig(), Commit.gpgsig)` and
> 
>     * `(Tag.raw_without_sig(), Tag.signature)`.
> 
> 
> Those are actually the only differences between Commit and Tag in the `verify()` method implementations. The `verify()` methods of both Tag and Commit are basically copy and pasted with the only differences being how the fundamental underlying crypto (message, sig) pairs are extracted from Tags vs Commits.

Yeah, I think that would be great.

PR #1450, opened 2024-11-21
add Tag.raw_without_sig method and unittest

castedo:


castedo:
@jelmer As a completely independent side note, this PR was prepared using https://github.com/castedo/git-prepr which makes it easy (kind of :sweat_smile:) to have clean squashed and rebased PR updates while simultaneously maintaining a separate parallel disposable feature branch that has the full complicated commit and merge history. The disposable feature branch I have local although I could push that server side too. Not sure what happens or if possible if I tried to push grafts to server side.

Issue #1399, opened 2024-10-26
Porcelain pull force mode triggered KeyError


Kiramei:
I use the dulwich in my project and originally I mean to keep the local repository synchronized with the remote repository. Since I tried to get the information mentioned in the previous issues, the problem of `dulwich.porcelain.DivergedBranches` which indicates that the local sha doesn't match with the remote seems to be concerning an unaddressed problem that the merge of the pull can't be solved until the moment. Well, after that, I turned to the command corresponding `git pull --force` like:
```python
  porcelain.pull(repo, REPO_URL_HTTP, 'master', force=True)
```
to settle this problem, but it turned out to be a case of KeyError.

This is the message of the traceback :
```PlainText
Traceback (most recent call last):
  File "c:\Program Files\xxxxxxxxxxxx\test_in.py", line 342, in check_install
    check_git(opt)
  File "c:\Program Files\xxxxxxxxxxxx\test_in.py", line 314, in check_git
    porcelain.pull(repo, REPO_URL_HTTP, 'master', force=True, no_ff=True)
  File "C:\Users\xxxxx\AppData\Local\Programs\Python\Python310\lib\site-packages\dulwich\porcelain.py", line 1349, in pull
    tree = r[b"HEAD"].tree
  File "C:\Users\xxxxx\AppData\Local\Programs\Python\Python310\lib\site-packages\dulwich\repo.py", line 839, in __getitem__
    return self.object_store[self.refs[name]]
  File "C:\Users\xxxxx\AppData\Local\Programs\Python\Python310\lib\site-packages\dulwich\object_store.py", line 150, in __getitem__
    type_num, uncomp = self.get_raw(sha1)
  File "C:\Users\xxxxx\AppData\Local\Programs\Python\Python310\lib\site-packages\dulwich\object_store.py", line 597, in get_raw
    raise KeyError(hexsha)
KeyError: b'aa35e3e861afd7c26171c9e1b3530647a930b0d8'
```

In my case, the work can be concluded as: The remote repository keeps updating, and the local repository should be updated with the remote, while the local changes whether stashed or unstashed are to be dropped. I'd appreciate it if there's someone that can give me a suggestion on it.

jelmer:
Have you done any debugging - is aa35e3e861afd7c26171c9e1b3530647a930b0d8 the correct commit sha? Did it get fetched to the local repository?

Kiramei:
[...]

PR #1458, opened 2024-11-27
add a compat test case which verifies that shallow clones can be deep…

stspdotname:
…ened

This test shows that deepening shallow clones works across all Git protocol versions. The trick to make this work is to pass depth=DEPTH_INFINITE

Related to issue #1399

### DIFF
--- a/tests/compat/test_client.py
+++ b/tests/compat/test_client.py
@@ -186,8 +186,11 @@ class DulwichClientTestBase:
                     committer=b"Joe Example <joe@example.com>",
                     tree=tree_id,
                 )
+        # Create a new client for the second fetch operation
+        # TCP connections are single-use for git protocol operations
+        c2 = self._client()
         with repo.Repo(self.dest) as local:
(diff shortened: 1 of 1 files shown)


jelmer (approved):

jelmer:
Please rebase - the ruff issues are fixed in master.

stspdotname:
These commits are already based on https://github.com/jelmer/dulwich/commit/4de310b6676856434391ab57b55e0e4489967abe
which seems to be the latest at present?

jelmer:
> These commits are already based on [4de310b](https://github.com/jelmer/dulwich/commit/4de310b6676856434391ab57b55e0e4489967abe) which seems to be the latest at present?

Whoops, the PR actually got stuck because of a spurious test failure. Current HEAD should be good.

jelmer (approved):

jelmer:
Changes LGTM, just needs a "ruff format" run?

jelmer:
FWIW it looks like these test failures are repeatable, and appear in the newly added tests. They might be related to general brittleness.

PR #1472, opened 2025-01-03
migrate setup.py to pyproject.toml

deronnax:
an initiative of my own.


setup.py
```diff
@@ -1,79 +0,0 @@
...
-            os.environ["ARCHFLAGS"] = ""
-
-tests_require = ["fastimport"]
-
-
-if "__pypy__" not in sys.modules and sys.platform != "win32":
-    tests_require.extend(["gevent", "geventhttpclient", "setuptools>=17.1"])
-
-
-optional = os.environ.get("CIBUILDWHEEL", "0") != "1"
```
jelmer, line 36: How is this behaviour preserved?

setup.py
```diff
@@ -1,79 +0,0 @@
...
-
-# Ideally, setuptools would just provide a way to do this
-if "PURE" in os.environ or "--pure" in sys.argv:
```
jelmer, line 39: Here as well, please preserve support for the --puro flag and PURE environment variable.

setup.py
```diff
@@ -1,79 +0,0 @@
...
-    if "--pure" in sys.argv:
-        sys.argv.remove("--pure")
-    setup_requires = []
-    rust_extensions = []
-else:
-    setup_requires = ["setuptools_rust"]
-    # We check for egg_info since that indicates we are running prepare_metadata_for_build_*
-    if "egg_info" in sys.argv:
```
jelmer, line 47: Same here

jelmer (changes requested):

deronnax:
Hmm, so we need to produce two different type of wheel depending on the flags passed at build time? I will look for it, but I am afraid that is not something available out of the box with static config.

jelmer:
Yeah, I don't think so either - this is why Dulwich currently has a custom setup.py

jelmer:
It's not the creation of wheels - individual users might want to do a non-wheel local build with one of these options.

jelmer:
Closing this since it seems to have stalled and I don't see a way forward. Feel free to reopen if you disagree.

deronnax:
If I got you right, you mean: do a local editable install `pip install -e`, without building the compiled extension, only using the pure python module, is it right?

jelmer:
Either that, or just running "pip install --extra-options=--pure" .

deronnax:
I played around a bit with `python -m build`, and what I found out:
* if there is a rust compiler, setuptools will build a wheel with the binary extension
* if there is is none, setuptools will output a warning and build a wheel without the binary extension

Isn't good as this?

jelmer:
> I played around a bit with `python -m build`, and what I found out:
> 
>     * if there is a rust compiler, setuptools will build a wheel with the binary extension
> 
>     * if there is is none, setuptools will output a warning and build a wheel without the binary extension
> 
> 
> Isn't good as this?

I'm not sure if I understand your question, but that isn't a sufficient replacement - please preserve the arguments and environment variables we currently support as people rely on them. There is no clear advantage in migrating to pyproject.toml AFAIK.

deronnax:
OK, I won't bother you with this anymore. All the best. Cheers.

Issue #1431, opened 2024-11-12
SSH signature allowed signers file format support


castedo:
Parsing the "allowed signers" file format is needed for SSH signature verification: #1394.

This issue it so track a new function in the dulwich library for parsing "allowed signers" files for SSH signatures. This is a feature implemented by ssh-keygen and I'm writing Python code to emulated its implementation. Even though the "allowed signers" format supports extra functionality that I doubt will be used for git ssh signatures, my thought is the dulwich parsing code should parse the full format without error. After parsing, other layers of Dulwich code can raise errors etc... saying certain features are not supported. But at least it would not be a cryptic file parsing error.

castedo:
@jelmer i've implemented parsing of SSH "allowed signers" format. The unit tests are this file:

https://gitlab.com/perm.pub/dulwich-sshsig-union/-/blob/main/sshsiglib/tests/test_allowed_signers.py

The implementation is the `load_allowed_signers_file` function here:

https://gitlab.com/perm.pub/hidos/-/blob/9277d178bfe544f67a378437f65b2d5e4418a229/hidos/sshsiglib/ssh_keygen.py#L117

The ~100 lines above that function are needed by it. These ~120 lines are the smallest chunk of functionality that I can think of that I could turn into a PR if you want.

castedo:
I have a very minimal level of support for the ssh-keygen "allowed signers" format. I've been calling it the "for-git" sub-format in comments and doc strings.

https://gitlab.com/perm.pub/sshsiglib/-/blob/e1c023d4556094fec9142287343cb65d1fbd710f/sshsig/allowed_signers.py

This minimal sub-format enables all the signature verification functionality that git enables. However, I have discovered there are _TONS_ of fancy power user features in the "allowed signers" format that in theory a git user might have and cgit with classic real ssh-keygen will parse correctly. There really isn't reason a git user _NEEDS_ the fancy format. They could stick to a simple "for-git" sub-format and get the same result.

I am currently not planning to implement any of these fancy power user bonus features of ssh-keygen's allowed signers file format. The sub-format is perfectly fine for git purposes.

It's also easy to imagine that many Dulwich users that want SSH sig verification do not actually want to get allowed public keys from this weird ssh-keygen allowed signers file format. In the end, the important interface is the list of public keys that are allowed to sign and not all the bonus feature of the allowed signers file format (that git doesn't actually need).

[...]

Issue #1394, opened 2024-10-20
verification of commit signatures with SSH key


jelmer:
### Discussed in https://github.com/jelmer/dulwich/discussions/1391

<div type='discussions-op-text'>

<sup>Originally posted by **castedo** October 20, 2024</sup>
I am planning to migrate from using GitPython to Dulwich, but I will need to be able to verify commit signatures with SSH keys.

Am I correct that Dulwich does not have this functionality? It looks like tag signatures with GPG keys is only supported.

Are there any libraries or existing Python code that can do this? I've searched long and far and it seems the only Python code that replicates the core functionality of the type of SSH key signatures with git is https://github.com/grawity/ssh-datasign (thank you @grawity!) This is the [SSHSIG type signature](https://github.com/openssh/openssh-portable/blob/fe4305c37ffe53540a67586854e25f05cf615849/PROTOCOL.sshsig) implemented by `ssh-keygen -Y verify` which is what `git` uses for SSH key signatures ([nice blog post](https://www.agwa.name/blog/post/ssh_signatures)).</div>

castedo:
I've created https://gitlab.com/perm.pub/dulwich-sshsig-union for testing, coordination, and a bit of experimentation.

castedo:
I somehow failed to recognize that my goal of switching hidos from GitPython to Dulwich will also require signing in addition verification. :facepalm: So the mission of this thread has expanded to include signing too. But the code from grawity should make this relatively easy if I'm already doing all the work to test verification well.

castedo:
Quick update on my thoughts, now that I've gotten verification partially working with hidos/sshsiglib/dulwich. I'm now planning to have hidos using Dulwich this month. But I think hidos is going to continue depending on GitPython & classic git for _creating/amending/writing_ document succession for a long time. But reading document successions will be via Dulwich. So verification with Dulwich is high priority but signing commits via Dulwich is low priority for me for awhile. I still want to work towards getting signing into Dulwich too, but it's just not as urgent. I hope signing isn't that much work once verification is working.

castedo:
[...]

PR #1473, opened 2025-01-04
parsing of ssh-keygen allowed signers file format

castedo:
Implements low-level parsing for enhancement #1431

### DIFF
--- a/dulwich/allowed_signers.py
+++ b/dulwich/allowed_signers.py
@@ -1,0 +1,152 @@
+# Copyright (C) 2024 E. Castedo Ellerman <castedo@castedo.com>
+#
+# SPDX-License-Identifier: Apache-2.0 OR GPL-2.0-or-later
+# Dulwich is dual-licensed under the Apache License, Version 2.0 and the GNU
+# General Public License as public by the Free Software Foundation; version 2.0
+# or (at your option) any later version. You can redistribute it and/or
+# modify it under the terms of either of these two licenses.
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# You should have received a copy of the licenses; if not, see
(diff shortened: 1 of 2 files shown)


jelmer:
This is all ssh-specific (rather than Git) and the referenced source file is from openssh, any reason this shouldn't live in sshsig?

castedo:
> This is all ssh-specific (rather than Git) and the referenced source file is from openssh, any reason this shouldn't live in sshsig?

In https://github.com/jelmer/dulwich/issues/1394#issuecomment-2555112203 you wrote:

> The support for parsing allowed signers is git-specific and should be a part of dulwich, not sshsig. Applying the rules in such a file is also non-trivial and you'd want support from dulwich to apply them.

But I'm fine with this code living in sshsig if you prefer it not be in dulwich.

I agree it is ssh specific. The utility`ssh-keygen` is defining the format.

jelmer:
> > This is all ssh-specific (rather than Git) and the referenced source file is from openssh, any reason this shouldn't live in sshsig?
> 
> In [#1394 (comment)](https://github.com/jelmer/dulwich/issues/1394#issuecomment-2555112203) you wrote:
> 
> > The support for parsing allowed signers is git-specific and should be a part of dulwich, not sshsig. Applying the rules in such a file is also non-trivial and you'd want support from dulwich to apply them.
> 
> But I'm fine with this code living in sshsig if you prefer it not be in dulwich.
> 
> I agree it is ssh specific. The utility`ssh-keygen` is defining the format.

Ah, sorry - to maybe make it a bit more nuanced: I think the parsing should be in sshsig and the application (which would be importing the parser from sshsig) in dulwich since that needs dulwich-specific context, if that makes sense? Maybe I am being overly cautious - we can always move the code later anyway.

castedo:
Makes sense to me that dulwich, not sshsig, would have (yet-to-exist) code that replicates the logic of the git CLI as to where to find the allowed signers file based on git config files (and probably environment variables too).

Seems reasonable to me that sshsig contain the basic allowed signers parsing code that I've already coded (and am using in 'hidos' in a hidos specific way).

I've go ahead and close this PR and create a new one in the sshsig repo to have the sshsig package include allowed signer parsing code.

PR #1482, opened 2025-01-23
porcelain: respect 'abbrev' argument when describing annotated-tagless repositories

### LINKED ISSUE #1477: porcelain: 'abbrev' parameter is ignored when describing tagless repositories
Hello,

This may be a known/expected behaviour, based on the [logic](https://github.com/jelmer/dulwich/blob/77c2d8d10695a6dfded196a927047eecf88c542f/dulwich/porcelain.py#L2223-L2225) and [`todo` comment](https://github.com/jelmer/dulwich/blob/77c2d8d10695a6dfded196a927047eecf88c542f/dulwich/porcelain.py#L2182-L2183), but I thought it'd be worth reporting since it has surprised me once or twice, and I don't find existing bugreports about it.

When running `porcelain.describe` on a repository that contains no annotated `git` tags, the `abbrev` argument (prefix-length to return from the `git` commit ID) appears to be ignored and the output includes a commit prefix length of `7` (the default).

For example, to replicate this:

```sh
$ git clone https://github.com/jelmer/dulwich/ && cd dulwich
$ for tag in $(git tag -l); do git tag --delete ${tag}; done;
$ PYTHONPATH=. python3
Python 3.12.8 (main, Dec 13 2024, 13:19:48) [GCC 14.2.0] on linux
>>> porcelain.describe(".", abbrev=10)
'g77c2d8d'
```

When an annotated tag exists in the history, a correctly-abbreviated commit prefix is produced (along with the expected contextual formatting describing the tag it relates to).

```sh
$ git tag testing a029d7b2cc83c26a53d8b2a24fa12c340fcfac59 --annotate --message "Archaeology Today"
$ PYTHONPATH=. python3
Python 3.12.8 (main, Dec 13 2024, 13:19:48) [GCC 14.2.0] on linux
>>> porcelain.describe(".", abbrev=10)
'testing-4746-g77c2d8d106'
```

Would it be acceptable to offer a patch to pass the `abbrev` parameter to the `find_unique_abbrev` method, even if the `todo` task isn't completed as part of that?

jayaddison:
Resolves #1477.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -2225,7 +2225,7 @@ def describe(repo, abbrev=None):
         if len(sorted_tags) == 0:
             object_id = r[r.head()].id
             if abbrev is not None:
-                return object_id[abbrev_slice].decode("ascii")
+                return object_id.decode("ascii")[abbrev_slice]
             return f"g{find_unique_abbrev(r.object_store, object_id)}"
 
         # We're now 0 commits from the top
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -116,7 +116,7 @@ from .objectspec import (
     parse_tree,
     to_bytes,
 )
-from .pack import ObjectContainer, write_pack_from_container, write_pack_index
+from .pack import write_pack_from_container, write_pack_index
 from .patch import write_tree_diff
 from .protocol import ZERO_SHA, Protocol
 from .refs import (
@@ -2177,14 +2177,14 @@ def ls_files(repo):
         return sorted(r.open_index())
 
 
-def find_unique_abbrev(object_store: ObjectContainer, object_id: str, abbrev: int | None = 7) -> str:
+def find_unique_abbrev(object_store, object_id):
     """For now, just return 7 characters."""
     # TODO(jelmer): Add some logic here to return a number of characters that
     # scales relative with the size of the repository
-    return object_id.decode("ascii")[:abbrev]
+    return object_id.decode("ascii")[:7]
 
 
-def describe(repo, abbrev=7):
+def describe(repo, abbrev=None):
     """Describe the repository version.
 
     Args:
@@ -2222,7 +2222,10 @@ def describe(repo, abbrev=7):
 
         # If there are no tags, return the current commit
         if len(sorted_tags) == 0:
-            return f"g{find_unique_abbrev(r.object_store, r[r.head()].id, abbrev=abbrev)}"
+            object_id = r[r.head()].id
+            if abbrev is not None:
+                return object_id[:abbrev].decode("ascii")
+            return f"g{find_unique_abbrev(r.object_store, object_id)}"
 
         # We're now 0 commits from the top
         commit_count = 0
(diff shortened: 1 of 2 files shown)


dulwich/porcelain.py
```diff
@@ -2177,11 +2177,11 @@ def ls_files(repo):
         return sorted(r.open_index())
 
 
-def find_unique_abbrev(object_store, object_id):
+def find_unique_abbrev(object_store: ObjectContainer, object_id: str, abbrev: int | None = 7) -> str:
```
jelmer, line 2180: find_unique_abbrev is meant to find the unique abbreviation for a commit; passing in the length is counter to that. Perhaps just use ``[:abbrev]`` if abbrev is specified and fall back to find_unique_abbrev otherwise?

jelmer: It should be in find_unique_abbrev, but it should be return a commit that's as long as necessary to prevent ambiguity - not something that the caller determines.

jayaddison: > It should be in find_unique_abbrev, but it should be return a commit that's as long as necessary to prevent ambiguity - not something that the caller determines.

Ok; to check if I interpret that correctly: a future complete implementation of `find_unique_abbrev` would handle ambiguous-commit avoidance, but `describe` can simply return the caller-specified `abbrev` prefix length.

dulwich/porcelain.py
```diff
@@ -2245,13 +2248,13 @@ def describe(repo, abbrev=7):
                         return "{}-{}-g{}".format(
                             tag_name,
                             commit_count,
-                            latest_commit.id.decode("ascii")[:abbrev],
+                            latest_commit.id.decode("ascii")[:abbrev or 7],
```
jayaddison, line 2251: Strictly speaking I suppose this would affect `abbrev=0` callsites; I could adapt the code to use a `kwargs` pattern if it'd be preferable to reduce the behavioural diff.

jayaddison: (or check if `is not None` in a ternary)

jelmer (approved):

dulwich/porcelain.py
jelmer: Lgtm, can you mark the pr as no longer a draft?

jayaddison: Done; thanks!

dulwich/porcelain.py
```diff
@@ -2222,7 +2223,10 @@ def describe(repo, abbrev=7):
 
         # If there are no tags, return the current commit
         if len(sorted_tags) == 0:
-            return f"g{find_unique_abbrev(r.object_store, r[r.head()].id)}"
+            object_id = r[r.head()].id
```
jayaddison, line 2226: Optional extra: we could relocate and re-use the `latest_commit` variable instead of creating `object_id` here.  That might make the code marginally more understandable, too.

dulwich/porcelain.py
```diff
@@ -2222,7 +2223,10 @@ def describe(repo, abbrev=7):
...
+            if abbrev is not None:
+                return object_id[abbrev_slice].decode("ascii")
```
jayaddison, line 2228: In fact: I think there's a bug here; this should be decoding _before_ the substring range-access.  Fixing that up momentarily.

jayaddison: And it's also missing the `g` prefix.  A little more work to do here after all.

jayaddison:
@jelmer I think this is ready for re-review.  Apologies for not noticing the missing `g` prefix in the short-circuit (re)implementation!

jelmer:
@jayaddison changes look good, but please run "ruff format"

jelmer (approved):

jayaddison:
Thanks @jelmer!

PR #1488, opened 2025-02-02
add support for the git option "index.skipHash"

rggjan:
Here is the comparable implementation in another git library for adding support for the same feature: https://github.com/libgit2/libgit2/pull/6738/files


jelmer (changes requested):
Shouldn't this be checking the configuration flag?

rggjan:
> Shouldn't this be checking the configuration flag?

In libgit, they also don’t seem to check the configuration flig: https://github.com/libgit2/libgit2/pull/6738/files

It seems to me that the checksum is all 0 in the index if the client used to write that index had `skipHash` enabled, which is probaby independent of the flags set at the time when it is read by dulwich?

dulwich/pack.py
```diff
@@ -1605,9 +1605,13 @@ def read(self, num=None):
...
         return data
 
-    def check_sha(self) -> None:
+    def check_sha(self, allow_empty: bool = False) -> None:
         stored = self.f.read(20)
-        if stored != self.sha1.digest():
+        # If git option index.skipHash is set the index will be empty
+        if stored != self.sha1.digest() and (
+            not allow_empty
+            or sha_to_hex(stored) != b"0000000000000000000000000000000000000000"
```
jelmer, line 1613: This should probably use the ZERO_SHA constant

jelmer:
> > Shouldn't this be checking the configuration flag?
> 
> In libgit, they also don’t seem to check the configuration flig: https://github.com/libgit2/libgit2/pull/6738/files
> 
> It seems to me that the checksum is all 0 in the index if the client used to write that index had `skipHash` enabled, which is probaby independent of the flags set at the time when it is read by dulwich?

They do set skipHash in the tests though: https://github.com/libgit2/libgit2/blob/913d4ea251366f006a5ced6ed1804acf973a6ce6/tests/resources/status_skiphash/.gitted/config#L8

jelmer:
Actually, I guess this is okay - but can you perhaps add support for skipHash on the creation side? We've been trying to remove the testdata files, so I'd rather not add any more.

PR #1495, opened 2025-02-12
Skip-worktree support, sparse checkout impl

lmmx:
This PR adds and refines functionality around sparse checkout, including:

- **Skip-worktree bit support** in `IndexEntry`, preserving `extended_flags` so that excluded files actually stay out of the working tree.  
- **Updates to `write_index`** so that index entries with extended flags are written as version 3 or higher.  
- **New TDD tests** in `test_porcelain.py` for cone-mode sparse checkout (currently skipped, as cone mode is not fully implemented yet, but the tests are in place for future development). I did my best but I decided to submit now and not potentially spoil it. (I can take these out but I am confident I'll be able to do cone mode with a bit more thought) 
- **Amended tests** in `test_index.py`
- **Local modifications checks** in the porcelain `sparse_checkout` function, so we can raise or force-remove on excluded paths.

All tests now pass (aside from the intentionally skipped cone-mode ones). I ran `make check` and hope this can land.


lmmx:
Removed the cone mode TDDs (doesn't make sense to introduce them prematurely) and ran the same tests the CI runs (I thought it was just `make check`!)

```sh
make check
python -m mypy dulwich
python -m ruff check .
python -m ruff format check .
```

Should go green now :crossed_fingers:

dulwich/porcelain.py
```diff
@@ -2098,6 +2099,123 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
...
+    Raises:
+      CheckoutError: If local modifications would be lost (force=False)
+      Error: If no patterns can be determined or an I/O error occurs
+    """
+    repo = Repo(repo) if not isinstance(repo, Repo) else repo
+
+    info_dir = os.path.join(repo.path, ".git", "info")
+    if not os.path.exists(info_dir):
+        os.makedirs(info_dir)
+    sparse_file = os.path.join(info_dir, "sparse-checkout")
```
jelmer, line 2124: since these are internal repository details, they should really be handled by the Repo class rather than porcelain.

dulwich/porcelain.py
```diff
@@ -2098,6 +2099,123 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
                 dir_path = os.path.dirname(dir_path)
 
 
+# --- Sparse Checkout Commands ---
+
+
+def sparse_checkout(repo, patterns=None, force=False):
+    """Perform a sparse checkout by updating the index and working tree
+    based on inclusion/exclusion patterns, using skip-worktree bits in
+    the index's extended_flags field (as supported by patched Dulwich).
```
jelmer, line 2108: not sure what you mean by patched Dulwich?

uv.lock
```diff
@@ -0,0 +1,370 @@
+version = 1
```
jelmer, line 1: please don't add uv.lock - at least not in this PR

jelmer (changes requested):

jelmer:
Thanks for working on this - overall looks great, but some minor comments inline and one larger one about placement of the code in porcelain.py.

tests/test_porcelain.py
```diff
@@ -3679,3 +3683,161 @@ def test_for_each_ref_pattern(self) -> None:
...
+
+            for filename in files:
+                file_rel = os.path.relpath(os.path.join(root, filename), self.repo_path)
+                found_files.add(file_rel)
+        return found_files
+
+    #
+    # -- Test Cases --
+    #
+    def test_TC_SC_001_only_included_paths_appear_in_wtree(self):
```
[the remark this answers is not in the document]
lmmx: dropped :ballot_box_with_check:

uv.lock
lmmx: removed :ballot_box_with_check:

dulwich/porcelain.py
lmmx: sorry, that was intermediate notes to self about the separate work with the `extended_flags` in the index, rewritten more clearly and in proper D205 format :ballot_box_with_check:

dulwich/porcelain.py
lmmx: Yeah I think `get`/`set` + `_sparse_checkout_patterns` would be a nice way to abstract the following if/else block too, so this would become internal to the Repo

jelmer:
Changes look good; please take a look at the ruff failures

jelmer (approved):

jelmer:
Thank you :)

Issue #405, opened 2016-01-14
Sparse Checkout


meatballs:
Implement the sparse checkout functionality described at:
http://schacon.github.io/git/git-read-tree.html#_sparse_checkout

i.e. if config.core.sparsecheckout = true, then only checkout files which match the patterns listed in the file .git/info/sparse-checkout

meatballs:
How about the following:

 * Add list parameters ```match_patterns``` and ```ignore_patterns``` to ```index.build_index_from_tree``` which can be used to pass unix style file pattern matches. The parameters would have a default value of ```None```

 * Create a new file path validator which uses the fnmatch module to match against the pattern parameters

meatballs:
I'm happy to pick this one up once we agree on the way forward!

jelmer:
On Fri, Jan 15, 2016 at 01:57:27AM -0800, Owen Campbell wrote:
> How about the following:
> 
>  * Add list parameters ```match_patterns``` and ```ignore_patterns``` to ```index.build_index_from_tree``` which can be used to pass unix style file pattern matches. The parameters would have a default value of ```None```
The syntax of these is the same as for gitignore files. I would suggest passing in a Matcher object which can be constructed from a .gitignore or .git/info/sparse-checkout file.

>  * Create a new file path validator which uses the fnmatch module to match against the pattern parameters
This is essentially the matcher mentioned above. 

You'll probably also need to touch the index object to allow setting the skip-workingtree bit.

Cheers,

Jelmer

jelmer:
On Mon, Jan 18, 2016 at 12:51:14PM +0000, Jelmer Vernooij wrote:
> On Fri, Jan 15, 2016 at 01:57:27AM -0800, Owen Campbell wrote:
> > How about the following:
> > 
> >  * Add list parameters ```match_patterns``` and ```ignore_patterns``` to ```index.build_index_from_tree``` which can be used to pass unix style file pattern matches. The parameters would have a default value of ```None```
> The syntax of these is the same as for gitignore files. I would suggest passing in a Matcher object which can be constructed from a .gitignore or .git/info/sparse-checkout file.
> 
> >  * Create a new file path validator which uses the fnmatch module to match against the pattern parameters
> This is essentially the matcher mentioned above. 
> 
> You'll probably also need to touch the index object to allow setting the skip-workingtree bit.
See also http://schacon.github.io/git/git-read-tree.html#_sparse_checkout

meatballs:
[...]

PR #1497, opened 2025-02-17
feat(cone-mode): sparse checkout cone mode support

lmmx:
Following up on my previous submission #1495 (from #405) which laid the groundwork for cone mode I wanted to complete the task and now have got the test cases all working. 

It's once again pretty late by the time I'm finishing up, so I'll review with fresh eyes but I've at least ran all the linters this time :-) Tests pass and are named normally.


lmmx:
Ooh OK docstring parse errors, I'll return tomorrow :pray:

dulwich/porcelain.py
```diff
@@ -2100,113 +2101,167 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
...
+    Args:
+      repo: Path to the repository or a Repo object.
+      dirs: List of directory names to add to the sparse-checkout.
+      force: Whether to forcibly discard local modifications (default False).
+
+    Returns:
+      None
+    """
+    if not isinstance(repo, Repo):
+        repo = Repo(repo)
```
jelmer, line 2249: Please use the common patterns used elsewhere in the file here: e.g. "with open_repo_closing(...)"

dulwich/porcelain.py
```diff
@@ -2100,113 +2101,167 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
...
+
+    sp_path = os.path.join(repo.path, ".git", "info", "sparse-checkout")
```
jelmer, line 2251: Please move the implementation-specific details to Repo

dulwich/porcelain.py
```diff
@@ -2100,113 +2101,167 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
...
+    config.set((b"core",), b"sparseCheckout", b"true")
+    config.set((b"core",), b"sparseCheckoutCone", b"true")
+    config.write_to_path()
+
+    # Initial lines: include top-level files, then exclude all subdirectories
+    new_patterns = ["/*", "!/*/"]
+
+    # For each directory to include, add an exclusion-style line that "undoes"
+    # the prior 'exclude' and re-includes that directory (and everything under it).
+    for d in dirs:
```
jelmer, line 2224: Move this code to a separate function in .sparse_patterns?

dulwich/porcelain.py
```diff
@@ -2100,113 +2101,167 @@ def checkout_branch(repo, target: Union[bytes, str], force: bool = False) -> Non
...
+        repo = Repo(repo)
+
+    sp_path = os.path.join(repo.path, ".git", "info", "sparse-checkout")
+    if not os.path.exists(sp_path):
+        cone_mode_init(repo)
+
+    with open(sp_path) as f:
+        existing = [ln.strip("\n") for ln in f if ln.strip()]
+
+    for d in dirs:
```
jelmer, line 2258: Move this to a separate function is .sparse_patterns?

dulwich/sparse_patterns.py
```diff
@@ -0,0 +1,462 @@
...
+    )
+
+    def __init__(self, is_neg, is_dir_only, anchored, regex):
+        self.is_neg = is_neg
+        self.is_dir_only = is_dir_only
+        self.anchored = anchored
+        self.regex = regex
+
+
+def fnmatch_to_regex(pat):
```
jelmer, line 227: Isn't this the same as ignore patterns? Can we reuse that code?

dulwich/sparse_patterns.py
```diff
@@ -0,0 +1,462 @@
...
+                try:
+                    blob = repo.object_store[entry.sha]
+                except KeyError:
+                    raise Error(f"Blob {entry.sha} not found for {path_bytes}.")
+                ensure_dir_exists(os.path.dirname(full_path))
+                with open(full_path, "wb") as f:
+                    f.write(blob.data)
+
+
+class ParsedPattern:
```
jelmer, line 203: Please reuse the existing gitignore pattern code.

dulwich/sparse_patterns.py
```diff
@@ -0,0 +1,462 @@
+# sparse_patterns.py -- Sparse checkout pattern handling.
```
jelmer, line 1: Please add tests for each of the functions/classes in this file.

jelmer (changes requested):
Thanks, great to see cone support!

Left a bunch of comments inline.

lmmx:
Super, thanks will address ! :smile:

dulwich/porcelain.py
lmmx: 93094f04 :heavy_check_mark:

dulwich/sparse_patterns.py
lmmx: So the gitignore style patterns (a.k.a. "full") are the simple ones, we want to also support "cone mode" patterns, so no, you're right though we don't need it all, some of this is getting deleted now...

dulwich/porcelain.py
lmmx: Yep, there was a lot of refactoring to be done for this to be porcelain tier code, thanks for the nudge, I've made dirs a porcelain-module concept and patterns are prepared/retrieved in the repo method.

dulwich/porcelain.py
lmmx: yes this was reused, refactored and moved out of tidy porcelain code

dulwich/sparse_patterns.py
lmmx: git docs on full vs. cone mode patterns are here:
- https://git-scm.com/docs/git-sparse-checkout#_internalsfull_pattern_set
- https://git-scm.com/docs/git-sparse-checkout#_internalscone_pattern_set

dulwich/sparse_patterns.py
lmmx: this bit was draft code that wasn't being called sorry, should've been cut before pushing

dulwich/sparse_patterns.py
lmmx: yep, this one is nice and simple to test as the file is just a readable list of patterns, will do tests next

dulwich/porcelain.py
lmmx: yep, there was actually already a method for accessing this path, reused that existing one

lmmx:
Whew, okay all of those comments were tidying/refactoring tasks that are all done now. Also I'd somehow managed to misread the meaning of "un-exclude", what the docs mean there is that there is always the following base paths:

```
/*
!/*/
```

i.e. "include `/*`" (include all top level files) and "exclude `/*/` (exclude all top level dirs)

In other words it says "don't include any dirs until told to do so"

So "un-exclude" is just a confusing way to say "include directories when told to do so".

To be super clear, as far as I know you never put excludes on except in that one line as a special case. This is why cone mode is *simpler*, because you are only dealing with the one type of pattern.

That should make it simple to test :-)

lmmx:
existing tests are now all passing again

lmmx:
Tests are almost all passing, will revisit and see if there are logic bugs or test bugs :bug: :hammer:

jelmer:
🎉 Looks great overall - I'll take a closer look when I have some more time to do a full review.

lmmx:
Whew okay all tests passing now, so that gives me confidence this is a functioning implementation of cone mode 😄 

The one last tricky bit was ensuring that “patterns suffixed with a slash (meaning they are dirs) must only match directories but can match files in subpaths of the directory pattern”. My initial implementation had just said “let’s just treat these as only the directory and not deal with the subpaths for now”.

I’ll be able to go over this again and refactor the logic here now that the functionality is pinned down by tests.

lmmx:
I’d also point out (and have just edited the docstrings accordingly) that this is (still!) not the full implementation of git’s sparse checkout with all features, namely it is only the recursive cone mode, not the parent cone mode. I.e. when you say “add directory foo to the cone” you get all subdirs of foo, not just the top one. The recursive mode would be more work and I think it’s a perfectly fine MVP to introduce just this, and ensure we’re happy with this before any further round. PRs typically go wrong when you change course midway and this is looking stable here.

dulwich/sparse_patterns.py
```diff
@@ -0,0 +1,342 @@
...
+            elif path_str == pattern:
+                matched = True
+            elif path_str.startswith(pattern + "/"):
+                matched = True
+            else:
+                matched = False
+        elif not matched:
+            # Not anchored: we can do a simple wildcard match or a substring match.
+            # For simplicity, let's use Python's fnmatch:
+            matched = fnmatch(path_str, pattern) or fnmatch(path_str, f"*/{pattern}")
```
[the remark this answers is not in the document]
lmmx: ahhh have I reinvented a wheel :man_facepalming: I will try to edit it to use that then, my bad!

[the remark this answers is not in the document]
jelmer: Thanks! Otherwise LGTM, thanks for adding tests and refactoring.

(I appreciate this isn't complete for sparse checkout support, but it's a big step towards fully supporting it so that's great!)

jelmer:
Please also take a look at the failing tests - you probably want to compare sets rather than sequences?

jelmer (approved):

PR #1513, opened 2025-03-05
porcelain: permit printing commits that contain no commit message

### LINKED ISSUE #1510: bug (3 of 3): empty commit message causes CLI log output to fail
The `dulwich log` CLI command does not seem to handle empty commit messages gracefully.  I think this is because there is an absence of a conditional check for empty commit messages here:

https://github.com/jelmer/dulwich/blob/6a52ce9ede3fb6681cfc01be943d2858c346cc3a/dulwich/porcelain.py#L771-L773

(I think it would be best not to emit the leading/pre-message empty newline if the commit message is entirely absent -- because that makes the output more concise and is a hint to the reader that no commit message exists at all for the relevant commit)

jayaddison:
Adds logic to detect an empty commit message during `print_commit`, preventing failures during attempts to decode `None`.

Resolves #1510.

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -768,9 +768,10 @@ def print_commit(commit, decode, outstream=sys.stdout) -> None:
     time_str = time.strftime("%a %b %d %Y %H:%M:%S", time_tuple)
     timezone_str = format_timezone(commit.author_timezone).decode("ascii")
     outstream.write("Date:   " + time_str + " " + timezone_str + "\n")
-    outstream.write("\n")
-    outstream.write(decode(commit.message) + "\n")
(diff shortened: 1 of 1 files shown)


jayaddison:
TODO: add example test coverage

jelmer:
Any chance you could add a test for this in tests/test_porcelain.py?

jayaddison:
> Any chance you could add a test for this in tests/test_porcelain.py?

Good shout, yep - will do, likely within the `LogTests` class.  Hopefully it won't take too long to write; I'll plan to get to that within the next day or two.

jelmer:
Added a test in #1515

PR #1515, opened 2025-03-06
Handle empty commit messages in porcelain.log

### LINKED ISSUE #1510: bug (3 of 3): empty commit message causes CLI log output to fail

jelmer:
Basically #1513 with tests.

Fixes #1510

### DIFF
--- a/tests/test_porcelain.py
+++ b/tests/test_porcelain.py
@@ -1082,9 +1082,39 @@ class LogTests(PorcelainTestCase):
             self.repo.object_store, [[1], [2, 1], [3, 1, 2]]
         )
         self.repo.refs[b"HEAD"] = c3.id
+        self.maxDiff = None
         outstream = StringIO()
         porcelain.log(self.repo.path, outstream=outstream)
-        self.assertEqual(3, outstream.getvalue().count("-" * 50))
+        self.assertEqual(
+            outstream.getvalue(),
+            """\
+--------------------------------------------------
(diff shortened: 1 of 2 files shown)


tests/test_porcelain.py
```diff
@@ -1095,6 +1125,28 @@ def test_max_entries(self) -> None:
         porcelain.log(self.repo.path, outstream=outstream, max_entries=1)
         self.assertEqual(1, outstream.getvalue().count("-" * 50))
 
+    def test_no_revisions(self) -> None:
+        outstream = StringIO()
+        porcelain.log(self.repo.path, outstream=outstream)
+        self.assertEqual("", outstream.getvalue())
+
+    def test_empty_message(self) -> None:
+        c1 = make_commit(message="")
```
jayaddison, line 1134: Thanks; I'm not sure why I didn't think to try using an empty string here.  I can confirm that this test fails in the expected manner without the fix applied.

jayaddison:
Thanks @jelmer!

PR #1575, opened 2025-06-06
Fix gitignore pattern matching for directory negation patterns

### LINKED ISSUE #1203: ignore: negation with single `*` wildcard is not applied to directory paths
Related: https://github.com/iterative/dvc/issues/6995

Given this directory structure:
```
$ tree .
.
└── data
    ├── test
    └── test.dvc

2 directories, 2 files
```

And this gitignore:
```
$ cat .gitignore
data/**
!data/*/
!*.dvc
```

CLI Git only ignores `data/test.dvc`, but dulwich treats `data/` as ignored:
```
$ git check-ignore data/ data/test.dvc data/test
data/test
$ python -m dulwich check-ignore data/ data/test.dvc data/test
data/
data/test
```

If you change negation pattern to use a `**` wildcard dulwich works as expected:
```
$ cat .gitignore
data/**
!data/**/
!*.dvc
$ git check-ignore data/ data/test.dvc data/test
data/test
$ python -m dulwich check-ignore data/ data/test.dvc data/test
data/test
```

jelmer:
Patterns like ``!data/*/`` now correctly unignore direct subdirectories while still ignoring files in the parent directory, matching Git's behavior. The ``is_ignored()`` method now documents that directory paths should end with ``/`` for consistent behavior.

Also, support quote_path flag.

Fixes #1203

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,13 @@
 0.22.9	UNRELEASED
 
+ * Fix gitignore pattern matching for directory negation patterns. Patterns like
+   ``!data/*/`` now correctly unignore direct subdirectories while still ignoring
+   files in the parent directory, matching Git's behavior. The ``is_ignored()`` method
+   now documents that directory paths should end with ``/`` for consistent behavior.
+   (Jelmer Vernooĳ, #1203)
+
+ * Support quote_path flag for ignore checking. (Jelmer Vernooĳ)
+
  * Add support for Git's ``feature.manyFiles`` configuration and index version 4.
    This enables faster Git operations in large repositories through path prefix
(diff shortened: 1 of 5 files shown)


ncoghlan:
Thanks for all your work on `dulwich`!

I came across this PR while checking the details of exactly how `check_ignore` works, and wondered if the change in default quoting behaviour when moving from 0.22.8 to 0.22.9 was intentional or not.

I'm using `check_ignore` as the basis of a `shutil.copytree` ignore filter function, so for my use case `quote_path=False` is the desired behaviour, but I don't know if my usage is considered an abuse of the API or not (since I'm definitely not debugging `.gitignore` files as the docstring suggests)

PR #1640, opened 2025-06-27
fix: handle missing sshCommand in config

skshetry:
Adds a check for KeyError when accessing the sshCommand in the config, which may or may not be present.

Following snippet of code was failing for us after 0.23 release (due to #1550).

```
from dulwich.porcelain import clone

clone("git@github.com:iterative/example-get-started.git")
```

Closes https://github.com/iterative/dvc/issues/10788.


jelmer (approved):

skshetry:
Hey @jelmer, could you please make a release with this fix when you have a moment? This is breaking dvc. 🙂

Issue #1561, opened 2025-05-30
shallow clone from a google server


her01n:
Shallow clone from googleserver.com url fails with an error.

Run

```
from dulwich import porcelain
porcelain.clone("https://go.googlesource.com/net", depth = 1)
```

The function raises an error: unknown command b'acknowledgments\n',
it should finish without an error.

```
Traceback (most recent call last):
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/client.py", line 478, in _read_shallow_updates
    cmd, sha = pkt.split(b" ", 1)
    ^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    porcelain.clone("https://go.googlesource.com/net", depth = 1)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/porcelain.py", line 566, in clone
    return client.clone(
           ~~~~~~~~~~~~^
        path,
        ^^^^^
    ...<9 lines>...
        protocol_version=protocol_version,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/client.py", line 874, in clone
    result = self.fetch(
        path,
    ...<5 lines>...
        protocol_version=protocol_version,
    )
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/client.py", line 981, in fetch
    result = self.fetch_pack(
        path,
    ...<7 lines>...
        protocol_version=protocol_version,
    )
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/client.py", line 2737, in fetch_pack
    (new_shallow, new_unshallow) = _read_shallow_updates(
                                   ~~~~~~~~~~~~~~~~~~~~~^
        resp_proto.read_pkt_seq()
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/heroin/Projekty/veteran/dulwich-source/dulwich/client.py", line 480, in _read_shallow_updates
    raise GitProtocolError(f"unknown command {pkt}")
[...]

PR #1681, opened 2025-07-08
Protov2 deepen

stspdotname:
These changes fix issue #1561


jelmer:
I tried writing a test myself, but found it hard to do so in a way that is actually meaningful - I don't really want to hardcode the entire conversation with the client lest it make it harder to change things in the future. So I'll just merge this as-is for now - and fingers crossed this won't regress in the future..

PR #1761, opened 2025-08-11
Add `object_store.iter_commit_contents()` and docs

### LINKED ISSUE #1740: Easier helper to get files out of a tree without creating a worktree?
I'm using Dulwich in an automated deployment script to get a clean copy of a repo's files in an arbitrary location in the filesystem (i.e. essentially what `bzr export` does). At the moment, the code I'm using is morally equivalent to this:

```python
from dulwich import porcelain as git

def checkout_git(repo_path: str, dst: str, folders: list[str] = [], files: list[str] = []):
    """Make a new, clean checkout of specified repo to dst. 
Worktree must be clean and HEAD must be at the desired commit"""
    repo = git.open_repo(repo_path)
    status = git.status(repo, untracked_files="no")
    clean = not list(chain(*status.staged.values(), status.unstaged))
    if not clean:
        raise CheckoutError("Git working directory has uncommitted changes, aborting checkout")

    dst = Path(dst)
    tmp_dst = dst.with_name(f"{dst.name}-tmp")
    worktree = None
    try:
        worktree = repo.worktrees.add(tmp_dst)
        dst.parent.mkdir(parents=True, exist_ok=True)
        # shutil.move() changes semantics if dst is an existing directory
        shutil.move(tmp_dst, dst.parent if dst.is_dir() else dst)
    except Exception as exc:
        raise CheckoutError(
            f'Unable to checkout source tree from Git: {type(exc).__name__}: {exc}'
        ) from exc
    finally:
        if worktree and Path(worktree.path).exists():
            repo.worktrees.remove(worktree.path)
```

This works, but registering a new worktree in the repo under a temporary path just to move it away feels pretty hacky and not as robust as I'd like. It also relies on the fact that `remove_worktree()` currently just silently skips removing the files if it the worktree path doesn't exit. 

Ideally, there would be a separate helper that takes care of the low-level plumbing involved in getting files out of a tree properly, but isn't inherently ties to worktrees. If I understand the implementation correctly, it would also need a helper that does the bulk of what `build_index_from_tree()` does, but without being tied to an index. It seems to me that it should be easy to make `add_worktree()` and `build_index_from_tree()` use those helpers and add their worktree and index logic on top, but going the other way around and taking that logic out isn't possible.

I don't know your goals with Dulwich, but to me at least, it makes sense to make it a toolkit that's easier to use, more robust, and more composable in Python code than `subprocess.run(["git", ...])`, and which also doesn't make me feel dirty after using it. Having composable helpers that make common tasks easy would be very much appreciated :)

mathrick:
Closes #1740.

I also added a new section to the documentation, "recipes", and added recipes demonstrating the use of `iter_commit_contents()` and `tree_lookup_path()` to retrieve files from the repository.


docs/recipes/index.txt
```diff
@@ -0,0 +1,99 @@
...
+------------------------------------------------------------------------------------
+
+:py:func:`dulwich.porcelain.checkout` is a very high-level function
+that operates on the working tree and behaves very similar to the
+``git checkout`` command. It packages a lot of functionality into a
+single command, just as Git's porcelain does, which is useful when
+matching Git's CLI is the goal, but might be less desireable for
+programmatic access to a repository's contents.
+
+Retrieve a single file's contents at a particular commit
```
jelmer, line 24: This is incomplete, since unlike checkout you'll be ignoring things like filters.

dulwich/tests/test_object_store.py
```diff
@@ -243,6 +244,59 @@ def test_iter_tree_contents_include_trees(self) -> None:
         actual = iter_tree_contents(self.store, tree_id, include_trees=True)
         self.assertEqual(expected, list(actual))
 
+    def test_iter_commit_contents(self) -> None:
```
jelmer, line 247: Please add a separate TestCase subclass for this.

dulwich/tests/test_object_store.py
```diff
@@ -243,6 +244,59 @@ def test_iter_tree_contents_include_trees(self) -> None:
...
+        ]
+
+        include3 = ["dir"]
+        expected3 = [
+            (b"dir/baz", b"baz"),
+            (b"dir/subdir/foo", b"subfoo"),
+            (b"dir/subdir/bar", b"subbar"),
+            (b"dir/subdir/baz", b"subbaz"),
+        ]
+
```
jelmer, line 288: perhaps add a testcase for providing overlapping included paths?

docs/recipes/index.txt
```diff
@@ -0,0 +1,99 @@
...
+    encoding = commit.encoding or "utf-8"
+
+    # Scan the repo at current HEAD. Retrieve all files marked as
+    # executable under bin/ and write them to disk
+    for entry in iter_commit_contents(repo, commit.id, include=[b"bin"]):
+        if entry.mode & stat.S_IXUSR:
+            # Strip the leading bin/ from returned paths, write to
+            # current directory
+            path = Path(entry.path.decode(encoding)).relative_to("bin/")
+            build_file_from_blob(
```
jelmer, line 95: This is going to ignore filters/eol/etc as well.

docs/recipes/index.txt
```diff
@@ -0,0 +1,99 @@
+.. _recipes:
```
jelmer, line 1: Perhaps move this to a separate PR? It seems mostly independent of the other change.

dulwich/tests/test_object_store.py
```diff
@@ -243,6 +244,59 @@ def test_iter_tree_contents_include_trees(self) -> None:
...
+        ]
+        blobs = {contents: make_object(Blob, data=contents) for path, contents in files}
+        for blob in blobs.values():
+            self.store.add_object(blob)
+        commit = make_commit(tree=commit_tree(
+            self.store,
+            [(path, blobs[contents].id, 0o100644) for path, contents in files]
+        ))
+        self.store.add_object(commit)
+
```
jelmer, line 266: please add a test for include=None

jelmer (changes requested):

dulwich/tests/test_object_store.py
mathrick: Done

jelmer:
Changes LGTM, but please take a look at the typing failures.

jelmer (approved):

dulwich/object_store.py
```diff
@@ -2674,6 +2675,57 @@ def iter_tree_contents(
...
+    This is a wrapper around iter_tree_contents() and
+    tree_lookup_path() to simplify the common task of getting the
+    contest of a repo at a particular commit. See also
+    dulwich.index.build_file_from_blob() for writing individual files
+    to disk.
+
+    Args:
+      store: Object store to get trees from
+      commit: Commit object, or SHA1 of a commit
+      include: if provided, only the entires whose paths are in the
```
jelmer, line 2695: codespell wants this to be entries rather than entires :)

jelmer, line 2695:
```suggestion
      include: if provided, only the entries whose paths are in the
```

PR #1767, opened 2025-08-13
Add recipes to documentation

mathrick:
Separated from #1761. I added code to handle blob normalisation to address your comment. This should probably be merged in after the other PR lands, since it makes reference to `iter_commit_contents()`.


jelmer (approved):

jelmer:
"recipes" seems a bit generic - creating a commit or closing a repository would fall under this as well presumably. But happy to merge this as-is and we can reorganise it later.

mathrick:
> "recipes" seems a bit generic - creating a commit or closing a repository would fall under this as well presumably. But happy to merge this as-is and we can reorganise it later.

Oh yes, that's very much the idea :) I meant it to be a generic place where answers to "how do I do `$x` with Dulwich?" can be collected for common values of `$x`. I just seeded it with this specific recipe because it's what I was trying to solve myself.

PR #1768, opened 2025-08-15
Improve new contributor experience

mathrick:
This fixes a couple frustrations I ran into in my recent PRs:

- Adds a single `make` target to run all style checks that will be required for a PR
- Adds the missing dev dependency on `codespell`
- Updates `CONTRIBUTING.rst` to mention how to install Dulwich in dev mode, the required checks for PRs, and fixes an apparent typo in the coding style section (which seems to be the opposite of what the code actually uses)
- Fixes code blocks in `CONTRIBUTING.rst`  for proper rendering on GitHub

You might also want to consider using the `make` targets inside the PR test workflow instead of inlining their corresponding commands, to guarantee they won't drift apart, but that's a larger decision and I leave it up to you.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -42,8 +42,7 @@ There are also other, optional dependencies which are needed to run the full tes
 implement optional features, and provide the full typing information. They are however not
 strictly necessary; the above is sufficient to start developing and have your PR pass the
 tests in most cases. Please consult the ``[project.optional-dependencies]`` section in
-``pyproject.toml`` and the various targets in ``Makefile`` for a full list of optional
-dependencies.
+``pyproject.toml``.
 
 Coding style
 ------------
@@ -53,7 +52,9 @@ and typing checks" below for details on running style checkers.
 
 Public methods, functions and classes should all have doc strings. Please use
 Google style docstrings to document parameters and return values.
-You can generate the documentation by running "make doc".
+You can generate the documentation by running ``pydoctor --docformat=google dulwich``
+from the root of the repository, and then opening
+``apidocs/index.html`` in your web browser.
 
 String Types
 ~~~~~~~~~~~~
@@ -98,39 +99,49 @@ dulwich package. This will ensure that the deprecation is handled correctly:
 
 Running the tests
 -----------------
-To run the testsuite, you should be able to simply run "make check". This
-will run the tests using unittest.
+To run the testsuite, you should be able to run ``dulwich.tests.test_suite``.
+This will run the tests using unittest.
 
 .. code:: console
 
-   $ make check
+   $ python -m unittest dulwich.tests.test_suite
 
 The compatibility tests that verify Dulwich behaves in a way that is compatible
 with C Git are the slowest, so you may want to avoid them while developing:
 
 .. code:: console
 
-   $ make check-nocompat
+   $ python -m unittest dulwich.tests.nocompat_test_suite
 
 testr and tox configuration is also present.
 
 Style and typing checks
 -----------------------
 
-Use ``make all-style`` to run all style-related checks. Use ``make typing`` for typing
-checks. Those checks are *mandatory*, a PR will not pass tests and will not be merged if
+Several static analysis tools are used to ensure code quality and consistency.
+
+* Use ``ruff check`` to run all style-related checks.
+* Use ``ruff format --check`` to check code formatting.
+* Use ``mypy dulwich`` for typing checks.
+* Use ``codespell`` to check for common misspellings.
+
+Those checks are *mandatory*, a PR will not pass tests and will not be merged if
 they aren't successful.
 
 .. code:: console
 
-   $ make all-style
-   $ make typing
+   $ ruff check
+   $ ruff format --check
+   $ mypy dulwich
+   $ codespell
 
-Some of these checks will modify the code to fix issues encountered, remember to commit
-these changes afterwards!
+In some cases you can automatically fix issues found by these tools. To do so, you can run:
 
-There are also individual ``make`` targets to run just a single style check, see the
-definition of ``all-style`` in ``Makefile``.
+.. code:: console
+
+   $ ruff check --fix  # or pass --unsafe-fixes to apply more aggressive fixes
+   $ ruff format
+   $ codespell --config .codespellrc -w
 
 Merge requests
 --------------
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -6,12 +6,50 @@ Where possible include updates to NEWS along with your improvements.
 
 New functionality and bug fixes should be accompanied by matching unit tests.
 
+Installing development dependencies
+-----------------------------------
+
+Contributing to Dulwich requires several more dependencies than are required to install
+the base package; they are used to run tests and various other checks.
+
+First, make sure your system has the Rust compiler and Cargo (the Rust package manager)
+installed. As this is a system-level requirement and not a Python library, the right way
+to install it depends on your platform. Please consult the `Rust documentation
+<https://www.rust-lang.org/learn/get-started>`__ to find out more.
+
+Next, you will need to set up your Python environment for Dulwich. An easy way to get
+started is to install the checked out Dulwich package in editable mode with ``dev``
+extras, preferably in a new virtual environment:
+
+.. code:: console
+
+   $ cd ~/path/to/checkouts/dulwich
+   # Create and activate a virtual environment via your favourite method, such as pyenv,
+   # uv, built-in venv module...
+   $ python -m venv .venv && . .venv/bin/activate
+   # Now install Dulwich and the required dependencies
+   $ pip install -e ".[dev]"
+
+This will ensure the tools needed to test your changes are installed. It is not necessary
+to install Dulwich in editable mode (``-e``), but doing so is convenient for development,
+as code changes will be visible immediately, without requiring a reinstall (although any
+running Python processes will need to be reloaded to see the updated module
+definitions). Editable mode only applies to Python code; if you modify any of the Rust
+extension code, you will need to reinstall the package for the extensions to be
+recompiled.
+
+There are also other, optional dependencies which are needed to run the full test suite,
+implement optional features, and provide the full typing information. They are however not
+strictly necessary; the above is sufficient to start developing and have your PR pass the
+tests in most cases. Please consult the ``[project.optional-dependencies]`` section in
+``pyproject.toml`` and the various targets in ``Makefile`` for a full list of optional
+dependencies.
+
 Coding style
 ------------
-Where possible, please follow PEP8 with regard to coding style. Run ruff.
-
-Furthermore, triple-quotes should always be """, single quotes are ' unless
-using " would result in less escaping within the string.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.rst
```diff
@@ -6,12 +6,29 @@ Where possible include updates to NEWS along with your improvements.
...
+Installing development dependencies
+-----------------------------------
+
+Install the checked out Dulwich package in editable mode with ``dev`` extras:
+
+.. code:: console
+
+   $ cd ~/path/to/checkouts/dulwich
+   # Make sure to run this in a virtual environment
+   $ pip install -e ".[dev]"
```
jelmer, line 18: This seems a little prescriptive - I've personally never used -e.

CONTRIBUTING.rst
```diff
@@ -6,12 +6,29 @@ Where possible include updates to NEWS along with your improvements.
...
+
+This will ensure all the tools needed to test the changes are installed. Since Dulwich is
+installed in editable mode, code changes will be visible immediately, without requiring a
+reinstall (but any running Python processes will need to be reloaded to see the updated
+module definitions).
+
 Coding style
 ------------
-Where possible, please follow PEP8 with regard to coding style. Run ruff.
```
jelmer, line 11: This should just recommend running ruff without going into details.

Makefile
```diff
@@ -9,6 +9,8 @@ DESTDIR=/
 
 all: build
 
+all-style:: fix reformat codespell
```
jelmer, line 12: style intentionally only reports on style, it doesn't make any changes. So it seems that all-style would.

fix makes actual code changes, it's not just a style thing.

jelmer (changes requested):

CONTRIBUTING.rst
mathrick: Not sure what you mean, which part are you referring to? If the "PEP8" part, that's just the original contents of the file, I didn't change it :). I can update it however.

CONTRIBUTING.rst
mathrick: Sure, I can make it less prescriptive and more of a "here's an easy way to start".

CONTRIBUTING.rst
mathrick: I adjusted it, does it look better to you?

CONTRIBUTING.rst
mathrick: I adjusted the wording to make it more of general outline, and expanded it to mention Rust / Cargo being required as well.

mathrick:
I changed it to address your feedback. 

I also added `setuptools` and `setuptools-rust` to the `dev` requirements. I actually forgot that it was another snag I ran into immediately after cloning Dulwich, as `make check` requires them by virtue of invoking `setup.py`. Unfortunately, I don't know of a better way of signalling that `build-system.requires` dependencies are also needed here; I'm afraid listing them again is the only option.

mathrick:
Dulwich code itself doesn't need them, but all `make check-*` targets do, as they call `setup.py`, which requires both to be present. When it's through the `build-system.requires`, apparently it's enough to only list `setuptools` and it takes care of installing Rust extensions itself, but when it's `python setup.py`, then the same automatic dependency mechanism doesn't kick in and `setuptools-rust` needs to be listed as well.

Makefile
mathrick: I went ahead and added separate `all-style` and `all-fix`. I also reordered the targets within `Makefile` so they're grouped by area; this doesn't change any of the rules, just makes them easier to scan to find the one you're interested in IMHO.

Makefile
```diff
@@ -42,36 +54,34 @@ check-noextensions:: clean
...
-	rm -f dulwich/*.so
-
 style:
 	$(RUFF) check .
 
+fix:
+	ruff check --fix .
+
+format:
+	ruff format --check .
```
jelmer, line 67: This is a change in behaviour, please just have "make format" do formatting - not just check whether reformatting is necessary.

Makefile
```diff
@@ -7,20 +7,32 @@ COVERAGE = python3-coverage
...
+	coverage coverage-html doc sphinx apidocs
+
 all: build
 
-doc:: sphinx
+all-style:: style format codespell
 
-sphinx::
-	$(MAKE) -C docs html
+all-fix:: fix reformat codespell
```
jelmer, line 21: codespell doesn't make changes so it seems odd to be included here.

I'd also argue that reformat does something fundamentally different from "ruff check --fix" which does more than just change formatting - it can affect behaviour.

Makefile
```diff
@@ -7,20 +7,32 @@ COVERAGE = python3-coverage
...
 	$(SETUP) build_ext -i
 
+clean::
+	$(SETUP) clean --all
+	rm -f dulwich/*.so
+
 install::
 	$(SETUP) install --root="$(DESTDIR)"
 
+check-all: check check-pypy check-noextensions
```
jelmer, line 34: This breaks when run with "make -j". 

"check-all" is a bit confusing name - what is this target really trying to do? "check" for most projects means running the test suite. I'm not sure why this also tests typing.

jelmer (changes requested):

CONTRIBUTING.rst
```diff
@@ -6,12 +6,50 @@ Where possible include updates to NEWS along with your improvements.
...
+dependencies.
+
 Coding style
 ------------
-Where possible, please follow PEP8 with regard to coding style. Run ruff.
+The code follows the PEP8 coding style. There are ``ruff`` rules in place that define the
+exact code style, please run it to make sure your changes are conformant. See also "Style
+and typing checks" below for details on running style checkers.
 
-Furthermore, triple-quotes should always be """, single quotes are ' unless
```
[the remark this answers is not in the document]
mathrick: Sure, I can do that.

Makefile
mathrick: But that's what `reformat` already does? This is a new target that only does checks, but doesn't change anything, to provide parity with other check/fix targets.

Makefile
mathrick: That's news to me, and the [Ruff docs](https://docs.astral.sh/ruff/formatter/) make no mention of that, are you sure? The formatter is supposed to reformat the code only, not change how it parses. You might be thinking of `ruff check --unsafe-fixes`, which does say it could potentially change behaviour, but none of our rules use that.

Makefile
mathrick: I don't know :). This is an existing rule, I just moved it around. I agree that it doesn't seem to have a clear purpose.

mathrick:
> > Dulwich code itself doesn't need them, but all `make check-*` targets do, as they call `setup.py`, which requires both to be present. When it's through the `build-system.requires`, apparently it's enough to only list `setuptools` and it takes care of installing Rust extensions itself, but when it's `python setup.py`, then the same automatic dependency mechanism doesn't kick in and `setuptools-rust` needs to be listed as well.
> 
> This is where it gets a bit muddy though - "pip install -e ." builds the rust bindings so when doing that, you probably don't want to recommend running make check at all, but just invoking the test suite directly (python -m unittest dulwich.tests.test_suite).

Hmm, that's a good point, but I also don't want to make people have to remember that. I'll see if I can make it work for both cases.

Makefile
jelmer: Ruff is not meant to make changes to behaviour but it's possible, especially in corner cases. Unlikely "ruff format" I'd be uncomfortable running it unconditionally, for example.

Makefile
jelmer: Let's just drop it?

Makefile
jelmer: Oh, I see. Having both format and reformat where one makes changes confused me. Don't really want to bikeshed on this and happy to merge as-is, but I don't think it's a great name.

Makefile
mathrick: I really think this would be considered a serious bug if `ruff format` changed any behaviour. IMHO splitting it out into separate rules that apply fixes, as opposed to simply checking, makes it pretty clear that that's their purpose. I don't think forcing people to apply all the mechanical formatting fixes manually to please the tool is an improvement, when the tool itself can apply them. This just extends the rules so that there's a comprehensive set of checks and matching fixes.

Makefile
mathrick: Can do.

Makefile
mathrick: I also added `codespell-fix` to match the rest. Since spelling fixes are not necessarily obvious, I made it ask for confirmation before applying each one.

mathrick:
@jelmer: I added support for detecting editable install. I don't know what workflow you use yourself, so I wasn't able to test it very extensively, let me know if you run into any issues with it. For me it adds a little bit of overhead for TAB-completion, but I found it to be within acceptable limits. My setup is probably the slowest possible case, as I'm using `pyenv` and its shims are definitely not the fastest.

Makefile
```diff
@@ -4,33 +4,45 @@ SETUP = $(PYTHON) setup.py
 TESTRUNNER ?= unittest
 RUNTEST = PYTHONHASHSEED=random PYTHONPATH=$(shell pwd)$(if $(PYTHONPATH),:$(PYTHONPATH),) $(PYTHON) -m $(TESTRUNNER) $(TEST_OPTIONS)
 COVERAGE = python3-coverage
+EDITABLE := $(shell pip show dulwich 2>/dev/null | grep -E "^Editable project location:")
```
jelmer, line 7: This doesn't work, it will suggest you're in an editable installation if there's a system-wide install of dulwich. 

pip maybe also not be available as "pip" but only as "pip3"

I'd keep this simple and avoid suggest running the Makefile to invoke the test suite in CONTRIBUTING.rst, but just running "python -m unittest dulwich.tests.test_suite".

jelmer (changes requested):

Makefile
jelmer: On Debian systems where there's packages that use Dulwich this is quite common. This is the case on my own system, and running "pip show dulwich" here picks up the system version of Dulwich. This is breaking my personal workflow with the Makefile, because now it stops building the extensions.

I'd be inclined to suggest that CONTRIBUTING.rst doesn't reference the Makefile at all and just suggests running plain commands. That matches what's in the CI workflow, and addresses your concerns about the Make targets. And then CI can avoid touching the Makefile.

jelmer:
Hi @mathrick - I've trimmed this down a bit to exclude the Makefile changes, since I think the other changes to CONTRIBUTING.rst are really useful and I don't want to block those. I hope that's okay. Thanks for the PR!

Still open to refactoring of the makefile, but hopefully this will at least help new contributors.

jelmer (approved):

mathrick:
> Hi @mathrick - I've trimmed this down a bit to exclude the Makefile changes, since I think the other changes to CONTRIBUTING.rst are really useful and I don't want to block those. I hope that's okay. Thanks for the PR!
> 
> Still open to refactoring of the makefile - and I appreciate this doesn't address all of your concerns (especially around CI being able to invoke the Makefile) but hopefully this will at least help new contributors.

Sure thing! Apologies for the late reply; I got pretty busy at work, so this will at least move things forward. I think splitting the make changes into their own thing is a very reasonable way of doing it.

Issue #1847, opened 2025-08-23
Enhance git branch with missing options


jelmer:
## Summary
The git branch command in dulwich is missing several useful options that are available in standard git.

## Missing Options
- `--all`: List both remote-tracking and local branches
- `--remotes`: List only remote-tracking branches
- `--merged`: List branches that have been merged into the current branch
- `--no-merged`: List branches that have not been merged into the current branch
- `--contains`: List branches that contain a specific commit
- `--sort`: Sort branches by specified criteria (name, date, etc.)
- `--column`: Display branch list in columns
- `--list`: List branches matching a pattern
- `--show-current`: Show the current branch name
- `--copy`: Copy a branch
- `--move`: Move/rename a branch
- `--edit-description`: Edit the description of a branch

## Benefits
These options would improve dulwich's git branch functionality by:
- Providing comprehensive branch listing capabilities with `--all`, `--remotes`, and filtering options
- Enabling branch management workflows with merge status checking via `--merged` and `--no-merged`
- Supporting advanced branch operations like copying and moving with `--copy` and `--move`
- Improving output formatting and usability with `--column` and `--sort`
- Adding branch metadata management with `--edit-description`

This would bring dulwich's branch command closer to feature parity with standard git and support more comprehensive branch management workflows.

xifOO:
hi, @jelmer , can I tryto do this?

PR #1873, opened 2025-08-29
add --merged to branch command 

xifOO:
This PR implements the --merged flag for the dulwich branch command, which lists branches that have been merged into the current branch. https://github.com/jelmer/dulwich/issues/1847

### DIFF
--- a/dulwich/cli.py
+++ b/dulwich/cli.py
@@ -2083,9 +2083,9 @@ class cmd_branch(Command):
 
         if args.merged:
             try:
-                branches = porcelain.merged_branches(".")
+                branches_iter = porcelain.merged_branches(".")
 
-                for branch in branches:
+                for branch in branches_iter:
                     sys.stdout.write(f"{branch.decode()}\n")
 
                 return 0


tests/test_cli.py
```diff
@@ -459,6 +459,34 @@ def test_branch_list_all(self):
...
+        test_file = os.path.join(self.repo_path, "test.txt")
+        with open(test_file, "w") as f:
+            f.write("test")
+        self._run_cli("add", "test.txt")
+        self._run_cli("commit", "--message=Initial")
+
+        master_sha = self.repo.refs[b"refs/heads/master"]
+
+        # Create a merged branch (points to same commit as master)
+        self.repo.refs[b"refs/heads/merged-branch"] = master_sha
```
jelmer, line 473: Please also test with a non-merged branch, so we know we don't simply include all branches.

dulwich/porcelain.py
```diff
@@ -3307,6 +3307,38 @@ def get_commit_date(branch_name: bytes) -> int:
...
+
+    Args:
+      repo: Path to the repository
+    Returns:
+      List of branch names (without refs/heads/ prefix) that are merged
+      into the current HEAD
+    """
+    with open_repo_closing(repo) as r:
+        current_sha = r.refs[b"HEAD"]
+        merged_branches = []
```
jelmer, line 3321: Perhaps yield branch names here rather than returning a list?

jelmer (changes requested):
LGTM, two minor comments.

dulwich/porcelain.py
xifOO: Let's stick with returning a list? The stdout thing makes tests ugly

tests/test_cli.py
xifOO: https://github.com/jelmer/dulwich/pull/1873/commits/3ddc0ee46aeede158019fc46f748310c71cd9187

dulwich/porcelain.py
xifOO: change to yield https://github.com/jelmer/dulwich/pull/1873/commits/da2d9614973fab8e8cba4f2c4f519ad06a6439c4

dulwich/porcelain.py
```diff
@@ -3307,6 +3307,35 @@ def get_commit_date(branch_name: bytes) -> int:
...
+      Branch names (without refs/heads/ prefix) that are merged
+      into the current HEAD
+    """
+    with open_repo_closing(repo) as r:
+        current_sha = r.refs[b"HEAD"]
+
+        for branch_ref in r.refs.keys(base=b"refs/heads/"):
+            full_ref = b"refs/heads/" + branch_ref
+
+            if full_ref not in r.refs:
```
jelmer, line 3325: this logic seems off - why would the ref be missing at this point?

dulwich/porcelain.py
```diff
@@ -3307,6 +3307,35 @@ def get_commit_date(branch_name: bytes) -> int:
...
+                continue
+
+            try:
+                branch_sha = r.refs[full_ref]
+
+                # Check if branch is an ancestor of HEAD (fully merged)
+                if can_fast_forward(r, branch_sha, current_sha):
+                    yield branch_ref
+
+            except (KeyError, ValueError):
```
jelmer, line 3335: same here - .keys() yielded this ref so it should exist. Why would KeyError or ValueError be raised?

tests/test_cli.py
```diff
@@ -459,6 +459,45 @@ def test_branch_list_all(self):
...
+        # Test --merged listing
+        result, stdout, stderr = self._run_cli("branch", "--merged")
+        self.assertEqual(result, 0)
+
+        branches = [line.strip() for line in stdout.splitlines()]
+        expected_branches = {"master", "merged-branch"}
+        self.assertEqual(set(branches), expected_branches)
+
+        # Verify that the non-merged branch is not included
+        self.assertNotIn("non-merged-branch", branches)
```
jelmer, line 499: This seems superfluous since you're already checking it's not in there on line 496

PR #1875, opened 2025-08-31
add --contains to branch command

xifOO:
Implements the --contains flag for the dulwich branch command, 
which lists branches that contain a specific commit.
https://github.com/jelmer/dulwich/issues/1847

### DIFF
--- a/dulwich/porcelain.py
+++ b/dulwich/porcelain.py
@@ -3358,6 +3358,34 @@ def no_merged_branches(repo: RepoPath) -> Iterator[bytes]:
             yield branch_name
 
 
+def contains_branches(repo: RepoPath, commit: str) -> Iterator[bytes]:
+    """List branches that contain the specified commit.
+
+    Args:
+        repo: Path to the repository
+        commit: Commit-ish string (SHA, branch name, tag, etc.)
+
+    Yields:
+        Branch names (without refs/heads/ prefix) that contain the commit
+
+    Raises:
+        ValueError: If the commit reference is malformed or not found
+    """
+    with open_repo_closing(repo) as r:
+        try:
+            commit_obj = parse_commit(r, commit)
+            commit_sha = commit_obj.id
+        except (KeyError, ValueError):
+            raise ValueError(f"malformed object name {commit}")
+
+        for branch_ref in r.refs.keys(base=LOCAL_BRANCH_PREFIX):
+            full_ref = LOCAL_BRANCH_PREFIX + branch_ref
+            branch_sha = r.refs[full_ref]
+
+            if can_fast_forward(r, commit_sha, branch_sha):
+                yield branch_ref
+
+
 def active_branch(repo: RepoPath) -> bytes:
     """Return the active branch in the repository, if any.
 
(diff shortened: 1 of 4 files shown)


dulwich/porcelain.py
```diff
@@ -3358,6 +3358,34 @@ def no_merged_branches(repo: RepoPath) -> Iterator[bytes]:
             yield branch_name
 
 
+def contains_branches(repo: RepoPath, commit: str) -> Iterator[bytes]:
```
jelmer, line 3361: This name is a bit odd. Perhaps "branches_containing"?

dulwich/porcelain.py
```diff
@@ -3358,6 +3358,34 @@ def no_merged_branches(repo: RepoPath) -> Iterator[bytes]:
...
+        Branch names (without refs/heads/ prefix) that contain the commit
+
+    Raises:
+        ValueError: If the commit reference is malformed or not found
+    """
+    with open_repo_closing(repo) as r:
+        try:
+            commit_obj = parse_commit(r, commit)
+            commit_sha = commit_obj.id
+        except (KeyError, ValueError):
```
jelmer, line 3378: No point in reraising ValueError with less data.

Also, KeyError is not a sign of corruption. Why is it being caught here?

dulwich/porcelain.py
```diff
@@ -3358,6 +3358,34 @@ def no_merged_branches(repo: RepoPath) -> Iterator[bytes]:
...
+            raise ValueError(f"malformed object name {commit}")
+
+        for branch_ref in r.refs.keys(base=LOCAL_BRANCH_PREFIX):
```
jelmer, line 3381: Just iterate over ref *and* contents at the same time, rather querying the ref again.

jelmer (changes requested):

dulwich/porcelain.py
xifOO: try:
            commit_obj = parse_commit(r, commit)
            commit_sha = commit_obj.id
        except KeyError:
            raise ValueError(f"commit '{commit}' not found")
        except ValueError:
            raise
Did I understand you correctly?

xifOO: But then I have an unhandled KeyError in the code. KeyError: b'invalid123'. Is this normal behavior?

xifOO: or do I need to handle KeyError at CLI level?

xifOO: thanks, sometimes I'm stupid

dulwich/porcelain.py
```diff
@@ -3358,6 +3358,32 @@ def no_merged_branches(repo: RepoPath) -> Iterator[bytes]:
...
+
+    Raises:
+        ValueError: If the commit reference is malformed
+        KeyError: If the commit reference does not exist
+    """
+    with open_repo_closing(repo) as r:
+        try:
+            commit_obj = parse_commit(r, commit)
+            commit_sha = commit_obj.id
+        except ValueError:
```
jelmer, line 3379: Please drop this - it's not doing anything. You're catching the error and then just reraising it.

PR #1876, opened 2025-09-01
add --column to branch command

xifOO:
Implements the --column flag for the dulwich branch command, 
which display branch list in columns.
https://github.com/jelmer/dulwich/issues/1847

### DIFF
--- a/dulwich/cli.py
+++ b/dulwich/cli.py
@@ -2076,6 +2076,9 @@ class cmd_branch(Command):
             const="HEAD",
             help="List branches that contain a specific commit",
         )
+        parser.add_argument(
+            "--column", action="store_true", help="Display branch list in columns"
+        )
         args = parser.parse_args(args)
 
         if args.all:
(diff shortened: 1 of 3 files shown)


dulwich/cli.py
```diff
@@ -2146,6 +2149,72 @@ def run(self, args) -> Optional[int]:
                 sys.stderr.write(f"{e}")
                 return 1
 
+        if args.column:
+            try:
+                branches = porcelain.branch_list(".")
+
+                try:
```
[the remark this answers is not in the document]
xifOO: yes you're right, for some reason I decided to make this a separate flag.

dulwich/cli.py
```diff
@@ -190,6 +191,74 @@ def launch_editor(template_content: bytes = b"") -> bytes:
         os.unlink(temp_file)
 
 
+def write_columns(items: Union[Iterator[bytes], list[bytes]]) -> None:
```
jelmer, line 194: Please also add some tests for the new write_columns function

dulwich/cli.py
```diff
@@ -190,6 +191,74 @@ def launch_editor(template_content: bytes = b"") -> bytes:
...
+    The function calculates the optimal number of columns to fit the terminal
+    width and displays the items in a formatted column layout with proper
+    padding and alignment.
+    """
+    try:
+        ter_width = os.get_terminal_size().columns
+    except OSError:
+        ter_width = 80
+
+    item_names = [item.decode() for item in items]
```
jelmer, line 209: this should handle invalid utf-8 - right now it will cause an exception

dulwich/cli.py
jelmer: In particular, the corner cases that this tries to deal with

PR #1880, opened 2025-09-04
Fix git-lfs smudge by quoting special characters path

### LINKED ISSUE #1878: dulwich fails to clone lfs files having names with special character when using smudge command
Hi, it seems that dulwich fails to clone repositories that contains lfs-tracked files having special characters in their names. When the lfs tracked file has a name like `my&file.png`, the https://github.com/jelmer/dulwich/blob/master/dulwich/filters.py#L296 doesn't quote the path properly. One possible solution could be the usage of quote from the shlex stdlib: https://docs.python.org/3/library/shlex.html#shlex.quote.

BTW, this is a problem for `dulwich 0.24.1` rather than the current unreleased master because the `0.24.1` has only the "smudge command" code path that is used as a fallback on the current master. The default filter process codepath on master seems to be working properly.

petrchmelar:
- fix git-lfs smudge with special-characters path for non filter process code path

Fixes #1878

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -81,6 +81,9 @@
    parallel stat operations when checking for unstaged changes. This improves
    performance on slow filesystems like NFS. (Jelmer Vernooĳ, #1851)
 
+ * Fix smudge filter subprocess fallback for special characters in path.
+   (Petr Chmelar, #1878)
+
 0.24.1	2025-08-01
 
  * Require ``typing_extensions`` on Python 3.10.
--- a/dulwich/filters.py
+++ b/dulwich/filters.py
@@ -22,6 +22,7 @@
 """Implementation of Git filter drivers (clean/smudge filters)."""
 
 import logging
+import shlex
 import subprocess
 import threading
 from typing import TYPE_CHECKING, Callable, Optional
@@ -331,6 +332,8 @@ class ProcessFilterDriver:
 
     def smudge(self, data: bytes, path: bytes = b"") -> bytes:
         """Apply smudge filter using external process."""
+        import os
+
         path_str = path.decode("utf-8", errors="replace")
 
         # Try process filter first (much faster)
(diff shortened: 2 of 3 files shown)


tests/test_filters.py
```diff
@@ -940,23 +940,32 @@ def test_special_characters_in_pathname(self):
...
         ]
 
         test_data = b"test data"
 
-        for path in special_paths:
-            result = driver.smudge(test_data, path)
-            self.assertEqual(result, b"test data")
+        for process_cmd, smudge_cmd in [
+            (f"{sys.executable} {self.test_filter_path}", None),
+            (None, "git-lfs smudge -- %f"),
```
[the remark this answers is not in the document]
petrchmelar: Sorry, I`m not that familiar with a codebase. What is recommended way to execute the `smudge_cmd` codepath in the `smudge` func so the test will check whether it can handle special characters or not?

NEWS
```diff
@@ -51,6 +51,9 @@
  * Add IPv6 support for git:// protocol URLs.
    (Jelmer Vernooĳ, #1796)
 
+ * Fix git-lfs smudge for special characters in path.
```
jelmer, line 54: This isn't git-lfs specific - perhaps just mention filters?

petrchmelar: Something like?

```
 * Fix smudge filter subprocess fallback for special characters in path.
   (Petr Chmelar, #1878)
```

jelmer: Thanks!

petrchmelar: Fixed

jelmer (approved):

dulwich/filters.py
```diff
@@ -328,7 +329,7 @@ def smudge(self, data: bytes, path: bytes = b"") -> bytes:
             return data
 
         # Substitute %f placeholder with file path
-        cmd = self.smudge_cmd.replace("%f", path_str)
+        cmd = self.smudge_cmd.replace("%f", shlex.quote(path_str))
```
petrchmelar, line 332: Unfortunatelly the shlex quoting does not produce a command compatible with `subprocess.run(..., shell=True)`. To do so I have two options:
1. use [barely documented](https://discuss.python.org/t/why-is-subprocess-list2cmdline-not-documented/25044) list2cmdline for windows codepath
2. change the subprocess call to be run without shell=True - I'm really not sure if it is a right thing to do in context of the commands beeing called

jelmer: Sorry for the slow reply.

I'm inclined to special-case Windows and use list2cmdline; using subprocess.run(shell=True) seems like the obvious correct thing to do on other platforms.

jelmer:
I've landed a fix based on yours - thanks!

PR #1882, opened 2025-09-05
add --list to branch command

xifOO:
Implements the --list flag for the dulwich branch command, 
which display list branches matching a pattern.


dulwich/cli.py
```diff
@@ -2177,9 +2185,9 @@ def print_branches(
 
         if args.all:
             try:
-                branches = porcelain.branch_list(".") + porcelain.branch_remotes_list(
```
jelmer, line 2180: Rather than reimplementing this three times (including the pattern matching), please:

* Obtain a branch_iter first (whether though args.all, args.merged, args.no_merged, etc)
* Then filter that for patterns, if specified
* Then print it

xifOO: How can I test the pattern filtering? Should I write it as a separate function and test it independently?

jelmer (approved):

jelmer:
> did something happen to the tests?

They're failing on Windows. You can't use fnmatch - it's case-insensitive on Windows. Perhaps fnmatchcase?

Issue #1369, opened 2024-09-27
replace gpg dependency to python-gnupg


DinhHuy2010:
change gpg dependency to [python-gnupg](https://github.com/vsajip/python-gnupg) so that gpg feature should be usable on windows

jelmer:
IIRC that module has security issues; if moving away from the upstream python bindings, I'd consider https://github.com/isislovecruft/python-gnupg. Though somebody else who

DinhHuy2010:
ok

PR #1888, opened 2025-09-16
WIP: use alternate GnuPG bindings

meejah:
Fixes https://github.com/jelmer/dulwich/issues/1369

This should work with either Isis Lovecruft's fork of python-gnupg, or upstream.

Work-in-progress: enough here to work for me locally, but needs proper packaging and tests. (e.g. depend on one of the above projects, remove remainder of `import gpg` and uses, tests).

What spurred me to do this is that I keep having problems installing `gpgme` bindings, and it seems Debian has removed the `gpgme-config` ..?

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -1050,20 +1050,19 @@ class Tag(ShaFile):
           keyid: Optional GPG key ID to use for signing. If not specified,
                  the default GPG key will be used.
         """
-        import gpg
+        import gnupg
+        gpg = gnupg.GPG()
 
-        with gpg.Context(armor=True) as c:
-            if keyid is not None:
-                key = c.get_key(keyid)
-                with gpg.Context(armor=True, signers=[key]) as ctx:
-                    self.signature, _unused_result = ctx.sign(
-                        self.as_raw_string(),
-                        mode=gpg.constants.sig.mode.DETACH,
-                    )
-            else:
-                self.signature, _unused_result = c.sign(
-                    self.as_raw_string(), mode=gpg.constants.sig.mode.DETACH
-                )
+        sig = gpg.sign(
+            self.as_raw_string(),
+            keyid=keyid,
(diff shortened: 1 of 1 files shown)


jelmer:
> Filed as an early draft for these reasons:
> 
>     * is this likely to be accepted?
>       
>       * if "yes", I'll find more motivation to finish it :)
> 
>     * which bindings to use?
> 
> 
> It's very odd that there's no interface in python-gnupg for "I have signature bytes, and data bytes" so maybe I missed something?

Yes, if we can find a way to get this to work well it's likely to be accepted. I don't want to overcomplicate this code though (i.e. I don't think it makes sense to support multiple backends). 

The other backend we could consider is pysequoia (https://github.com/wiktor-k/pysequoia), which seems to at least have the verify/sign functions.

jelmer:
> Yeah, I don't want to support multiple -- but at this point you could choose "upstream" or "Lovecruft's fork" without changes.
> 
> The API of PySequoia indeed looks way nicer -- I'll see if I can get it to find keys ("Certs") from e.g. `~/.gnupg/` etc automagically?

So there are a lot of packages out there:

* gpg: the upstream package, doesn't seem to upload to pypi and is tricky to build
* pygpgme: looks like a reasonable candidate? I think I've used it before, and it's actively maintained
* gnupg, pre-lovecruft: has a number of well-documented security issues
* lovecruft's fork - pretty_bad_protocol: hasn't been touched in 7 years
* pysequoia: based on sequoia rather than gnupg (big plus!) but might be harder/impossible to use with existing ~/.gnupg 
* probably others I've missed?

Not supporting reading keys from ~/.gnupg I think would be acceptable, so I'd lean towards either pygpgme or pysequoia. But this is just from some casual research - open to being convinced that something else is more appropriate.

meejah:
superseded by https://github.com/jelmer/dulwich/pull/1896

PR #1896, opened 2025-10-01
Sequoia instead of "gpg"

### LINKED ISSUE #1369: replace gpg dependency to python-gnupg

meejah:
Fixes #1369 

This replaces the use of "gpg" library with sequoia. This branch does this by shelling out to "sq".

I'd rather be using a Python library here, but `pysequoia` doesn't seem to be able to do most of what we get from using "sq" directly. I do have an attempt at that, too, but `pysequoia.verify()` wants a "store" argument -- but `pysequoia.Store` was removed in 0.1.24.

In addition "`sq`" seems to do fairly well at interacting with a `gpg-agent`, so the "shell out" implementation lets the API stay basically the same, and deal with key-id's ("fingerprints" to match sequoia).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -64,6 +64,10 @@
    parallel stat operations when checking for unstaged changes. This improves
    performance on slow filesystems like NFS. (Jelmer Vernooĳ, #1851)
 
+ * Use Sequoia for tag signing and verification instead of GnuPG. The
+   API remains the same but uses the "sq" tool to accomplish OpenPGP
+   operations (instead of importing "gpg" as before). (meejah, #1369)
+
 0.24.1	2025-08-01
 
  * Require ``typing_extensions`` on Python 3.10.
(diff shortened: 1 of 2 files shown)


meejah:
Ah, trying to write a ticket helped, rubber-duck style ;)

Turns out that although the "store" code was removed, the "store" argument to "verify" is now a callable that does "whatever the user wants" to fetch certificates.

So, I've updated this branch with my "pysequoia" attempts. This means the only "shell out to sq" is a default implementation of the "store" callback -- which could be moved to docs instead (i.e. force the user to call that if they need). Similar with the porcelain: now the user has to get `PySigner` instances.

The type-checking here is probably wrong, as `pysequoia` doesn't actually export `PySigner` (I filed a ticket for that).

If this looks preferable, the implementation of `Commit.sign` still needs to be adjusted as well.

jelmer:
Obsoleted by the new signature interface.

PR #2002, opened 2025-11-20
skip tests that require merge3 when it's not available

kulikjak:
Hi, there are still several tests that seemingly need `merge3` installed to pass and are not skipped gracefully when it's not available.

I needed to apply this change to make the test suite green again (resolved in a same way as in #1760).


kulikjak:
Uh yes - sorry for that. Will do.

PR #2039, opened 2025-12-19
Add support for parsing commits with broken author/committer lines

jelmer:
This adds parse_commit_broken() to handle various broken formats found in real-world Git repositories:
- Missing angle brackets around email addresses
- Unsigned timezones (e.g., "0000" instead of "+0000")
- Double-negative timezones (e.g., "--700")
- Negative timestamps
- Long/short/nonsensical timezone values

The function returns a Commit object but includes a warning in the docstring that commits parsed this way may not round-trip correctly through serialization.

Based on PR #927 by Valentin Lorentz with real examples from broken commits.

### DIFF
--- a/dulwich/objects.py
+++ b/dulwich/objects.py
@@ -61,8 +61,7 @@ __all__ = [
     "object_class",
     "object_header",
     "parse_commit",
-    "parse_time_entry",
-    "parse_timezone",
+    "parse_commit_broken",
     "parse_tree",
     "pretty_format_tree_entry",
     "serializable_property",
@@ -75,6 +74,7 @@ __all__ = [
 import binascii
 import os
 import posixpath
+import re
 import stat
 import sys
 import zlib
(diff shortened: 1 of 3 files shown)


dulwich/objects.py
```diff
@@ -1730,6 +1736,41 @@ def parse_timezone(text: bytes) -> tuple[int, bool]:
...
+        text = b"+" + text
+
+    # cgit parses the first character as the sign, and the rest
+    #  as an integer (using strtol), which could also be negative.
+    #  We do the same for compatibility. See #697828.
+    sign = text[:1]
+    offset = int(text[1:])
+    if sign == b"-":
+        offset = -offset
+    unnecessary_negative_timezone = offset >= 0 and sign == b"-"
```
progval:
```suggestion
    if sign == b"-":
        unnecessary_negative_timezone = offset < 0
        offset = -offset
```

easier to read IMO

dulwich/objects.py
```diff
@@ -2258,6 +2388,99 @@ def _get_extra(self) -> list[tuple[bytes, bytes | None]]:
     _TYPE_MAP[cls.type_num] = cls
 
 
+# Public API functions
+
+
+@replace_me(since="0.21.0", remove_in="0.24.0")
```
progval: I'm not sure I understand how this work. What should `parse_commit` be replaced with? The private `_parse_commit` function?

PR #2045, opened 2025-12-29
simplify FileLocked

dimbleby:
Making callers handle `str | bytes` is already an annoyance; having them handle `str | bytes | os.PathLike[str] | os.PathLike[bytes]` is really making life harder than it need be

### DIFF
--- a/dulwich/file.py
+++ b/dulwich/file.py
@@ -145,7 +145,7 @@ class FileLocked(Exception):
 
     def __init__(
         self,
-        filename: str | bytes | os.PathLike[str] | os.PathLike[bytes],
+        filename: str | bytes,
         lockfilename: str | bytes,
     ) -> None:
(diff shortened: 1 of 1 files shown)


jelmer:
Why is this a problem for callers?

dimbleby:
eg suppose I want to print the file as an error to show to users.

if I just try to print `f"{e.filename}"` then mypy warns
```
If x = b'abc' then f"{x}" or "{}".format(x) produces "b'abc'", not "abc". If this is desired behavior, use f"{x!r}" or "{!r}".format(x). Otherwise, decode the bytes  [str-bytes-safe]
```
fair enough, indeed I would have preferred to have a string all along

but if I start down the tedious road of writing myself a `def to_str(filename: str | bytes) -> str` - well currently I have four cases to handle instead of two.

(reiterating: one would have been best!)

PR #2123, opened 2026-03-25
[bug fix] contrib: fix paramiko host key verification

quart27219:
## Summary

Fix host key verification in `contrib/paramiko_vendor.py` by loading known hosts and rejecting unknown SSH host keys by default.


jelmer (approved):

jelmer:
For those coming across this linked in a security vulnerability report: this code lives in contrib/, which is explicitly not a part of Dulwich itself and intentionally completely unsupported:

https://github.com/jelmer/dulwich/blob/main/contrib/README.md

PR #2142, opened 2026-04-17
Fix gitignore parent re-include handling

### LINKED ISSUE #2141: gitignore negation fails after re-including an excluded parent directory
Dulwich 1.1.0 reports a path as ignored when official Git re-includes it after
the parent directory has first been excluded and then explicitly re-included by
a later negation pattern.

Official rule
-------------

The Git gitignore manual says that, within one precedence level, "the last
matching pattern decides the outcome". It also says that a leading `!` negates
a previous exclusion, so a matching path becomes included again, except when a
parent directory of that file is still excluded.

Reference:
https://git-scm.com/docs/gitignore

In this case the parent directory is not still excluded: `dist/` is excluded,
then re-included by `!dist/`, and only then are children excluded/re-included.

Minimal PoC repository
----------------------

Repository layout:

```text
poc-repo/
  .gitignore
  dist/
    drop.txt
    keep.txt
```

`.gitignore`:

```gitignore
dist/
!dist/
dist/*
!dist/keep.txt
```

Construction commands:

```bash
rm -rf /tmp/dulwich-gitignore-reinclude-poc
mkdir -p /tmp/dulwich-gitignore-reinclude-poc/dist
cd /tmp/dulwich-gitignore-reinclude-poc
git init -q
cat > .gitignore <<'EOF'
dist/
!dist/
dist/*
!dist/keep.txt
EOF
printf 'ignored\n' > dist/drop.txt
printf 'visible\n' > dist/keep.txt
```

Official Git reproduction command:

```bash
git --version
for path in dist/drop.txt dist/keep.txt; do
  echo "== $path =="
  git check-ignore -v -n "$path"
  git check-ignore --quiet "$path"
  echo "ignored_exit=$? # 0=ignored 1=visible"
done
```

Official Git output, using Git v2.53.0:

```text
official_git=git version 2.53.0
== dist/drop.txt ==
.gitignore:3:dist/*	dist/drop.txt
verbose_exit=0
ignored_exit=0 # 0=ignored 1=visible
== dist/keep.txt ==
.gitignore:4:!dist/keep.txt	dist/keep.txt
verbose_exit=0
ignored_exit=1 # 0=ignored 1=visible
```

Dulwich reproduction command:

```bash
python - <<'PY'
import os
from dulwich.ignore import IgnoreFilterManager
from dulwich.repo import Repo
import dulwich

repo = Repo(".")
manager = IgnoreFilterManager.from_repo(repo)
print(f"dulwich={dulwich.__version__}")
for path in ("dist/drop.txt", "dist/keep.txt"):
    print(f"== {path} ==")
    print(f"is_ignored={manager.is_ignored(path)}")
PY
```

Dulwich 1.1.0 output:

```text
dulwich=(1, 1, 0)
== dist/drop.txt ==
is_ignored=True
== dist/keep.txt ==
is_ignored=True
```

Expected behavior:

```text
dist/drop.txt  ignored
dist/keep.txt  visible / not ignored
```

Actual Dulwich behavior:

```text
dist/drop.txt  ignored
dist/keep.txt  ignored
```

Root cause
----------

The issue appears to be in `dulwich.ignore._check_parent_exclusion`. The current
logic treats an earlier directory exclusion, such as `dist/`, as permanently
blocking a later file-level negation, even when the parent directory was
re-included by `!dist/` before `!dist/keep.txt` is evaluated.

During `IgnoreFilterManager` path traversal, the same directory-only pattern can
also appear more than once in the match list. If parent-exclusion handling uses
that expanded match list directly, duplicate directory matches can distort the
original pattern ordering.

Proposed fix
------------

I have a patch that changes the parent-exclusion check to compute the effective
state of each parent directory in pattern order before applying the final
negation:

- find the final negation that would include the file;
- for each parent directory, replay earlier matching patterns in original order;
- allow a later parent-directory negation such as `!dist/` to clear an earlier
  parent exclusion;
- ignore duplicate matches of the same `Pattern` object introduced during path
  traversal.

With this patch, Dulwich matches official Git for the PoC:

```text
dulwich=(1, 1, 0)
== dist/drop.txt ==
is_ignored=True
== dist/keep.txt ==
is_ignored=False
```

Regression test added
---------------------

```python
def test_reincluded_parent_allows_file_reinclude(self) -> None:
    ...
    self.assertFalse(m.is_ignored("dist/"))
    self.assertTrue(m.is_ignored("dist/drop.txt"))
    self.assertFalse(m.is_ignored("dist/keep.txt"))
```

Validation
----------

Patch branch:
https://github.com/N0zoM1z0/dulwich/tree/fix-gitignore-reinclude-after-reopened-dir

Commit:
`a993fc07 Fix gitignore parent re-include handling`

Validation commands run locally:

```bash
python3 -m unittest tests.test_ignore.IgnoreFilterManagerTests.test_reincluded_parent_allows_file_reinclude \
  tests.test_ignore.IgnoreFilterManagerTests.test_parent_directory_exclusion \
  tests.test_ignore.IgnoreFilterManagerTests.test_parent_exclusion_with_wildcards \
  tests.test_ignore.IgnoreFilterManagerTests.test_parent_exclusion_with_double_asterisk \
  tests.test_ignore.IgnoreFilterManagerTests.test_no_parent_exclusion_with_glob_contents \
  tests.test_ignore.IgnoreFilterManagerTests.test_issue_1721_directory_negation_with_double_asterisk

python3 -m unittest tests.test_ignore

python3 -m ruff check dulwich/ignore.py tests/test_ignore.py
python3 -m ruff format --check dulwich/ignore.py tests/test_ignore.py
```

Validation output summary:

```text
......
Ran 6 tests in 0.020s

OK

.............................
Ran 29 tests in 0.042s

OK

All checks passed!
2 files already formatted
```

N0zoM1z0:
Fixes #2141.

## Summary
- update parent-exclusion handling so a later `!dir/` re-include can allow a subsequent file-level negation to take effect
- ignore duplicate matches of the same pattern object that can be produced during path traversal
- add a regression test for `dist/`, `!dist/`, `dist/*`, `!dist/keep.txt`

## Tests
- `python3 -m unittest tests.test_ignore`
- `python3 -m ruff check dulwich/ignore.py tests/test_ignore.py`
- `python3 -m ruff format --check dulwich/ignore.py tests/test_ignore.py`

### DIFF
--- a/dulwich/ignore.py
+++ b/dulwich/ignore.py
@@ -68,21 +68,46 @@ def _check_parent_exclusion(path: str, matching_patterns: Sequence["Pattern"]) -
     Returns:
         True if parent exclusion applies (negation should be ineffective), False otherwise
     """
-    # Find the last negation pattern
-    final_negation = next(
-        (p for p in reversed(matching_patterns) if not p.is_exclude), None
-    )
-    if not final_negation:
+    final_negation_index = None
+    for i in range(len(matching_patterns) - 1, -1, -1):
+        if not matching_patterns[i].is_exclude:
+            final_negation_index = i
+            break
+    if final_negation_index is None:
         return False
 
+    final_negation = matching_patterns[final_negation_index]
     final_pattern_str = _pattern_to_str(final_negation)
+    parent_status: dict[str, bool | None] = {
+        parent: None for parent in _parent_directories(path)
+    }
+    seen_pattern_ids: set[int] = set()
 
-    # Check if any exclusion pattern excludes a parent directory
-    return any(
-        pattern.is_exclude
-        and _pattern_excludes_parent(_pattern_to_str(pattern), path, final_pattern_str)
-        for pattern in matching_patterns
-    )
+    for pattern in matching_patterns[:final_negation_index]:
+        pattern_id = id(pattern)
+        if pattern_id in seen_pattern_ids:
+            continue
+        seen_pattern_ids.add(pattern_id)
+        pattern_str = _pattern_to_str(pattern)
+        for parent in parent_status:
+            if not pattern.match(os.fsencode(parent)):
+                continue
+            if pattern.is_exclude:
+                if _pattern_excludes_parent(pattern_str, path, final_pattern_str):
+                    parent_status[parent] = True
+            else:
+                parent_status[parent] = False
+
+    return any(status is True for status in parent_status.values())
+
+
+def _parent_directories(path: str) -> list[str]:
+    """Return parent directories for a path, with trailing slashes."""
+    path = path.rstrip("/")
+    if not path or "/" not in path:
+        return []
+    parts = path.split("/")
+    return ["/".join(parts[:i]) + "/" for i in range(1, len(parts))]
 
 
 def _pattern_excludes_parent(
--- a/tests/test_ignore.py
+++ b/tests/test_ignore.py
@@ -404,6 +404,27 @@ class IgnoreFilterManagerTests(TestCase):
         self.assertFalse(m.is_ignored("a/b.txt"))
         self.assertTrue(m.is_ignored("a/c.dat"))
 
+    def test_reincluded_parent_allows_file_reinclude(self) -> None:
+        tmp_dir = tempfile.mkdtemp()
+        self.addCleanup(shutil.rmtree, tmp_dir)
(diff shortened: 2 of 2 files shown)


jelmer:
Thanks!

jelmer (approved):

PR #2144, opened 2026-04-19
Add initial Hypothesis config property tests

ctoth:
This adds a small first slice of property-based testing using Hypothesis.

I kept this deliberately small so the dependency, CI behavior, and test style can be reviewed independently from bug fixes or broader property-test expansion.

The goal is to establish a separate property-test shape without changing Dulwich's regular test suite:

- keeps `tests.test_suite` unchanged
- keeps the regular `unittest` runner
- keeps OSS-Fuzz/Atheris fuzzing separate and unchanged
- makes Hypothesis an optional test dependency
- puts property tests under `property_tests`, outside the regular test suite
- runs the new tests in CI as a separate step with a deterministic Hypothesis profile
- starts with `ConfigFile` because it is pure, fast, and parser/serializer-oriented

The new config property tests cover two things:

- arbitrary bytes either parse or fail with an expected config parse error
- generated canonical config files round-trip through `write_to_file()` and `from_file()`

While testing this locally, this already exposed at least one small config round-trip issue. I have kept the fix out of this PR so this first review can stay focused on the Hypothesis setup and initial test shape. I'll send the bug fix separately.

Also included:

- ignore `.hypothesis/`
- update the contributor docs to use the current `tests.test_suite` entry point
- document the separate property-test command

### DIFF
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -77,6 +77,12 @@ jobs:
         run: |
           pip install --upgrade coverage
           python -m coverage run -p -m unittest tests.test_suite
+      - name: Property-based tests
+        run: |
+          pip install --upgrade ".[test]"
+          python -m unittest property_tests.test_config
+        env:
+          HYPOTHESIS_PROFILE: ci
       - name: Run contrib tests
         run: |
           python -m unittest contrib.test_suite
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -79,7 +79,7 @@ jobs:
           python -m coverage run -p -m unittest tests.test_suite
       - name: Property-based tests
         run: |
-          pip install --upgrade ".[test]"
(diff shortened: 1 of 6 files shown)


jelmer:
Thanks for working on this! This is really useful.

I don't think this should be a part of the regular test suite, since it won't behave predictably. We don't want contributors to get failures others won't see.

I'd prefer if this was under a separate directory. We could optionally run it as part of the tests, but only if we can prohibit any randomness.

ctoth:
Thanks, that makes sense. I updated this to keep the property tests out of `tests.test_suite` and move them into a separate top-level `property_tests` package.

The regular coverage suite no longer installs the `test` extra or runs Hypothesis. CI now has a separate `Property-based tests` step, and the property-test module loads a deterministic Hypothesis profile by default (`derandomize=True`). The `ci` and `local-deep` profiles are deterministic too.

That should keep these useful for parser/serializer coverage without making the regular contributor test run locally variable.

PR #2145, opened 2026-04-20
Preserve quoted trailing whitespace in config values

ctoth:
Follow-up to #2144.

The initial Hypothesis config round-trip property exposed a case where quoted trailing whitespace was not preserved when reading config values back.

For example, a value like `b" "` is written as a quoted config value, but parsing discarded the whitespace because `_parse_string` buffered whitespace even while inside quotes.

This changes `_parse_string` so spaces and tabs inside quotes are preserved immediately, while trailing whitespace outside quotes is still ignored.

Also included:

- a focused regression test for round-tripping quoted trailing space and tab values
- removal of the temporary property-test filter that excluded values ending in space or tab

### DIFF
--- a/dulwich/config.py
+++ b/dulwich/config.py
@@ -793,7 +793,10 @@ def _parse_string(value: bytes) -> bytes:
             # the rest of the line is a comment
             break
         elif c in _WHITESPACE_CHARS:
-            whitespace.append(c)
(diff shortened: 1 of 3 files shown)


jelmer (approved):

jelmer:
Nice, thanks!

PR #2146, opened 2026-04-20
Fix timezone roundtrip with negative marker

ctoth:
Follow-up to the property-test work.

The timezone parse/format round-trip property exposed a case where `format_timezone()` does not round-trip positive sub-hour offsets when `unnecessary_negative_timezone=True`.

`format_timezone(60, True)` should produce a value that parses back to `(60, True)`, matching the existing double-negative behavior covered for whole-hour offsets such as `--700`.

This updates the timezone formatting arithmetic to build the HHMM value from the absolute offset before applying the negative marker.

Also included:

- a focused regression test for the minimized case
- an objects property test for timezone parse/format round-trips
- CI/docs update to discover all modules under `property_tests`

### DIFF
--- a/.github/workflows/pythontest.yml
+++ b/.github/workflows/pythontest.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Property-based tests
         run: |
           pip install --upgrade ".[hypothesis]"
-          python -m unittest property_tests.test_config
+          python -m unittest discover property_tests
         env:
           HYPOTHESIS_PROFILE: ci
       - name: Run contrib tests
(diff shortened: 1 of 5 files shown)


jelmer:
I'm not sure I follow what this is trying to do, unnecessary_negative_timezone is only really relevant for timezone 0000.

ctoth:
Oops, sorry - I think I overgeneralized the existing double-negative timezone compatibility behavior into a broader round-trip property than Dulwich actually wants.

If `unnecessary_negative_timezone` is only intended to be meaningful for `-0000`, then this PR is not the right direction. Feel free to close this one.

I do have a separate property-test change around pack deltas that found and fixes an actual Rust extension panic, so I'll keep that separate rather than trying to make this PR carry the property-testing work.

jelmer:
So in this specific case I think the correct fix is just to raise an exception in more cases - and have the property tests assert that.

PR #2149, opened 2026-04-21
Honor GIT_PROTOCOL env var when picking default protocol version (#1862)

mvanhorn:
## Summary

Adds `GIT_PROTOCOL` environment variable support to the Dulwich transport layer.

Fixes #1862

## Why

Git sets `GIT_PROTOCOL=version=N` to request a specific wire-protocol version for network operations (debugging, testing, servers that only support v2, etc.). Dulwich was ignoring it and always falling back to `DEFAULT_GIT_PROTOCOL_VERSION_SEND` when the caller didn't pass an explicit `protocol_version`.

## Change

- `dulwich/client.py`: new internal `_protocol_version_from_env()` helper that parses the colon-separated `key=value` format (e.g. `version=2` or `feature=extra:version=2`) and returns the requested version as an `int`, or `None` if unset / unparseable / missing a `version` key.
- `_handle_upload_pack_head`: at the single point that substitutes `DEFAULT_GIT_PROTOCOL_VERSION_SEND` for a `None` `protocol_version`, consult the env var first. Explicit callers (CLI `--protocol`, `porcelain.clone(protocol_version=...)`, etc.) still win because this only fires when the value arrives as `None`.
- `tests/test_client.py`: new `TestProtocolVersionFromEnv` suite covering unset / empty / `version=1` / `version=2` / multi-pair / non-version-key / unparseable inputs (7 cases).
- `NEWS`: 1.2.0-UNRELEASED entry with issue link.

## Testing

```
python3 -m pytest tests/test_client.py
...
223 passed, 1 skipped
```

Including the 7 new `TestProtocolVersionFromEnv` cases.

## Scope

Kept the change narrow: a single parse-and-wire site so the explicit-arg-wins precedence is preserved and no public signatures change. Other `GIT_*` env vars mentioned by #1860 / #1861 stay out of scope for this PR and can follow the same pattern if helpful.

PR #2156, opened 2026-04-22
Move GIT_SSH/GIT_SSH_COMMAND env lookup from client.py to cli.py

mvanhorn:
Follow-up to #2149 (review comment from @jelmer: "PRs to move them to dulwich.cli would be great").

Lifts the GIT_SSH_COMMAND / GIT_SSH env lookup out of `SSHGitClient.__init__` and into a new `dulwich.cli._ssh_command_from_env()` helper, matching the pattern already used for `GIT_PROTOCOL` in #2149. The CLI entry points (`cmd_archive`, `cmd_fetch_pack`, `cmd_clone`, `cmd_pull`) now pass the resolved value down; `SSHGitClient` itself only consults the explicit `ssh_command` parameter and `core.sshCommand` config.

Behavior:
- GIT_SSH_COMMAND wins over GIT_SSH (matches git).
- Empty `GIT_SSH_COMMAND` falls back to `GIT_SSH`.
- Library callers (anything using `SSHGitClient`/`porcelain` directly without passing `ssh_command`) no longer pick up these env vars. This is the intended behavior change.

Tests:
- Updated `tests/test_client.py` `test_ssh_command_precedence` / `test_ssh_command_config` so they assert the library-layer precedence only (config + explicit param).
- Added `tests/cli/test_cli.py::SshCommandFromEnvTest` covering GIT_SSH_COMMAND wins, GIT_SSH fallback, empty GIT_SSH_COMMAND fallback, both-unset.

Verified: `pytest tests/test_client.py tests/cli/test_cli.py` -> 503 passed, 2 skipped.

Depends on the same policy established in #2149 but touches disjoint code, so ordering against #2149 shouldn't matter.

### DIFF
--- a/dulwich/cli.py
+++ b/dulwich/cli.py
@@ -99,6 +99,25 @@ from .stripspace import stripspace
 logger = logging.getLogger(__name__)
 
 
+def _ssh_command_from_env() -> str | None:
+    """Return the ssh command requested via ``GIT_SSH_COMMAND`` / ``GIT_SSH``.
+
+    ``GIT_SSH_COMMAND`` wins over ``GIT_SSH``, matching git's own precedence.
+    Returns ``None`` when neither is set, so callers can fall through to the
+    ``core.sshCommand`` config or the transport default.
+
+    Env lookup lives here in the CLI layer rather than in the transport
+    library so that :mod:`dulwich.client` stays process-environment-free.
+    """
+    env_ssh_command = os.environ.get("GIT_SSH_COMMAND")
+    if env_ssh_command:
+        return env_ssh_command
+    env_ssh = os.environ.get("GIT_SSH")
+    if env_ssh:
+        return env_ssh
+    return None
+
+
 def to_display_str(value: bytes | str) -> str:
     """Convert a bytes or string value to a display string.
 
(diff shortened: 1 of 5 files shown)


jelmer (approved):

jelmer:
@mvanhorn please take a look at the test failures

mvanhorn:
Fixed in 4b70bb42 - ran `ruff format` on `dulwich/client.py`. `ruff check` and `ruff format --check` both clean locally.

mvanhorn:
Diagnosed: the cli call was passing `ssh_command=_ssh_command_from_env()` (which returns `str | None`) into `porcelain.clone()`'s `**kwargs`, which was typed `str | bytes | Sequence[str | bytes]` and didn't accept None.

Fixed in 53760a07 by adding `ssh_command: str | None = None` as an explicit parameter on `clone()`, matching the pattern already used by `fetch()` and `push()`. It's forwarded into `transport_kwargs` when non-None so `get_transport_and_path()` receives it the same way it did before.

Verified: `mypy dulwich/porcelain/__init__.py dulwich/cli.py` is clean, and the 12 clone tests in `tests/porcelain/` still pass. The 8 other mypy errors (dulwich/lfs.py and dulwich/client.py) pre-exist on main.

mvanhorn:
Thanks for the merge, @jelmer. Right place for the SSH env lookup.

PR #2181, opened 2026-05-19
tests: skip non-UTF8 filename tests on UTF-8-only filesystems (#2174)

### LINKED ISSUE #2174: tests broken on OpenZFS with utf8only=on
My root filesystem is ZFS. It has the option utf8only=on set. That prevents writing filenames with non-UTF8. The result is that a few tests fail: BuildIndexTests.test_no_decode_encode, DiskRefsContainerTests.test_cyrillic, and BuildRepoRootTests.test_commit_no_encode_decode.

I can (and will) just skip the tests, but it would be nice if they just worked regardless of filesystem. I see that there are already some platform checks - maybe they could be extended.

Attached is a log of the test failures. If it matters, I'm building with nix.

[test-failures.txt](https://github.com/user-attachments/files/27815945/test-failures.txt)

mvanhorn:
Extend the per-test `OSError` handlers in the three non-UTF8 filename tests (`test_no_decode_encode`, `test_cyrillic`, `test_commit_no_encode_decode`) to also skip on `errno.EILSEQ` so they pass on filesystems that reject non-UTF-8 names (OpenZFS with `utf8only=on`, ZFS on FreeBSD, etc.). Matches the shape of the pre-existing `errno==92 darwin` arm in `test_no_decode_encode`. Closes #2174.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -26,6 +26,7 @@ __all__ = [
     "SkipTest",
     "TestCase",
     "expectedFailure",
+    "filesystem_supports_non_utf8_filenames",
     "skipIf",
 ]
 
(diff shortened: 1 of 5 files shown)


tests/__init__.py
```diff
@@ -50,6 +51,41 @@ def __init__(self, dependency: str) -> None:
...
+
+    Args:
+      directory: Directory to probe. Defaults to the system temp directory.
+
+    Returns:
+      ``True`` if the filesystem accepted a probe filename containing a
+      non-UTF8 byte; ``False`` otherwise.
+    """
+    if directory is None:
+        directory = tempfile.gettempdir()
```
jelmer: Using tempdir here seems nonsensical, it might well be on a different filesystem than where the tests are running (with different supported features).

tests/__init__.py
```diff
@@ -50,6 +51,41 @@ def __init__(self, dependency: str) -> None:
...
+    probe_dir = tempfile.mkdtemp(dir=directory)
+    try:
+        # 0xff is not a valid UTF-8 lead byte, so any UTF-8-only filesystem
+        # will reject this name with OSError (Linux ZFS utf8only=on returns
+        # EILSEQ on Linux, EINVAL on FreeBSD).
+        probe = os.path.join(os.fsencode(probe_dir), b"\xff_dulwich_probe")
```
jelmer: per the comments in the bug, this should specifically be check for the ZFS feature we care abou, not probe - the risk of probing and just catching any OSError/UnicodeError is that we'll end up silently skipping this even on regular Linux.

tests/test_index.py
```diff
@@ -753,6 +753,10 @@ def test_no_decode_encode(self) -> None:
         repo_dir = tempfile.mkdtemp()
         repo_dir_bytes = os.fsencode(repo_dir)
         self.addCleanup(shutil.rmtree, repo_dir)
+        if not filesystem_supports_non_utf8_filenames(repo_dir):
```
jelmer: Why is this guarded? It currently runs on other platforms AFAICT

jelmer:
Please don't use overly long AI generated PR descriptions; it could be a single paragraph long, while saying exactly the same thing.

jelmer:
It's not enough to just skip the tests, this should also verify that we do the right thing on these filesystems when there is a non-utf8 character. Does the code correctly avoid writing these files when there is a non-utf8 character? How does C git handle it?

And we should test with this scenario in CI ideally, otherwise this will likely bitrot.

mvanhorn:
Good points -- sorry for the wall-of-text PR description, and the helper was overreach. Pushed a rework that:

- drops the `filesystem_supports_non_utf8_filenames` helper entirely;
- in `test_no_decode_encode`, extends the existing `except OSError` block to also skip on `errno.EILSEQ` (same shape as the `errno == 92 darwin` arm right next to it);
- in `test_cyrillic` and `test_commit_no_encode_decode`, wraps the file write in a narrow `try/except OSError` checking the same errno;
- updated the NEWS entry under 1.2.3 UNRELEASED.

On your deeper question -- when the OS rejects the filename, the existing code path bubbles the OSError up unchanged: `build_index_from_tree` (and the worktree stage path in test_commit_no_encode_decode) doesn't catch it, so dulwich behaves like C git, which also fails with the OS error and stops without writing a partial index. I left that as the current behaviour rather than adding new error-translation code, but happy to add an explicit assertion test for that propagation if you'd prefer it spelled out.

Rebased on current main.

jelmer:
Thanks!

Have you tested this on ZFS with utf8 enforcement, or is there some way we could test this in CI?

PR #2191, opened 2026-05-20
Use decorator form for properties

stephenfin:
mypy cannot handle the function form, meaning these properties appear as `Any`.

By fixing this, we highlight a few issues in our own types.


stephenfin:
Missed the fact tests were failing. Will need to take another go at this

PR #2196, opened 2026-05-28
reject backslash and colon in validate_path_element_ntfs

netliomax25-code:
Tree entries like `..\evil` or `evil:.git\config` slip through `validate_path_element_ntfs` today: `split(b"/")` keeps them as one element, and the normalizer only rstrip-lowercases. On Windows that lets a tree path escape the working tree at `build_index_from_tree` / checkout time, the same shape recently fixed in archive.py. Mirroring git's verify_path under core.protectNTFS by rejecting elements that contain `\` or `:`.

### DIFF
--- a/dulwich/index.py
+++ b/dulwich/index.py
@@ -1979,6 +1979,14 @@ def validate_path_element_ntfs(element: bytes) -> bool:
     Returns:
       True if path element is valid for NTFS, False otherwise
     """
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the PR; unfortunately we're in the process of preparing a security release that already includes a fix for this.

PR #2198, opened 2026-05-28
shell-quote path placeholders in merge driver command

netliomax25-code:
%P in a custom merge driver command is the file's tree path, which an attacker controls. ProcessMergeDriver.merge substitutes it (and %O/%A/%B) raw into a shell=True call, so a tree entry named `a$(touch /tmp/pwned).txt` runs that command during a three-way merge once a merge=<driver> attribute matches. git shell-quotes these placeholders with sq_quote; use shlex.quote to match.

### DIFF
--- a/dulwich/merge_drivers.py
+++ b/dulwich/merge_drivers.py
@@ -29,6 +29,7 @@ __all__ = [
 ]
 
 import os
+import shlex
 import subprocess
 import tempfile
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the PR; unfortunately we're in the process of preparing a security release that already includes a fix for this.

PR #2201, opened 2026-05-29
drop credentials when info/refs redirects to a different origin

netliomax25-code:
_discover_references at client.py:4506 rebases base_url to the server-supplied redirect Location, and every following request re-sends pool_manager.headers, so a redirect to another host hands the configured basic-auth / http.extraHeader credentials to that host. Drop the Authorization header when the redirect changes scheme, host, or port.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -4463,6 +4463,9 @@ class AbstractHttpGitClient(GitClient):
         """
         raise NotImplementedError(self._http_request)
 
+    def _reset_credentials(self) -> None:
+        """Drop stored credentials before following a cross-origin redirect."""
+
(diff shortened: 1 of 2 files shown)


jelmer:
Thanks for the PR. For defense in depth it would be good to fix this, but I think the actual scope of this is very limited:

* "attacker" is a misnomer here since that's the server you originally sent your credentials to. Maybe there's a more complicated scenario in which this can be exploited, but I don't see it in the context of a regular git server and dulwich.
* It doesn't happen for the first request as stated in the PR description, just subsequent ones

The fix violates layering, alternative submission in https://github.com/jelmer/dulwich/pull/2202

PR #2206, opened 2026-06-01
don't expand config includes when parsing .gitmodules

netliomax25-code:
read_submodules parses the untrusted .gitmodules of a cloned tree with config includes enabled, so a hostile `[include] path` makes dulwich open and parse arbitrary files during `clone --recurse-submodules`; disable include expansion there, like git does.


jelmer:
closing in favour of https://github.com/jelmer/dulwich/pull/2208

PR #2212, opened 2026-06-02
contain loose ref names within the ref store in refpath

netliomax25-code:
Loose ref names reach DiskRefsContainer.refpath from lookups that skip the validation the write paths get (_check_refname): read_loose_ref, follow's resolved symref target, and git-upload-archive's client-supplied `argument` in server.py. os.path.join honors `..` and absolute names, so resolving a name like `../../secret` reads a file outside the ref store and hands back its first line as the ref value. Enforce containment in refpath so every read and write site is covered at the one chokepoint.


jelmer:
Addressed in #2234

PR #2222, opened 2026-06-08
contain get_named_file reads within the control dir

netliomax25-code:
1. get_named_file joins the supplied name onto the control dir and only strips a leading separator, so a name like ../../config resolves outside the control dir and opens an unrelated file.
2. dulwich.web and dulwich.aiohttp feed request-derived names straight into this helper (HEAD, info/refs, pack files), the dumb-serving case the method's own TODO flagged.
3. Resolve the joined path and return None (the existing missing-file result) when it falls outside basedir, mirroring the containment check in FileSystemBackend.open_repository; legitimate names still resolve.

Regression test reads an out-of-tree file before the change and gets None after.

### DIFF
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -1804,15 +1804,23 @@ class Repo(BaseRepo):
             control dir.
         Returns: An open file object, or None if the file does not exist.
         """
-        # TODO(dborowitz): sanitize filenames, since this is used directly by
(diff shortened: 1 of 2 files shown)


jelmer:
This is an internal function, not meant to provide a security layer. Callers are responsible for sanitizing. This function gets called a lot so adding many more calls here is not great for performance.

PR #2241, opened 2026-06-17
validate object id in LFS server request handlers

netliomax25-code:
1. The `/objects/<oid>` handlers (`handle_download`, `handle_upload`, and `handle_verify` via `_object_exists`) read the object id straight from the request path and pass it to `LFSStore.open_object`.
2. `open_object` maps the id onto a file with `_sha_path` (`objects/<oid[:2]>/<oid[2:4]>/<oid>`), so an unvalidated id escapes the store: `GET /objects/....` resolves above the objects directory, and on Windows a backslash id like `..\..\..\..\Windows\win.ini` traverses to an arbitrary file and is returned to the client.
3. Validate the id is a 64-character lowercase hex sha256 at each handler before it reaches the store, returning 404 otherwise. Upload still computes the real digest and reports the mismatch as before.

Added a regression test that plants a file where `....` would resolve and confirms it is no longer served, plus checks for backslash and non-hex ids.

### DIFF
--- a/dulwich/lfs_server.py
+++ b/dulwich/lfs_server.py
@@ -29,12 +29,23 @@ __all__ = [
 
 import hashlib
 import json
+import re
 import tempfile
 import typing
 from collections.abc import Mapping
 from http.server import BaseHTTPRequestHandler, HTTPServer
 
 from .lfs import LFSStore
+
+# LFS object ids are SHA-256 digests, i.e. 64 lowercase hex characters. The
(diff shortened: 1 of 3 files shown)


netliomax25-code:
Looks like this landed directly as c6cfa65, with the regex swapped for the length + set-membership check in 06a1c64. That reads cleaner than what I had, no objection. Thanks for taking it.

PR #2247, opened 2026-06-19
reject internal addresses when fetching bundle uris

netliomax25-code:
1. fetch_bundle_uri validated only the URI scheme, so the entry URIs in a bundle list (served by an untrusted remote) were fetched with no check on the destination host.
2. apply_bundle_uri fetches each of those entries during clone(bundle_uri=...), and urllib3 follows redirects by default, so a hostile list (or a redirect from a public host) could drive GET requests at loopback, link-local 169.254.169.254, or private addresses (SSRF).
3. validate the resolved address at the fetch chokepoint, rejecting private/loopback/link-local/reserved/multicast/unspecified targets (including IPv4-mapped IPv6), and follow redirects manually so each hop is validated too.

Confirmed with a local server: before the change, fetch_bundle_uri("http://127.0.0.1:<port>/x") issued the request; after, it raises BundleURIError without connecting.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,10 @@
 1.2.7	2026-06-12
 
+ * SECURITY: Refuse to fetch a bundle URI that resolves to a loopback,
+   link-local, private or otherwise non-public address, and validate every
+   redirect hop, so a hostile bundle list can no longer point ``clone`` at
+   internal services (SSRF). (netliomax25-code)
+
  * Verify that an object retrieved by id actually hashes to the requested
    id, raising ``ChecksumMismatch`` otherwise. (Jelmer Vernooĳ, #2223)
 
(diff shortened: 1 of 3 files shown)


jelmer:
Dulwich current' behaviour matches c Git afaik, perhaps minus the redirect support but I don't think that significantly changes the picture. 

I dont think we should diverge. Have you contacted c git about this?

netliomax25-code:
Fair point, and no, I haven't raised it with the git project. I came at this from the dulwich side only.

You're right that the underlying behavior matches git, and I agree dulwich shouldn't go its own way on this unilaterally. The one genuine difference is the redirect handling: urllib3 follows redirects by default, so a public bundle host can bounce the fetch to an internal address without the bundle list itself ever looking hostile. But I take your point that on its own that's probably not enough to justify diverging.

I'll take it to the git list first and we can revisit here if they decide to add a check. Fine to close this in the meantime.

jelmer:
closing this for now

PR #2250, opened 2026-06-20
cap decompressed size of gzip request bodies

netliomax25-code:
1. `LimitedInputFilter` caps the gzip-encoded request body at Content-Length, but `GunzipFilter` then decompresses it with no bound on the output size.
2. DEFLATE expands by up to ~1000x, so a small compressed POST body (a git client sends `Content-Encoding: gzip` on upload-pack requests, reachable before authentication on a public repo) is a decompression bomb: a ~200 KiB body inflates to ~200 MiB that the handler reads into memory.

`GunzipFilter` now inflates the body up front, bounded at 10 MiB by default, and rejects anything larger with `413 Request Entity Too Large`. The check happens before the request is dispatched because handlers call `start_response` before reading the body, at which point a 413 can no longer be sent. Added regression tests and a NEWS entry.

Note c git doesn't cap the decompressed size itself; http-backend bounds the compressed body it buffers (`GIT_HTTP_MAX_REQUEST_BUFFER`) and streams the inflated bytes into a separate service process, while dulwich inflates in-process.

### DIFF
--- a/dulwich/web.py
+++ b/dulwich/web.py
@@ -26,7 +26,9 @@ __all__ = [
     "HTTP_FORBIDDEN",
     "HTTP_NOT_FOUND",
     "HTTP_OK",
+    "HTTP_REQUEST_ENTITY_TOO_LARGE",
     "MAX_CHUNK_SIZE",
+    "MAX_DECOMPRESSED_REQUEST_SIZE",
     "MAX_REQUEST_SIZE",
     "ChunkedEncodingError",
     "GunzipFilter",
@@ -160,6 +162,7 @@ logger = log_utils.getLogger(__name__)
 HTTP_OK = "200 OK"
 HTTP_NOT_FOUND = "404 Not Found"
 HTTP_FORBIDDEN = "403 Forbidden"
+HTTP_REQUEST_ENTITY_TOO_LARGE = "413 Request Entity Too Large"
 HTTP_ERROR = "500 Internal Server Error"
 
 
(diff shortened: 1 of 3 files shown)


netliomax25-code:
gentle ping

jelmer:
This shouldn't silently truncate; return a 413?

I'm on the fence about this PR; the web server (the outer layer) is really more of a demo than meant to be production ready, so perhaps the right call here is to just remove the input limiting support and add a warning about this at startup.

dulwich/web.py
```diff
@@ -159,6 +159,11 @@ def __call__(
 HTTP_FORBIDDEN = "403 Forbidden"
 HTTP_ERROR = "500 Internal Server Error"
 
+# Upper bound on the decompressed size of a gzip-encoded request body. A small
+# compressed body can otherwise inflate to gigabytes before the request is even
+# dispatched, so the cap mirrors git's GIT_HTTP_MAX_REQUEST_BUFFER (10 MiB).
+DEFAULT_MAX_DECOMPRESSED_REQUEST_SIZE = 10 * 1024 * 1024
```
jelmer: This isn't the case for git, from its manpage:

he GIT_HTTP_MAX_REQUEST_BUFFER environment variable (or the http.maxRequestBuffer config option) may be set to change the largest ref negotiation request that git will handle during a fetch; any fetch requiring a larger buffer will not succeed. This value should not normally need to be changed, but may be helpful if you are fetching from a repository with an extremely large number of refs. The value can be specified with a unit (e.g., 100M for 100 megabytes). The default is 10 megabytes.

As far as I can tell git doesn't have specific decompression protection. But happy to be corrected.

netliomax25-code: You're right, I had that backwards. http-backend caps the compressed body it buffers (via GIT_HTTP_MAX_REQUEST_BUFFER) and streams the inflated bytes into the service process without any decompressed-size limit. The difference is dulwich inflates in-process, so the decompressed bytes land in server memory rather than a child's stdin. I've reworded the comment to say that instead of claiming git has an equivalent cap; kept the 10 MiB default since legitimate gzipped negotiation bodies stay well under it.

netliomax25-code:
Reworked to return a 413 instead of truncating. One wrinkle: handlers call start_response before reading the body, so the filter now inflates the body up front (bounded at the cap) and rejects oversized requests before dispatch. Memory stays bounded at the cap either way. Also rebased onto main and added a NEWS entry.

On dropping the input limiting entirely: since the chunked-body bounding went into main I kept this consistent with that direction, but if you'd rather remove the limits and just warn at startup that the server isn't production ready, I can rework it that way.

jelmer (changes requested):
The web server is not for production use, so I don't consider this a security issue. Like the other PR, I think the proper fix here as well is to stream (and plumb that through) rather than to add arbitrary caps and pre-read all data.

netliomax25-code:
Makes sense. Working through what the streaming version looks like:

1. With the cap and the pre-read dropped, the filter is back to the lazy GzipFile wrap that main already has, which inflates on demand as the handler reads instead of buffering the body.
2. Downstream is already plumbed for that: receive-pack streams the pack to disk through add_thin_pack (bounded by receive.maxInputSize) and upload-pack parses pkt-lines off the stream through ReceivableProtocol, so neither holds the inflated body in memory.

So once the cap goes away this branch has no remaining delta against main, and I don't see a code change left to push here. Fine to close this one; if you'd still like the startup warning that the server isn't meant for production, I can send that separately.

jelmer:
Thanks, let's close this then. The server already warns that it's non-production at startup.

PR #2251, opened 2026-06-20
validate object ids received from the server

netliomax25-code:
A malformed object id advertised by a remote (e.g. one containing path separators like `../../config`) could be fed into object-store membership checks and stored as a ref value during fetch/clone.

1. `DiskObjectStore._get_shafile_path` joined the id straight into a filename via `hex_to_filename`. Because `os.path.join` drops the prefix once the tail is absolute, an id like `../../config` escaped the objects directory, letting a hostile server use a `sha in object_store` check as an existence/parse oracle for arbitrary local paths.
2. On the refs side, `RefsContainer.__setitem__` already validated values, but `set_if_equals`/`add_if_new` (used by `import_refs` during fetch) did not, so a bad advertised id was persisted as a ref.

Fix keeps the rejection in the layers that consume the id, not the protocol parser:

1. `_get_shafile_path` now rejects an id that is neither valid hex nor a raw oid before building the path.
2. `set_if_equals`/`add_if_new` validate the new value through a shared `_check_ref_value` helper (the same check `__setitem__` already used), so the object store and the refs container refuse malformed ids regardless of caller.

Added regression tests in `tests/test_object_store.py` and `tests/test_refs.py`.

### DIFF
--- a/dulwich/refs.py
+++ b/dulwich/refs.py
@@ -426,6 +426,18 @@ class RefsContainer:
             return
         raise RefFormatError(name)
 
+    def _check_ref_value(self, ref: ObjectID) -> None:
+        """Ensure a ref value is a valid object id or a symref.
+
+        Args:
+          ref: The value a ref is being set to.
+
+        Raises:
+          ValueError: if the value is neither a valid sha nor a symref.
+        """
+        if not (valid_hexsha(ref) or ref.startswith(SYMREF)):
+            raise ValueError(f"{ref!r} must be a valid sha (40 chars) or a symref")
+
     def read_ref(self, refname: Ref) -> bytes | None:
         """Read a reference without following any references.
 
@@ -549,8 +561,7 @@ class RefsContainer:
           name: The refname to set.
           ref: The new sha the refname will refer to.
         """
-        if not (valid_hexsha(ref) or ref.startswith(SYMREF)):
-            raise ValueError(f"{ref!r} must be a valid sha (40 chars) or a symref")
+        self._check_ref_value(ref)
         self.set_if_equals(name, None, ref)
 
     def remove_if_equals(
@@ -728,6 +739,7 @@ class DictRefsContainer(RefsContainer):
         Returns:
           True if the set was successful, False otherwise.
         """
+        self._check_ref_value(new_ref)
         if old_ref is not None and self._refs.get(name, ZERO_SHA) != old_ref:
             return False
         # Only update the specific ref requested, not the whole chain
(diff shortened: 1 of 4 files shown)


tests/test_client.py
```diff
@@ -690,6 +690,29 @@ def test_read_pkt_refs_v2_error(self):
             client.read_pkt_refs_v2([b"ERR An error occurred on the git server"])
         self.assertEqual("An error occurred on the git server", str(cm.exception))
 
+    def test_read_pkt_refs_v1_invalid_sha(self):
+        with self.assertRaises(errors.GitProtocolError):
+            client.read_pkt_refs_v1(
```
jelmer: It's fine for the server to reject invalid shas early.

However, it's up to the ObjectStore and RefsContainer to reject these, not at the Protocol layer.

jelmer (changes requested):

tests/test_client.py
netliomax25-code: Makes sense, moved it down a layer and dropped the client.py changes.

1. DiskObjectStore._get_shafile_path now rejects an id that is neither valid hex nor a raw oid before building the path, so a membership check (sha in object_store) can't be turned into a read outside objects/.
2. Ref values go through a shared _check_ref_value in set_if_equals/add_if_new (the same check __setitem__ already did), so import_refs during fetch won't persist a malformed value either.
3. Regression tests moved to test_object_store.py and test_refs.py accordingly.

dulwich/refs.py
```diff
@@ -426,6 +426,18 @@ class needs to be able to touch HEAD. Also, check_ref_format expects
...
+        """Ensure a ref value is a valid object id or a symref.
+
+        Args:
+          ref: The value a ref is being set to.
+
+        Raises:
+          ValueError: if the value is neither a valid sha nor a symref.
+        """
+        if not (valid_hexsha(ref) or ref.startswith(SYMREF)):
+            raise ValueError(f"{ref!r} must be a valid sha (40 chars) or a symref")
```
jelmer, line 439: hexshas don't have to be 40 characaters, the length can be different depending on the algorithm used

jelmer (changes requested):
This looks alright, but needs fixing of the CI and resolving of conflicts.

netliomax25-code:
Looks like this already landed on main as 9e2e64f, with the sha-length wording cleaned up in d4724c6 on top, so the CI/conflict and 40-char points are both handled. Closing this out since the branch is now identical to main.

PR #2254, opened 2026-06-22
escape repository path in ssh command to match git sq_quote

netliomax25-code:
1. `SSHGitClient._connect` wraps the repository path in single quotes but does not escape an embedded single quote, so a path containing a quote closes the quoting.
2. the SSH vendor passes that command string to the remote login shell, so the bytes after the closed quote run as shell. The path comes from the clone/fetch URL and is attacker-controllable, for example a malicious repo's `.gitmodules` `ssh://` submodule url during a recursive clone.

dulwich already rejects option-like (dash-prefixed) hosts but never escaped the path. Escaped embedded quotes the way git's `sq_quote` does (`'` becomes `'\\''`) so the remote shell receives the path verbatim. A path like `/repo'; touch pwned; '` no longer breaks out.

### DIFF
--- a/dulwich/client.py
+++ b/dulwich/client.py
@@ -3834,18 +3834,6 @@ class PLinkSSHVendor(SSHVendor):
 get_ssh_vendor: Callable[[], SSHVendor] = SubprocessSSHVendor
 
 
-def _quote_remote_path(path: bytes) -> bytes:
-    r"""Single-quote a repository path for the remote shell over SSH.
-
-    The git command we hand to the SSH vendor is run by the remote login
-    shell, so the path has to be quoted the way git does it (sq_quote): wrap
-    it in single quotes and rewrite every embedded single quote as ``'\\''``.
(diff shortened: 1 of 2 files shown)


dulwich/client.py
```diff
@@ -3990,9 +4002,8 @@ def _connect(
             path = path[1:]
         argv = (
             self._get_cmd_path(cmd)
-            + b" '"
-            + path.encode(self._remote_path_encoding)
```
jelmer, line 3994: why not just shlex.quote() ?

netliomax25-code: Good point. Switched to shlex.quote and dropped the helper.

One thing worth noting: shlex leaves all-safe paths unquoted, so a simple path now comes through as `git-command /path/to/repo` instead of being wrapped in quotes. It's shell-equivalent, but it did change the existing test_connect assertion, so I updated that one too.

PR #2299, opened 2026-07-11
detect delta cycles in resolve_object

netliomax25-code:
1. `Pack.resolve_object` walks a pack object's delta chain, but the only cycle guard rejects a single self-referencing REF_DELTA (`base_offset == prev_offset`), so a pack holding two or more REF_DELTA objects that name each other is not caught.
2. `get_raw` drives that walk, so reading such an object never returns: each iteration re-inflates a delta and appends to `delta_stack`, so memory grows without bound. A pack file is untrusted input (a cloned or downloaded repository, a submodule, a pack read from disk). `git index-pack` and dulwich's own streaming `DeltaChainIterator` both reject delta cycles; only this random-access path did not.
3. Track the REF_DELTA base offsets already followed on the chain and raise `UnresolvedDeltas` when one repeats. Every cycle crosses at least one REF_DELTA edge (OFS_DELTA offsets strictly decrease), so this is enough to break any cycle, and valid chains visit distinct offsets so their result is unchanged.

Verified with a two-object REF_DELTA cycle: before the change `get_raw` hangs; after it raises `UnresolvedDeltas`. Added a regression test covering that pack.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,13 @@
    land setuid. Modes are now run through ``cleanup_mode`` like checkout,
    matching git's archive writer.
 
+ * Detect delta cycles when resolving an object from a pack. The random-access
+   ``Pack.resolve_object`` walk only rejected a single self-referencing delta,
+   so a crafted pack with two or more REF_DELTA objects that name each other
+   sent ``get_raw`` into an unbounded loop that grew memory without bound. Such
+   chains now raise ``UnresolvedDeltas``, matching git's delta-cycle rejection
+   and the streaming resolver.
+
 1.2.10	2026-07-07
 
  * Fix regression in 1.2.9 where loose objects whose content inflates to
--- a/NEWS
+++ b/NEWS
@@ -15,12 +15,9 @@
    land setuid. Modes are now run through ``cleanup_mode`` like checkout,
    matching git's archive writer.
 
- * Detect delta cycles when resolving an object from a pack. The random-access
-   ``Pack.resolve_object`` walk only rejected a single self-referencing delta,
-   so a crafted pack with two or more REF_DELTA objects that name each other
-   sent ``get_raw`` into an unbounded loop that grew memory without bound. Such
-   chains now raise ``UnresolvedDeltas``, matching git's delta-cycle rejection
-   and the streaming resolver.
+ * Detect delta cycles in ``Pack.resolve_object``: a crafted pack with
+   REF_DELTA objects that name each other sent ``get_raw`` into an
+   unbounded loop. Such chains now raise ``DeltaCycle``.
 
 1.2.10	2026-07-07
 
--- a/dulwich/pack.py
+++ b/dulwich/pack.py
@@ -4510,6 +4510,9 @@ class Pack:
         base_type = type
         base_obj = obj
         delta_stack = []
+        # Offsets of REF_DELTA bases already followed on this chain, used to
+        # detect delta cycles.
+        seen_ref_offsets: set[int] = {offset}
         while base_type in DELTA_TYPES:
             prev_offset = base_offset
             if get_ref is None:
@@ -4539,6 +4542,16 @@ class Pack:
                 base_offset = base_offset_temp
                 if base_offset == prev_offset:  # object is based on itself
                     raise UnresolvedDeltas([basename])
+                # A REF_DELTA base already followed on this chain means the
+                # deltas form a cycle (e.g. two objects that delta against each
+                # other). Every cycle crosses at least one REF_DELTA edge, so
+                # tracking resolved REF base offsets is enough to break it.
+                # Without this the loop spins forever while delta_stack grows
+                # without bound; git rejects such packs.
+                if base_offset is not None:
+                    if base_offset in seen_ref_offsets:
+                        raise UnresolvedDeltas([basename])
+                    seen_ref_offsets.add(base_offset)
             else:
                 raise AssertionError(f"Unexpected delta type: {base_type}")
             delta_stack.append((prev_offset, base_type, delta))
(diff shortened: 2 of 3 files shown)


jelmer:
What does C Git do here?

netliomax25-code:
Two separate paths in C Git:

1. At index time (`index-pack` / `unpack-objects`) it detects REF_DELTA cycles and refuses the pack. That's covered by `t/t5309-pack-delta-cycles.sh` ("index-pack detects REF_DELTA cycles"), so a cyclic pack doesn't reach the object store through normal channels.

2. The random-access read path (`unpack_entry` in packfile.c) is the equivalent of `resolve_object`/`get_raw`, and as far as I can tell it has no cycle guard of its own. The phase-1 drill-down just follows the resolved base offset and relies on the index-time check having already rejected cycles. Point it at a hand-crafted cyclic pack and it loops the same way we do.

Dulwich already matches (1): the streaming `DeltaChainIterator` rejects cycles via `_ensure_no_pending` -> `UnresolvedDeltas`. What this PR adds is a guard on (2) for `get_raw`, which reads packs straight from their shipped .idx without necessarily re-resolving them, so it's a bit more exposed than C Git's read path. The check is cheap and only trips on chains that were already invalid, so valid packs are unaffected.

jelmer:
does this address all codepaths that resolve deltas?

netliomax25-code:
I went through everything that follows a delta chain to check.

1. `Pack.resolve_object` is the only iterative chain walk on the random-access side, and `Pack.get_raw` is its sole caller (`Pack.__getitem__` and the object store's `get_raw` route through it). That's the path this PR guards.

2. Every other resolver goes through `DeltaChainIterator`: `PackIndexer` (index creation via `PackData.iterentries`/`create_index`, thin-pack completion), `PackInflater` (`iterobjects`, `Pack.check`, `unpack_objects`, pack ingestion) and `UnpackedObjectIterator`. That walk resolves bases before children, so members of a cycle never unblock; they stay in `_pending_ref` and `_ensure_no_pending` raises `UnresolvedDeltas`. `test_ext_ref_deltified_object_based_on_itself` already covers the self-referencing case there.

3. OFS_DELTA can't cycle on either path: base offsets strictly decrease, and `unpack_object` rejects a zero base offset at read time.

4. `iter_unpacked_subset` and `get_unpacked_object` don't apply deltas at all (they relabel OFS to REF in a single pass), so there's no walk to trap.

The one theoretical leftover is a cycle split across two packs, which needs `resolve_ext_ref` wired to the store's `get_raw`. Store packs are opened without it, and the paths that do wire it up are the DeltaChainIterator ones above, so I don't believe it's reachable in practice; constructed by hand it recurses and dies with a RecursionError rather than hanging. Can add a guard for that too if you'd like, but it seemed out of scope here.

test_pack.py is green on the branch with the regression test included.

netliomax25-code:
any update?

dulwich/pack.py
```diff
@@ -4539,6 +4542,16 @@ def resolve_object(
...
                     raise UnresolvedDeltas([basename])
+                # A REF_DELTA base already followed on this chain means the
+                # deltas form a cycle (e.g. two objects that delta against each
+                # other). Every cycle crosses at least one REF_DELTA edge, so
+                # tracking resolved REF base offsets is enough to break it.
+                # Without this the loop spins forever while delta_stack grows
+                # without bound; git rejects such packs.
+                if base_offset is not None:
+                    if base_offset in seen_ref_offsets:
+                        raise UnresolvedDeltas([basename])
```
jelmer: Shouldn't this be a dedicated exception, e.g. DeltaCycle?

dulwich/pack.py
```diff
@@ -4510,6 +4510,9 @@ def resolve_object(
         base_type = type
         base_obj = obj
         delta_stack = []
+        # Offsets of REF_DELTA bases already followed on this chain, used to
```
jelmer: This comment seems superfluous

dulwich/pack.py
```diff
@@ -4539,6 +4542,16 @@ def resolve_object(
                 base_offset = base_offset_temp
                 if base_offset == prev_offset:  # object is based on itself
                     raise UnresolvedDeltas([basename])
+                # A REF_DELTA base already followed on this chain means the
+                # deltas form a cycle (e.g. two objects that delta against each
+                # other). Every cycle crosses at least one REF_DELTA edge, so
+                # tracking resolved REF base offsets is enough to break it.
+                # Without this the loop spins forever while delta_stack grows
+                # without bound; git rejects such packs.
```
jelmer: didn't we establish that git doesn't reject such packs?

NEWS
```diff
@@ -15,6 +15,13 @@
    land setuid. Modes are now run through ``cleanup_mode`` like checkout,
    matching git's archive writer.
 
+ * Detect delta cycles when resolving an object from a pack. The random-access
```
jelmer: Please keep news entries to 3-4 lines

jelmer (changes requested):

dulwich/pack.py
netliomax25-code: Done, added a DeltaCycle exception. Made it a subclass of UnresolvedDeltas so anything already catching that keeps working.

dulwich/pack.py
netliomax25-code: Dropped.

dulwich/pack.py
netliomax25-code: Right, index-pack rejects them but the read path doesn't. Reworded the comment to drop that claim (and the same wording in the test).

NEWS
netliomax25-code: Shortened to 3 lines.

PR #2309, opened 2026-07-16
validate object ids in receive-pack ref commands

netliomax25-code:
1. ReceivePackHandler.handle splits each client ref-update line with ref_line.split() and wraps the fields in ObjectID()/Ref() without validating them, while the sibling upload-pack parser _split_proto_line checks valid_hexsha and git's receive-pack validates each id with parse_oid_hex.
2. A line that does not split into three fields raises an uncaught ValueError that aborts the handler, and a non-hex old/new value is accepted into client_refs and passed to the pre/post-receive and update hooks and to set_if_equals.
3. Check the field count and both object ids with valid_hexsha, raising GitProtocolError on malformed input, matching the sibling parser.

Added receive-pack handler tests in tests/test_server.py for the malformed-line, non-hex-id, and valid-delete cases.

### DIFF
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -1744,7 +1744,16 @@ class ReceivePackHandler(PackHandler):
 
         # client will now send us a list of (oldsha, newsha, ref)
         while ref_line:
-            (oldsha, newsha, ref_name) = ref_line.split()
+            fields = ref_line.rstrip(b"\n").split(b" ")
+            if (
+                len(fields) != 3
+                or not valid_hexsha(fields[0])
+                or not valid_hexsha(fields[1])
+            ):
+                raise GitProtocolError(
+                    f"Invalid ref update line from client: {ref_line!r}"
+                )
+            oldsha, newsha, ref_name = fields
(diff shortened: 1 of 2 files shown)


dulwich/server.py
```diff
@@ -1744,7 +1744,16 @@ def handle(self) -> None:
...
+            fields = ref_line.rstrip(b"\n").split(b" ")
+            if (
+                len(fields) != 3
+                or not valid_hexsha(fields[0])
+                or not valid_hexsha(fields[1])
+            ):
+                raise GitProtocolError(
+                    f"Invalid ref update line from client: {ref_line!r}"
+                )
+            oldsha, newsha, ref_name = fields
```
jelmer: please ask for forgivness, not permission; i.e. just catch the ValueError, don't check for the number of items.

netliomax25-code: Makes sense, done. The unpack now just catches the ValueError and re-raises as GitProtocolError; the valid_hexsha checks on oldsha/newsha stay as a separate check since the unpack can't cover those.

PR #2315, opened 2026-07-19
bound inflated size of dumb HTTP loose objects

netliomax25-code:
DumbHTTPObjectStore._fetch_loose_object inflated a dumb HTTP server's loose-object response with a single unbounded zlib.decompress, so a hostile or intercepted server could expand a small reply into gigabytes of memory during clone or fetch. Following review, the response is now streamed the way C git's dumb-http walker handles fetched loose objects: inflated incrementally through a bounded buffer, spooled to a temporary file past 1 MiB, and hashed as it goes, with the content only returned once the stream hashes to the requested object id. No size limit is involved, matching C git. Added regression tests for the sha-mismatch, truncated and larger-than-spool cases.

### DIFF
--- a/dulwich/dumb.py
+++ b/dulwich/dumb.py
@@ -45,7 +45,6 @@ from .errors import NotGitRepository, ObjectFormatException
 from .object_store import BaseObjectStore
 from .objects import (
     ZERO_SHA,
-    Blob,
     Commit,
     ObjectID,
     RawObjectID,
@@ -53,6 +52,7 @@ from .objects import (
     Tag,
     Tree,
     hex_to_sha,
+    object_class,
     sha_to_hex,
 )
 from .pack import Pack, PackData, PackIndex, UnpackedObject, load_pack_index_file
@@ -60,6 +60,38 @@ from .protocol import split_peeled_refs
 from .refs import Ref, read_info_refs
 
 logger = logging.getLogger(__name__)
+
+# Inflated loose objects larger than this spill to disk while streaming.
+_LOOSE_OBJECT_SPOOL_SIZE = 1024 * 1024
+
+
+def _inflate_chunks(
+    read: Callable[..., bytes], read_size: int = 4096, max_length: int = 65536
+) -> Iterator[bytes]:
+    """Incrementally inflate a zlib stream read in chunks.
+
+    Yields decompressed chunks of at most max_length bytes, so memory use
+    stays bounded regardless of the stream's compression ratio.
+
+    Raises:
+      zlib.error: If the stream is corrupt or truncated.
+    """
+    decompressor = zlib.decompressobj()
+    while not decompressor.eof:
+        chunk = read(read_size)
+        if not chunk:
+            break
+        buf = decompressor.decompress(chunk, max_length)
+        while buf:
+            yield buf
+            if not decompressor.unconsumed_tail:
+                break
+            buf = decompressor.decompress(decompressor.unconsumed_tail, max_length)
+    tail = decompressor.flush()
+    if tail:
+        yield tail
+    if not decompressor.eof:
+        raise zlib.error("incomplete or truncated zlib stream")
 
 
 class DumbHTTPObjectStore(BaseObjectStore):
@@ -125,6 +157,12 @@ class DumbHTTPObjectStore(BaseObjectStore):
     def _fetch_loose_object(self, sha: bytes) -> tuple[int, bytes]:
         """Fetch a loose object by SHA.
 
+        The response is inflated incrementally through a bounded buffer and
+        spooled to a temporary file, hashing as it goes, so a malicious dumb
+        server cannot expand a small response into an unbounded amount of
+        memory. The content is only loaded once the stream has hashed to the
+        requested object ID, like C git's dumb-http walker.
+
         Args:
           sha: SHA1 of the object (hex string as bytes)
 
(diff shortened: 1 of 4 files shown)


dulwich/dumb.py
```diff
@@ -142,8 +148,10 @@ def _fetch_loose_object(self, sha: bytes) -> tuple[int, bytes]:
         except OSError:
             raise KeyError(sha)
 
-        # Decompress and parse the object
```
jelmer, line 145: This should probably use the regular ShaFile decompression instead, so we get to use e.g. the git config for reading files. That's better from a code reuse perspective as well.

dulwich/dumb.py
```diff
@@ -72,6 +73,7 @@ def __init__(
             [str, dict[str, str]], tuple[Any, Callable[..., bytes]]
         ],
         object_format: "ObjectFormat | None" = None,
+        max_object_size: int = DEFAULT_LOOSE_OBJECT_SIZE_LIMIT,
```
jelmer: this should read from git config, not hardcode somtehing.

dulwich/dumb.py
netliomax25-code: Done, _fetch_loose_object now parses the response with ShaFile.from_file, which also let me drop the hand-rolled header parsing and type map here.

dulwich/dumb.py
netliomax25-code: Switched to reading core.bigFileThreshold, same as DiskObjectStore.from_config, with the 512 MiB default as fallback. The HTTP client passes its config down through DumbRemoteHTTPRepo, and the test now sets the threshold via config.

jelmer: Does that really match what C git does?

netliomax25-code: Not exactly, no. Two things C git does differently here:

1. In C git, core.bigFileThreshold is a streaming/delta heuristic (store deflated, skip delta compression, treat as binary), not a hard limit; nothing gets rejected for exceeding it.
2. Its dumb-http walker doesn't need a cap at all: fwrite_sha1_file streams the response through git_inflate into a tempfile while hashing, and finish_http_object_request then verifies the hash against the requested oid, so memory stays bounded regardless of object size.

Since our dumb store keeps the fetched object fully in memory, a cap seemed like the practical equivalent, and I reused core.bigFileThreshold because that's the bound 43dfeb5 already established for on-disk loose objects (ShaFile.from_file's max_size, read from config in DiskObjectStore.from_config). If you'd rather match C git properly here, I can rework _fetch_loose_object to stream-inflate to a temp file and verify the sha instead of capping; that's a bigger change but closer to what C git actually does.

netliomax25-code: Not exactly. C git doesn't bound this path at all: its dumb-http walker writes the compressed body straight to a local file and inflates through a fixed 4 KiB scratch buffer just to compute the hash (http.c, fwrite_sha1_file), so its memory use stays constant without any cap. Since this store buffers the whole inflated object in memory, I reused the core.bigFileThreshold bound that 43dfeb5 introduced for on-disk loose objects, so both loose read paths fail the same way. If you'd rather match C git exactly here, the alternative is streaming the inflate to a temp file with bounded memory, but that's a bigger change to how this store hands objects around. Happy to go either way.

dulwich/client.py
```diff
@@ -4965,7 +4965,9 @@ def fetch_pack(
 
             # Pass http_request function
             with DumbRemoteHTTPRepo(
-                url, functools.partial(self._http_request, raise_for_status=False)
+                url,
+                functools.partial(self._http_request, raise_for_status=False),
+                config=getattr(self, "config", None),
```
jelmer: using getattr is usually a bad smell, this should probably just be self.config

dulwich/dumb.py
jelmer: I think properly streaming is the right approach here.

netliomax25-code: Done, reworked it to stream the way C git's walker does:

1. The response is now inflated incrementally through a bounded buffer (4 KiB reads, 64 KiB inflate steps) and spooled to a temporary file past 1 MiB, so memory stays bounded regardless of the compression ratio. The size cap and the config plumbing are gone.
2. The stream is hashed as it inflates and the content is only returned once it hashes to the requested object id, like finish_http_object_request. A 50 KiB body inflating to 50 MiB now gets rejected with about 1.4 MiB peak memory.
3. Also rebased onto main to resolve the NEWS conflict. test_client's dumb HTTP fixture advertised a made-up sha for its blob, which the verification now rejects, so it uses the blob's real id.

dulwich/client.py
netliomax25-code: That line is gone now: the streaming rework removed the size cap, so there's no config to read anymore and this call site is back to what main has.

dulwich/dumb.py
netliomax25-code: Done, reworked to stream the way C git's walker does:

1. The response now inflates incrementally through a bounded buffer, spooling the content to a temporary file past 1 MiB, so memory use stays bounded no matter how large the response inflates.
2. The inflated stream is hashed as it arrives, and the content is only handed back once it hashes to the requested object id (matching fwrite_sha1_file / finish_http_object_request); a substituted or truncated object is rejected instead.
3. With that, the core.bigFileThreshold cap wasn't adding anything, so the config plumbing is gone again.

Also rebased on main to resolve the NEWS conflict. test_dumb.py, test_client.py and test_object_store.py pass locally and CI is green.

dulwich/client.py
netliomax25-code: Agreed. fetch_pack lives on AbstractHttpGitClient, which doesn't define self.config (only Urllib3HttpGitClient does), which is why I'd reached for getattr. The streaming rework removed the need to pass config down at all, so that call is back to its original form.

PR #2317, opened 2026-07-21
validate shallow object ids before writing .git/shallow

netliomax25-code:
1. update_shallow writes each server-advertised shallow id to .git/shallow as `sha + b"\n"`, and get_shallow reads it back, neither checking the id is well-formed hex.
2. the ids arrive via _read_shallow_updates, which parses a `shallow <sha>` pkt-line with sha.strip(); pkt-lines are length-framed, so one packet can carry an internal newline that strip() leaves in place, letting a hostile server forge an extra graft line in .git/shallow (parse_graftpoints then reads it as a parent-less commit, hiding history).
3. validate every shallow/unshallow id with check_hexsha in update_shallow before the write and in get_shallow on read, matching git's get_oid_hex on shallow lines and the sibling parse_graftpoints reader. The object-store and refs id checks don't cover this file.

Verified with a fetch-shaped reproducer: a `shallow <40hex>\n<40hex>` packet wrote two lines to .git/shallow (a forged graft that survived reopen) before the change, and now raises. Existing shallow round-trip tests still pass; added regression tests for the write and read paths.


jelmer:
While not invalid to add validation for hardening here, this needs to be fixed in the client.

netliomax25-code:
Makes sense.

1. Moved the real fix into the client: _read_shallow_updates now rejects a non-hex shallow/unshallow id with GitProtocolError, same valid_hexsha pattern as the receive-pack validation in #2309, so a malformed id never reaches update_shallow.
2. Kept the repo.py checks as hardening per your note, and added client tests covering the embedded-newline packet, a non-hex id, and the normal 40/64-hex round-trip.

jelmer:
FWIW this is not a security issue, the server can control what commits are considered shallow anyway.

NEWS
```diff
@@ -1,3 +1,17 @@
+1.2.13	UNRELEASED
+
+ * SECURITY: Validate shallow object ids received from the remote.
```
jelmer: Please keep this brief - ideally three lines, maybe more if absolutely necessary. Also, this should be HARDEN rather than SECURITY: the server can control what is considered shallow anyway.

PR #2323, opened 2026-07-29
don't grant other write for core.sharedRepository=all

netliomax25-code:
1. `parse_shared_repository` maps `core.sharedRepository` = `all`/`world`/`everybody`/`2` to `(0o666, 0o2777)`, so every path it governs is chmodded writable by any user on the machine.
2. those masks reach the control dir, loose objects, packs, refs, the index and reflogs through `_init_maybe_bare` and `DiskObjectStore.from_config`, so `.git/hooks` and `.git/refs/heads` end up `02777` and a local user can drop a hook or rewrite a ref in someone else's repository.

Returned git's `PERM_EVERYBODY` values instead (`0o664` file, `0o2775` directory). Others keep the read bit that separates `all` from `group`, and execute on directories, but not write. The explicit-octal branch is left alone since that is stated user intent.

Checked against git 2.39: `git init --shared=all` gives `775` for `.git/hooks`, `.git/refs` and `.git/objects`, and `664` for `.git/config`. In `path.c`, `PERM_EVERYBODY` is `0664` and the directory tweak only copies read bits into execute bits (`tweak |= (tweak & 0444) >> 2`), so write for others is never set.

Before the change, with umask 0:

```
$ python -c "from dulwich.repo import Repo; Repo.init('/tmp/r', shared_repository='all')"
$ stat -c '%a %n' /tmp/r/.git/hooks /tmp/r/.git/refs/heads /tmp/r/.git/objects
2777 /tmp/r/.git/hooks
2777 /tmp/r/.git/refs/heads
2777 /tmp/r/.git/objects
```

After it, the same walk finds no world-writable entry under `.git`.

Two existing tests in `SharedRepositoryTests` asserted the old masks and are updated. Added `test_init_bare_shared_all_not_world_writable`, which fails on the unpatched tree with `hooks is world-writable: 0o2777`. `tests/test_repository.py` and `tests/test_object_store.py` pass, and ruff check, ruff format and codespell are clean.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,13 @@
 1.2.13	UNRELEASED
 
+ * SECURITY: Stop granting write permission to other users for
+   ``core.sharedRepository = all``. ``parse_shared_repository`` returned masks
+   of ``0o666``/``0o2777``, so the control directory, loose objects, packs,
+   refs, the index and reflogs were chmodded world-writable, letting any local
+   user drop a hook or rewrite a ref in the repository. Git's
+   ``PERM_EVERYBODY`` grants others read, and execute on directories, but never
+   write. (Kartik Kenchi)
+
  * Always single-quote the repository path in the SSH command, matching git's
    ``sq_quote()``. ``shlex.quote`` left paths without shell metacharacters
    bare, which broke cloning from servers that parse the command themselves
--- a/dulwich/repo.py
+++ b/dulwich/repo.py
@@ -422,8 +422,9 @@ def parse_shared_repository(
         return (0o664, 0o2775)
 
     if value_lower in ("all", "world", "everybody", "2"):
-        # World readable/writable (with setgid bit)
-        return (0o666, 0o2777)
+        # World readable (with setgid bit). Like git's PERM_EVERYBODY, others
+        # gain read, and execute on directories, but never write.
+        return (0o664, 0o2775)
 
     if value_lower == "umask":
         # Explicitly use umask
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -2246,11 +2246,31 @@ class SharedRepositoryTests(TestCase):
         repo = Repo.init_bare(tmp_dir, shared_repository="all")
         self.addCleanup(repo.close)
 
-        # Expected permissions for world sharing
-        expected_dir_mode = 0o2777  # setgid + rwxrwxrwx
-        expected_file_mode = 0o666  # rw-rw-rw-
+        # Expected permissions for world sharing. Others get read, and
+        # execute on directories, but no write bit.
+        expected_dir_mode = 0o2775  # setgid + rwxrwxr-x
+        expected_file_mode = 0o664  # rw-rw-r--
 
         self._check_permissions(repo, expected_file_mode, expected_dir_mode)
+
+    def test_init_bare_shared_all_not_world_writable(self):
+        """Test that sharedRepository=all never grants write to others."""
+        tmp_dir = tempfile.mkdtemp()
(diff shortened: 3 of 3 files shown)


jelmer:
Thanks for digging into this - the underlying problem is real. Under a
default umask 022, Repo.init_bare(..., shared_repository="all") gives
.git/hooks mode 2777 where git gives 2775, and a world-writable hooks
directory is a genuine local privilege-escalation vector. 

That said, I don't think the diagnosis is right, and I'd rather not take the fix in this form.

PERM_EVERYBODY in path.c is 0002, not 0664 - it's an adjustment value that
gets ORed in, not a target mode. The reason you saw 775 and 664 from git init
--shared=all is your umask, not the constant. With umask 0, git 2.55 gives
me:

```
$ (umask 0; git init --bare --shared=all /tmp/gt_all)
$ stat -c '%a %n' /tmp/gt_all/hooks /tmp/gt_all/refs /tmp/gt_all/objects
/tmp/gt_all/config
2777 /tmp/gt_all/hooks
2777 /tmp/gt_all/refs
2777 /tmp/gt_all/objects
666  /tmp/gt_all/config
```

which is exactly what dulwich produces. Could you re-run your 2.39 comparison under umask 0? I expect you'll see the same.

The actual divergence is in how the mask is applied. adjust_shared_perm only
ever loosens: it ORs the shared bits onto the mode the umask already produced
and never restores a bit the umask stripped. Dulwich instead calls
os.chmod(dir_path, dir_mode) (repo.py:2347), which sets the mode absolutely
and bypasses the umask entirely. So under umask 022 git lands on 2775 because
the umask removed other-write, while dulwich forces 2777 regardless.

Hardcoding the constant down to (0o664, 0o2775) gets the right answer for the
common umask by coincidence, but it has some side effects I'd want to avoid:

- all and group become byte-identical, which erases the distinction git
  draws. Under umask 077, git's --shared=all gives 2775 and --shared=group
  gives 2770; after this patch dulwich would give 2775 for both. That breaks
  the actual use case for all — a repo readable by a web server or another uid.
- The umask bug survives everywhere else. parse_shared_repository also feeds
  DiskObjectStore.from_config, _put_named_file, reflog writing and the index,
  and the explicit-octal branch still chmods absolutely - so sharedRepository =
  0666 still produces a world-writable tree. Git applies the same loosen-only
  rule there, so I don't think that branch should be exempt.
- test_init_bare_shared_all_not_world_writable passes only because the
  constant was lowered. A correct umask-respecting implementation would give
  2777 under umask 0, matching git, and this test would fail it - so it locks
  in the wrong model.

What I'd suggest instead: keep all at (0o666, 0o2777) as the ceiling and
change how it's applied, mirroring adjust_shared_perm - read the current
mode, OR in the shared bits, then mask off what the process umask would have
stripped. Ideally that lives in one helper used by all the call sites, which
fixes all, group and the explicit-octal branch together and preserves the
read-vs-write distinction.

  That's a bigger change than what you've got here — let me know if you'd like
  to take it on, and I'm happy to review. Either way the NEWS entry is worth
  keeping, just reworded around the umask handling rather than the constant.

  Two smaller notes for whatever form the fix takes:

  - os.umask(0) in the new test is never restored, so it leaks into every test
  that runs after it in the same process. It needs an addCleanup to put the old
  value back.
  - Worth a test that asserts the all/group difference under a restrictive
  umask, since that's the property most at risk here.

jelmer (changes requested):
see last comment

jelmer:
superseded by https://github.com/jelmer/dulwich/pull/2346

PR #2332, opened 2026-08-01
Add Pack.unkeep() to remove a pack's .keep file

bojanz:
Pack.keep() creates a .keep file to stop git from garbage collecting a pack, but there is no supported way to remove one again, so callers have to reconstruct the path from the private _basename attribute and unlink it themselves.

This comes up whenever a .keep file is used for its intended purpose of protecting a pack only while it is being wired up: a server receiving a push keeps the incoming pack so a concurrent gc cannot delete objects the refs do not point at yet, then drops the .keep once the refs are updated.

Add the counterpart to keep(). Removing a .keep file that does not exist is not treated as an error, since the common use is an unconditional cleanup in a finally block; the return value distinguishes the two cases for callers that care.

### DIFF
--- a/dulwich/pack.py
+++ b/dulwich/pack.py
@@ -4504,6 +4504,20 @@ class Pack:
                 keepfile.write(b"\n")
         return keepfile_name
 
+    def unkeep(self) -> bool:
+        """Remove the .keep file for the pack, allowing git to garbage collect it.
+
+        This is the counterpart of :meth:`keep`. It is not an error to call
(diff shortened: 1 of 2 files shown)


bojanz:
Long time users and fans of Dulwich at Upsun. We wrote about it recently in https://developer.upsun.com/posts/how-it-works/write-a-git-server-with-per-branch-permissions-in-five-minutes

This PR is one of several changes we've been running internally.

jelmer (approved):

PR #2333, opened 2026-08-01
Invalidate the packed-refs cache when the file is replaced.

bojanz:
DiskRefsContainer cached packed refs for its entire lifetime. If another process ran git gc or git pack-refs, the container could continue returning refs that had been added, updated, or removed.
    
Track the identity of the packed-refs file and reload the cache when the file is created, removed, or replaced. Read the identity from the open file descriptor so a concurrent rename cannot associate cached contents with the wrong file.
    
Also clear the cache after Dulwich writes packed-refs, including when a write fails, so the next read reflects the file currently on disk.

### DIFF
--- a/dulwich/refs.py
+++ b/dulwich/refs.py
@@ -857,6 +857,26 @@ class DictRefsContainer(RefsContainer):
         self._peeled.update(peeled)
 
 
+#: Identity of a particular packed-refs file, used to detect that the file a
+#: cache was populated from has since been replaced.
+_PackedRefsKey = tuple[int, int, int, int, int]
+
+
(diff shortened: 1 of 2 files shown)


jelmer (approved):
LGTM but please rebase on main

bojanz:
@jelmer Rebased, thank you!

bojanz:
@jelmer Should I add a NEWS entry for this? I originally kept those changes out of PRs to minimize conflicts (and because I wasn't sure which changes you'd want to accept).

(I am also assuming that the two previous PRs are not NEWS-worthy but happy to be corrected)

PR #2334, opened 2026-08-01
Support uploadpack.allow*SHA1InWant on the server

bojanz:
Honor uploadpack.allowTipSHA1InWant, uploadpack.allowReachableSHA1InWant, and uploadpack.allowAnySHA1InWant when validating unadvertised wants, and advertise the corresponding protocol capabilities. All three remain disabled by default.
    
Batch and deduplicate reachability checks to avoid repeated graph walks. Treat only exact ref object IDs as tips, and leave the feature disabled for custom backends without configuration access.

### DIFF
--- a/dulwich/server.py
+++ b/dulwich/server.py
@@ -112,6 +112,8 @@ from .pack import ObjectContainer, write_pack_from_container
 from .protocol import (
     CAPABILITIES_REF,
     CAPABILITY_AGENT,
+    CAPABILITY_ALLOW_REACHABLE_SHA1_IN_WANT,
+    CAPABILITY_ALLOW_TIP_SHA1_IN_WANT,
     CAPABILITY_ATOMIC,
     CAPABILITY_DELETE_REFS,
     CAPABILITY_FILTER,
@@ -162,6 +164,9 @@ from .refs import Ref, RefsContainer
 from .repo import Repo
 
 logger = log_utils.getLogger(__name__)
+
+# Maximum number of nested tag objects to follow when peeling to a commit.
+_MAX_TAG_PEEL_DEPTH = 100
 
 
 class Backend:
@@ -455,6 +460,29 @@ class PackHandler(Handler):
         self._done_received = True
 
 
+def _peel_to_commit(object_store: "BaseObjectStore", sha: ObjectID) -> ObjectID | None:
+    """Peel an object down to a commit.
+
+    Args:
+      object_store: Object store to look objects up in.
+      sha: Object ID to peel.
+    Returns: The object ID of the commit, or None if sha is missing or does
+        not peel to a commit.
+    """
+    # Bounded so that a cycle of tag objects cannot spin forever.
+    for _ in range(_MAX_TAG_PEEL_DEPTH):
+        try:
+            obj = object_store[sha]
+        except KeyError:
+            return None
+        if obj.type_name == b"commit":
+            return sha
+        if obj.type_name != b"tag":
+            return None
+        sha = obj.object[1]  # type: ignore[attr-defined]
+    return None
+
+
 class UploadPackHandler(PackHandler):
     """Protocol handler for uploading a pack to the client."""
 
@@ -485,6 +513,8 @@ class UploadPackHandler(PackHandler):
         self._processing_have_lines = False
         # Filter specification for partial clone support
         self.filter_spec: FilterSpec | None = None
+        self._sha1_in_want_policy: tuple[bool, bool, bool] | None = None
+        self._sha1_in_want_ref_tips: set[ObjectID] | None = None
 
     def capabilities(self) -> list[bytes]:
         """Return the list of capabilities supported by upload-pack.
@@ -492,7 +522,7 @@ class UploadPackHandler(PackHandler):
         Returns:
             List of capabilities including object-format for the repository
         """
-        return [
+        caps = [
             CAPABILITY_MULTI_ACK_DETAILED,
             CAPABILITY_MULTI_ACK,
             CAPABILITY_SIDE_BAND_64K,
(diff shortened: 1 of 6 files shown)


dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
+        return caps
+
+    def _get_sha1_in_want_policy(self) -> tuple[bool, bool, bool]:
+        """Read the uploadpack.allow*SHA1InWant settings for this repository.
+
+        Returns: A (any, reachable, tip) tuple of booleans. Setting
+            ``allowAnySHA1InWant`` implies the other two, as in git.
+        """
+        if self._sha1_in_want_policy is None:
+            get_config_stack = getattr(self.repo, "get_config_stack", None)
```
jelmer: Please avoid getattr

dulwich/server.py
```diff
@@ -455,6 +460,29 @@ def notify_done(self) -> None:
...
+def _peel_to_commit(object_store: "BaseObjectStore", sha: ObjectID) -> ObjectID | None:
+    """Peel an object down to a commit.
+
+    Args:
+      object_store: Object store to look objects up in.
+      sha: Object ID to peel.
+    Returns: The object ID of the commit, or None if sha is missing or does
+        not peel to a commit.
+    """
+    # Bounded so that a cycle of tag objects cannot spin forever.
```
jelmer: Pretty sure we already have this implemented somewhere.

dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
+                continue
+            seen.add(commit_id)
+            matching_wants = targets.pop(commit_id, None)
+            if matching_wants is not None:
+                reachable.update(matching_wants)
+            try:
+                commit = object_store[commit_id]
+            except KeyError:
+                continue
+            pending.extend(commit.parents)  # type: ignore[attr-defined]
```
jelmer: Please use assert instance(commit, Commit) rather than overriding type checks

dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
+        allowed = existing & tips
+        if not allow_reachable:
+            return allowed
+        allowed.update(self._reachable_wants(existing - allowed, tips))
+        return allowed
+
+    def _ref_tips(self) -> set[ObjectID]:
+        """Return the exact object IDs at the tips of every ref."""
+        if self._sha1_in_want_ref_tips is None:
+            self._sha1_in_want_ref_tips = set(self.repo.get_refs().values())
```
jelmer: How is this meant to work with annotated tags?

dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
+        Returns: A (any, reachable, tip) tuple of booleans. Setting
+            ``allowAnySHA1InWant`` implies the other two, as in git.
+        """
+        if self._sha1_in_want_policy is None:
+            get_config_stack = getattr(self.repo, "get_config_stack", None)
+            if not callable(get_config_stack):
+                # BackendRepo intentionally does not require configuration
+                # access. Preserve compatibility with custom backends by
+                # leaving all opt-in policies disabled when it is unavailable.
+                self._sha1_in_want_policy = (False, False, False)
```
jelmer: Please don't use anonymous tuples like this, they're not very readable

dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
+        commit graph to answer a single want, which is why git leaves it off
+        by default.
+
+        Args:
+          sha: Object ID the client asked for.
+        Returns: True if the object may be served to the client.
+        """
+        return sha in self.allowed_wants({sha})
+
+    def allowed_wants(self, wants: set[ObjectID]) -> set[ObjectID]:
```
jelmer: Is this really walking the entire graph for every want?

dulwich/server.py
```diff
@@ -505,6 +535,133 @@ def capabilities(self) -> list[bytes]:
...
             capability_object_format(self.repo.object_format.name),
         ]
+        _allow_any, allow_reachable, allow_tip = self._get_sha1_in_want_policy()
+        if allow_tip:
+            caps.append(CAPABILITY_ALLOW_TIP_SHA1_IN_WANT)
+        if allow_reachable:
+            caps.append(CAPABILITY_ALLOW_REACHABLE_SHA1_IN_WANT)
+        return caps
+
+    def _get_sha1_in_want_policy(self) -> tuple[bool, bool, bool]:
```
jelmer: Why can't this apply policy during the fetch process?

jelmer (changes requested):

dulwich/server.py
bojanz: Indeed, replaced by peel_sha().

dulwich/server.py
bojanz: allowTipSHA1InWant accepts only the tag object referenced by the ref, not its peeled target. allowReachableSHA1InWant peels annotated tags with peel_sha before walking the commit graph, allowing the tagged commit and its ancestors. Nested annotated tags are covered by the regression test.

dulwich/server.py
bojanz: No, the fetch path deduplicates all wants and validates them in one _reachable_wants call. It performs a single traversal from all ref tips and stops once every requested target is found (or the graph is exhausted). The regression test verifies reachability is checked only once per request

dulwich/server.py
bojanz: It is enforced during fetch: determine_wants() batches the requested objects and calls allowed_wants() before pack generation. The policy is loaded earlier because it also determines which capabilities the server advertises; caching it ensures advertisement and enforcement use the same policy.

bojanz:
@jelmer Thank you for the speedy review!

bojanz:
@jelmer Are you still interested in this change?

(And apologies if the ping feels rude, just not sure if I should continue pursuing this PR)

jelmer:
> @jelmer Are you still interested in this change?
> 
> (And apologies if the ping feels rude, just not sure if I should continue pursuing this PR)

No worries. I am still interested in this change, but it's in a core part of the code so I want to make sure I take enough time to properly understand the changes and their implications.

dulwich/server.py
```diff
@@ -505,6 +524,141 @@ def capabilities(self) -> list[bytes]:
...
+        return caps
+
+    def _get_sha1_in_want_policy(self) -> _SHA1InWantPolicy:
+        """Read the uploadpack.allow*SHA1InWant settings for this repository.
+
+        Returns: The configured policy. ``allowAnySHA1InWant`` implies the
+            other two options, as in git.
+        """
+        if self._sha1_in_want_policy is None:
+            if not isinstance(self.repo, BaseRepo):
```
jelmer: Please don't use isinstance(), this is a bad smell - BackendRepo shouldn't support config, but we also shouldn't be special casing here. Perhaps add a dedicated function for getting the SHA1InWantPolicy?

bojanz: Is https://github.com/jelmer/dulwich/pull/2334/changes/bbeb4d432e470ff12d91d46ef313a88a6b5f28c4 what you had in mind?

dulwich/server.py
```diff
@@ -505,6 +521,100 @@ def capabilities(self) -> list[bytes]:
...
+        if not (policy.allow_reachable or policy.allow_tip):
+            return set()
+
+        object_store = self.repo.object_store
+        if policy.allow_any:
+            return {sha for sha in wants if sha in object_store}
+
+        # Both of the remaining options accept the tip of any ref, including
+        # refs that were never advertised to this client.
+        tips = set(self.repo.get_refs().values())
```
jelmer: Please add a # TODO here about supporting uploadpack.hideRefs

PR #2335, opened 2026-08-01
Remove a packed ref even when packed-refs has not been read yet

bojanz:
DiskRefsContainer treated an uninitialized packed-refs cache as proof that a ref was not packed. As a result, deleting a packed ref from a newly opened repository reported success but left the ref in packed-refs.
    
Load packed refs before deciding whether the ref needs to be removed. This also avoids locking packed-refs when deleting a loose-only ref.


bojanz:
Replaced the previous AI-generated summary because it was too dense.

PR #2338, opened 2026-08-03
Retry loose object write when the fanout directory is pruned

### LINKED ISSUE #2336: DiskObjectStore.add_object() can race with "git gc"
`DiskObjectStore.add_object()` creates an object's fanout directory before it
opens the object lock file. A concurrent `git gc` or `git prune` can remove the
directory while it is still empty, causing the write to fail with
`FileNotFoundError`.

The race is between these two operations:

```python
os.mkdir(dir)
# git prune removes the still-empty directory here
with GitFile(path, "wb", ...) as f:
    ...
```

Git uses `rmdir()` for empty fanout directories, so once `GitFile` creates the
lock file the directory is safe. We only need to handle the small window before
that happens.

Here is a deterministic reproduction that removes the directory just before
`GitFile` opens it:

```python
import os
import shutil
import tempfile

import dulwich.object_store as object_store
from dulwich.object_store import DiskObjectStore
from dulwich.objects import Blob

tmp = tempfile.mkdtemp()
store = DiskObjectStore.init(os.path.join(tmp, "objects"))
blob = Blob.from_string(b"hello race")

real_gitfile = object_store.GitFile

def racing_gitfile(path, *args, **kwargs):
    shutil.rmtree(os.path.dirname(path))
    return real_gitfile(path, *args, **kwargs)

object_store.GitFile = racing_gitfile
store.add_object(blob)
```

This fails while `_GitFile` tries to create `<object>.lock`.

We carry a production fix that recreates the directory and retries the write.
For upstream, I suggest a small bounded retry in `add_object()` so a genuinely
missing object store still surfaces as an error instead of looping forever.

Would that approach, with a monkeypatched regression test like the reproduction
above, be acceptable?

Sanjays2402:
Fixes #2336

`DiskObjectStore.add_object()` created the fanout directory and then opened the object lock file; a concurrent `git gc`/`git prune` `rmdir()`s still-empty fanout directories, so the directory could disappear in that window and the write failed with `FileNotFoundError`. The write is now wrapped in a bounded retry (3 attempts) that recreates the directory only when it is actually gone — an existing directory re-raises immediately, so a genuinely broken object store still errors instead of looping.

Regression test `test_add_object_retries_when_fanout_dir_is_pruned` patches `dulwich.object_store.GitFile` to remove the directory on the first call; it fails on main with the reported `FileNotFoundError` and passes with the fix. `tests/test_object_store.py` is green (153 passed).

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -55,6 +55,7 @@ __all__ = [
 ]
 
 import binascii
+import errno
 import logging
 import os
 import stat
@@ -223,6 +224,10 @@ PACKDIR = "pack"
 # TODO: should packs also be non-writable on Windows? if so, that
 # would requite some rather significant adjustments to the test suite
 PACK_MODE = 0o444 if sys.platform != "win32" else 0o644
+
+# Number of times add_object() retries a loose object write when a concurrent
+# "git gc"/"git prune" removes the (still empty) fanout directory.
+_ADD_OBJECT_MAX_ATTEMPTS = 3
 
 # Grace period for cleaning up temporary pack files (in seconds)
 # Matches git's default of 2 weeks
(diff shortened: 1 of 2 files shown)


jelmer (changes requested):
This seems overly complicated. What does C Git do here?

jelmer:
See conversation in #2336 - I think the proper fix here is to address #2340 instead.

Issue #2054, opened 2026-01-03
Failure to process commit graph (octopus merges?)


dnelson-1901:
It looks like there's an issue with the commitgraph parsing code in `_parse_extra_edges`, where it runs off the end of the edge_data array in some cases.  I tried creating some small repos with fancy merges, but `git commit-graph write` doesn't want to create a commit-graph file for them, so I don't have a true minimal reproducer.  I can reproduce it with a clone of a small branch of the FreeBSD source tree, though, which has two 3-way merges in it.  It takes a minute or so to clone.

Here's the script I used to reproduce the issue, and sample output from an AWS cloudshell instance:

```sh
#! /bin/sh

E=dulwich-venv
rm -rf $E

python3 -m venv --upgrade-deps $E

cd $E
. ./bin/activate 

pip install dulwich 

git clone --branch stable/3 --single-branch --no-checkout https://github.com/freebsd/freebsd-src.git
cd freebsd-src
git commit-graph write
dulwich log > /dev/null
python3 --version
git --version
pip list | grep dulwich
```

<details><summary>Details<pre>
  File "/home/cloudshell-user/dulwich-venv/lib64/python3.9/site-packages/dulwich/commit_graph.py", line 258, in _parse_extra_edges
    parent_pos = struct.unpack(">L", edge_data[offset : offset + 4])[0]
struct.error: unpack requires a buffer of 4 bytes
</pre></summary>
<p>
<pre>
/home/cloudshell-user> ./dulwich-venv.sh   
Requirement already satisfied: pip in ./dulwich-venv/lib/python3.9/site-packages (21.3.1)
Collecting pip
  Using cached pip-25.3-py3-none-any.whl (1.8 MB)
Requirement already satisfied: setuptools in ./dulwich-venv/lib/python3.9/site-packages (59.6.0)
Collecting setuptools
  Using cached setuptools-80.9.0-py3-none-any.whl (1.2 MB)
Installing collected packages: setuptools, pip
  Attempting uninstall: setuptools
    Found existing installation: setuptools 59.6.0
    Uninstalling setuptools-59.6.0:
      Successfully uninstalled setuptools-59.6.0
  Attempting uninstall: pip
    Found existing installation: pip 21.3.1
    Uninstalling pip-21.3.1:
      Successfully uninstalled pip-21.3.1
Successfully installed pip-25.3 setuptools-80.9.0
Collecting dulwich
[...]

Issue #1015, opened 2022-09-01
flaky network tests on Windows


jelmer:
Some of the network tests appear to be flaky on windows, and fairly regularly fail:

```
 ======================================================================
ERROR: test_fetch_pack (dulwich.tests.compat.test_client.DulwichTCPClientTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\a\dulwich\dulwich\dulwich\protocol.py", line 230, in read_pkt_line
    sizestr = read(4)
  File "C:\hostedtoolcache\windows\Python\3.10.6\x64\lib\socket.py", line 705, in readinto
    return self._sock.recv_into(b)
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "D:\a\dulwich\dulwich\dulwich\tests\compat\test_client.py", line 238, in test_fetch_pack
    result = c.fetch(self._build_path("/server_new.export"), dest)
  File "D:\a\dulwich\dulwich\dulwich\client.py", line 837, in fetch
    result = self.fetch_pack(
  File "D:\a\dulwich\dulwich\dulwich\client.py", line 1183, in fetch_pack
    _handle_upload_pack_tail(
  File "D:\a\dulwich\dulwich\dulwich\client.py", line 629, in _handle_upload_pack_tail
    _read_side_band64k_data(
  File "D:\a\dulwich\dulwich\dulwich\client.py", line 509, in _read_side_band64k_data
    for pkt in pkt_seq:
  File "D:\a\dulwich\dulwich\dulwich\protocol.py", line 290, in read_pkt_seq
    pkt = self.read_pkt_line()
  File "D:\a\dulwich\dulwich\dulwich\protocol.py", line 242, in read_pkt_line
    raise GitProtocolError(e)
dulwich.errors.GitProtocolError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

jelmer:
From what I can tell, this is specific to the TCP Client and it might not be isolated to Windows (though more likely to be triggered there).

jelmer:
I believe this is fixed - we haven't actually seen this error in a while. It was probably related to the deepen fix.

Issue #584, opened 2017-11-16
Unable to delete cloned repo from within same Python process


MattSable:
If I clone a repo and then try to delete it, I get:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Temp\\tmp2mjvpymq\\.git\\objects\\pack\\pack-acc7c293230a64c24a6c46faa4fdf8f2f3fe9a12.idx`

This can be seen with the following code:

```
from dulwich import porcelain
import shutil

repo_path = "./to_delete"
porcelain.clone("https://github.com/jelmer/dulwich.git", repo_path)
shutil.rmtree(repo_path)

```

jelmer:
On Thu, Nov 16, 2017 at 10:35:45PM +0000, Matt wrote:
> If I clone a repo and then try to delete it, I get:
> 
> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Temp\\tmp2mjvpymq\\.git\\objects\\pack\\pack-acc7c293230a64c24a6c46faa4fdf8f2f3fe9a12.idx'
Can you provide some code that reproduces this?

-- 
Jelmer Vernooĳ <jelmer@jelmer.uk>
PGP Key: https://www.jelmer.uk/D729A457.asc

MattSable:
Updated the OP. My actual goal is to just get the files from a repo, without any association to git. I'm trying to do this by cloning the repo and removing the .git folder. Is there any other way to do this?

jelmer:
Ah, thanks. Can you try calling .close() on the object returned by porcelain.clone() before calling rmtree() ?

MattSable:
That works. Thanks!

moresayali246:
> Ah, thanks. Can you try calling .close() on the object returned by porcelain.clone() before calling rmtree() ?

can u please elaborate

PR #2339, opened 2026-08-03
fix commit-graph extra edge indexing in _parse_extra_edges

netliomax25-code:
1. For an octopus merge the commit data chunk stores an "array position within this list" into the Extra Edge List (gitformat-commit-graph), and that list holds 4-byte values, but `_parse_extra_edges` used the stored value directly as a byte offset into the EDGE chunk.
2. Position 0 is the only value where a position and a byte offset agree, so the first octopus merge in a graph reads correctly and every later one reads from a misaligned offset that splices halves of two neighboring entries together. Those values land past the end of the commit list, where the existing `parent_pos < len(oids)` check drops them without raising, so `get_parents()` quietly returns a short parent list.
3. `ParentsProvider.get_parents` consults the commit graph before falling back to the object store, so the wrong ancestry reaches `find_merge_base`, `can_fast_forward` and the shallow-depth walk.

Scaled the stored position by 4 at the single site that reads it. Dulwich's own writer never emits an EDGE chunk, so only graphs written by git are affected.

Checked against git 2.52 on a repository with seven octopus merges of 3 to 6 parents: six of the seven disagreed with `git rev-list --parents` before the change, and all 41 commits agree after. A sha256 repository was checked the same way. The tolerance for a truncated EDGE chunk added for #2054 is unchanged, and its test still passes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,13 @@
 1.2.13	UNRELEASED
 
+ * Fix reading the commit-graph extra edge list, which stores a position in a
+   list of 4-byte values rather than a byte offset. Only the first octopus
+   merge in a graph sits at position 0, where the two coincide; every later
+   one was read from a misaligned offset, so ``CommitGraph.get_parents()``
+   dropped or invented parents. This fed ``ParentsProvider`` and therefore
+   merge-base, fast-forward and shallow-depth computations.
+   (netliomax25-code)
+
  * HARDEN: Validate shallow object ids received from the remote.
    (netliomax25-code)
 
--- a/NEWS
+++ b/NEWS
@@ -1,12 +1,8 @@
 1.2.13	UNRELEASED
 
- * Fix reading the commit-graph extra edge list, which stores a position in a
-   list of 4-byte values rather than a byte offset. Only the first octopus
-   merge in a graph sits at position 0, where the two coincide; every later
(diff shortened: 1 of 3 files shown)


netliomax25-code:
The one red check is `test-minimal (windows-latest, 3.13)`, which failed in `tests.compat.test_client.DulwichTCPClientTest.test_archive` with `PermissionError: [WinError 32] The process cannot access the file because it is being used by another process` while `shutil.rmtree` cleaned up its temp directory.

The same commit passes that job on Windows 3.10 and on the other five `test-minimal` cells, and all fifteen `test` cells are green. That test does not read a commit graph. This looks like the Windows temp-cleanup race from #584 / #1015 rather than anything in this change, though I can't re-run the job from a fork to confirm.

Locally the full suite is green in the same pure-Python configuration that job uses: `Ran 4516 tests ... OK (skipped=247, expected failures=1)`.

NEWS
```diff
@@ -1,5 +1,13 @@
 1.2.13	UNRELEASED
 
+ * Fix reading the commit-graph extra edge list, which stores a position in a
```
jelmer: please keep NEWS messages shorter than this, ideally just 3-4 lines.

netliomax25-code: Done, trimmed it to three lines.

PR #2341, opened 2026-08-03
Freshen existing loose objects in add_object (#2340)

### LINKED ISSUE #2340: DiskObjectStore.add_object() skips the write for an existing object without refreshing its mtime
Found in #2336 while comparing the logic with C git.

Right now add_object() returns early when the loose file is already present (dulwich/object_store.py:2370):
```
if os.path.exists(path):
    return  # Already there, no need to write again
```
C git does not do a plain existence check here. write_loose_object() calls odb_freshen_object(), which reaches check_and_freshen_file() -> freshen_file() -> utime(). If the utime() fails, git falls through and writes the object anyway.

This is mitigation documented in git's own git-gc NOTES (Documentation/git-gc.adoc:163): "Most operations that add an object to the database update the modification time of the object if it is already present" so that the --prune expiry cutoff protects it.

Consequence for dulwich: if the object is unreachable and older than the prune cutoff, add_object() reports success while leaving it a prune candidate. A concurrent git gc unlinks it (prune_object() skips only st.st_mtime > expire), and the caller then commits a ref pointing at a missing object. Unlike #2336 this doesn't raise.

The window isn't narrow, either. With the default two-week gc.pruneExpire, any loose object that has been sitting around unreferenced qualifies - a fairly normal state for objects written by an earlier aborted operation and re-added later.

Suggested fix, mirroring git: replace the os.path.exists() check with an os.utime(path, None), returning early only if it succeeds and falling through to the normal write on OSError.

bojanz:
Closes #2340

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -5,11 +5,8 @@
    the first got a wrong parent list. (netliomax25-code)
 
  * Refresh the mtime of an existing loose object in
-   ``DiskObjectStore.add_object()`` instead of only checking for its
-   existence. A stale mtime leaves the object a candidate for age-based
-   pruning, so a concurrent ``git gc`` could remove an object that
-   ``add_object()`` had reported as present and that the caller was about to
-   reference. Matches git's ``freshen_loose_object()``.
+   ``DiskObjectStore.add_object()``, so that a concurrent ``git gc`` cannot
+   prune an object that was just reported as present.
    (Bojan Zivanovic, #2340)
 
  * HARDEN: Validate shallow object ids received from the remote.
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,14 @@
    list of 4-byte entries, not a byte offset, so every octopus merge after
    the first got a wrong parent list. (netliomax25-code)
 
+ * Refresh the mtime of an existing loose object in
+   ``DiskObjectStore.add_object()`` instead of only checking for its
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -1,5 +1,13 @@
 1.2.13	UNRELEASED
 
+ * Refresh the mtime of an existing loose object in
```
jelmer, line 7: Please keep news items shorter, at most 3-4 lines.

dulwich/object_store.py
```diff
@@ -2367,8 +2367,18 @@ def add_object(self, obj: ShaFile) -> None:
...
-            return  # Already there, no need to write again
+        try:
+            # Refresh the mtime instead of just checking for existence. A
+            # loose object with a stale mtime is a candidate for age-based
+            # pruning, so a concurrent "git gc" could remove it before the
+            # caller has created a reference to it. If the utime() fails the
+            # object is missing or cannot be freshened, in which case it is
+            # not safe to skip the write.
+            os.utime(path, None)
+        except OSError:
```
jelmer: Please don't catch blanket exceptions, but specific ones.

jelmer (changes requested):

bojanz:
@jelmer Thanks for the review, pushed a followup.

jelmer (approved):

bojanz:
Rebased

PR #2348, opened 2026-08-10
Keep the directory constraint on patterns ending in "**/"

adarshsm:
A trailing `**/` makes a gitignore pattern directory-only, but the translation dropped the
trailing slash:

```python
>>> from dulwich.ignore import translate
>>> translate(b"foo/**/")
b'(?ms)foo/.*\\Z'      # before
b'(?ms)foo/.*/\\Z'     # after
```

Without the slash the pattern also matches `foo/` itself and every file sitting directly in
`foo`, while Git only ignores the directories below `foo`.

Confirmed against Git in a real working tree, which is the behaviour that matters here:

```console
$ cat .gitignore
foo/**/
$ git status --porcelain -uall --ignored
?? foo/f.txt          # not ignored
!! foo/sub/g.txt      # ignored
```

`IgnoreFilterManager.is_ignored("foo/f.txt")` returned `True` before this change, so
`foo/f.txt` was hidden from anything walking a tree through dulwich.

Since `dulwich.attrs` shares the same `wildmatch` translation, gitattributes patterns
ending in `**/` were affected in the same way.

## A note on how this was verified

`git check-ignore` turned out to be an unreliable oracle for this, twice over, so I want to
flag it rather than have a reviewer trip on the same thing:

- With a path already in the index it reports "not ignored" regardless of the patterns,
  unless `--no-index` is passed.
- Even with `--no-index`, `git check-ignore foo/` reports `foo/` as ignored under
  `foo/**/`, which contradicts the working-tree behaviour above — if `foo/` really were
  ignored, `foo/f.txt` could not show up as untracked.

So the ground truth used here is `git status --porcelain -uall --ignored` on a real
repository. Comparing that against `IgnoreFilterManager` over a matrix of patterns
(`abc/**/`, `abc/**`, `abc/`, `**/tmp/`, `logs/**/`, `abc/**/def/`) and their contents:
**2 of 16 cases disagreed with Git before this change, 0 of 16 after.**

## Tests

Three regression tests, all of which fail without the change:

- `translate(b"foo/**/")` keeps the trailing slash (`TRANSLATE_TESTS`).
- `foo/bar/` and `foo/bar/bla/` match `foo/**/` (`POSITIVE_MATCH_TESTS`), while `foo/bla.c`
  and `foo/` do not (`NEGATIVE_MATCH_TESTS`).
- `IgnoreFilterManagerTests.test_trailing_double_asterisk_slash_is_directory_only` covers
  the end-to-end behaviour through a real repository.

```
pytest tests/test_ignore.py tests/test_attrs.py tests/test_wildmatch.py   # 118 passed
pytest tests/                                                            # 3732 passed
```

The full run also reports 12 failures in `tests/test_signature.py` and
`tests/test___main__.py`, all from `gpg`/`gpgsm` not being installed in my environment. The
failing test ids are identical with and without this change.

`ruff check` and `ruff format --check` are clean, and there is a NEWS entry.

## Out of scope

The same comparison surfaced a second, unrelated divergence: `abc/*` does not match `abc/`
in dulwich but does in Git. Different root cause, so I have left it out of this PR rather
than widen the change. Happy to look at it separately if you would like.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -40,6 +40,11 @@
    compat layer, so repositories carrying such paths could not be cloned
    elsewhere. (Jelmer Vernooĳ, #2351)
 
+ * Fix ``ignore`` and ``attrs`` matching for patterns ending in ``**/``. The
+   trailing slash was dropped when translating the pattern, so ``foo/**/`` also
+   matched ``foo/`` itself and every file directly inside it. Git only ignores
+   the directories below ``foo``. (adarshsm)
+
  * Catch ``PackFileDisappeared`` in the bitmap probe in
    ``get_reachability_provider``, which only guarded ``FileNotFoundError``.
    (Jelmer Vernooĳ, #2344)
--- a/dulwich/wildmatch.py
+++ b/dulwich/wildmatch.py
@@ -242,6 +242,12 @@ def _translate_double_asterisk(segments: Sequence[bytes], i: int) -> tuple[bytes
     # Check if ** is at end
     remaining = segments[i + 1 :]
     if all(s == b"" for s in remaining):
+        if remaining:
+            # Trailing "**/" is a directory pattern, so it has to consume at
+            # least one directory and end in a slash. Without this, "abc/**/"
+            # also matches "abc/" itself and every file directly inside it,
+            # while Git only ignores the directories below "abc".
+            return b".*/", False
         # ** at end - matches everything
         return b".*", False
 
--- a/tests/compat/test_check_attr.py
+++ b/tests/compat/test_check_attr.py
@@ -143,6 +143,45 @@ class CheckAttrCompatTestCase(CompatTestCase):
             self._create_file(path)
         self._assert_attr_match("text", paths)
 
+    def test_trailing_double_asterisk_slash(self) -> None:
+        # ``foo/**/`` in .gitattributes goes through the same wildmatch
+        # translation as .gitignore, so the trailing slash must be preserved:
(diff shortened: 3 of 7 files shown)


jelmer (dismissed):
Please also add these expressions to tests/compat/test_ignores.py

adarshsm:
Added them in 124769ef — `tests/compat/test_check_ignore.py` now checks the trailing `**/` matrix against `git check-ignore` (I assume that is the file you meant; there is no `tests/compat/test_ignores.py`). Reverting `wildmatch.py` makes two of the three fail, so they hold the fix rather than just describing it.

Two things I found while writing them, both outside this fix:

**`foo/` itself.** I left it out of the matrix because git answers it two different ways. `git status --ignored` leaves the directory untracked, but `git check-ignore foo/` calls it ignored, matched by `foo/**/`. That is the trailing slash of the argument meeting a `**` that will match nothing — `git status` never passes the slash, so it never sees the match. The fix follows `git status` here; the note is in the test docstring.

**Negation under a `**/` directory does not re-include.** With

```
foo/**/
!foo/keep/
```

`git status` leaves `foo/keep/bla.c` untracked, but `IgnoreFilterManager` reports it ignored. This predates the fix — it behaves the same with `wildmatch.py` reverted — and `Pattern.match` matching a path through its ancestor directories looks like the cause, so I did not chase it here. Happy to open a separate issue if it is not already known.

PR #2349, opened 2026-08-10
Keep borrowed packs open after cache eviction

bojanz:
Several code paths close a `Pack` at the moment it leaves the object-store cache: `_add_cached_pack` closes the pack it replaces, `_enforce_packed_git_limit` closes the least-recently-used pack, `_remove_pack` closes on removal, and `_update_pack_cache` closes packs whose files disappeared.

Closing at eviction time is unsafe when the store is shared: `packs` and `get_raw` hand out references to cached `Pack` objects, so another consumer — a concurrent request in a threaded/greenlet server, or an iterator like `iter_unpacked` that is still mid-pack — may be reading from the pack when eviction closes it under its feet. The reader then fails on a closed file even though the pack file itself is still perfectly valid on disk.

This changes eviction to release ownership instead of closing: the pack is dropped from the cache, and the underlying files are closed by `Pack.__del__` once the last borrower is done with it. Since that hand-off is intentional, `__del__` no longer emits the "unclosed Pack" `ResourceWarning` for packs released this way — explicit leaks still warn as before.

The new test exercises the LRU-limit path: a pack evicted by `packed_git_limit` while an `iter_unpacked` iterator is mid-stream stays readable until the iterator completes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,8 @@
 1.2.13	UNRELEASED
 
+ * Keep packs usable by existing readers after they are evicted from the
+   object-store cache. (Bojan Zivanovic)
+
  * SECURITY: Don't follow symlinks when writing patch files in
    ``porcelain.format_patch``. A symlink pre-planted at a patch's filename in
    the output directory was followed, writing the patch outside that directory.
--- a/dulwich/object_store.py
+++ b/dulwich/object_store.py
@@ -1065,7 +1065,7 @@ class PackBasedObjectStore(PackCapableObjectStore, PackedObjectContainer):
         if prev_pack is not pack:
             self._pack_cache[base_name] = pack
             if prev_pack:
-                prev_pack.close()
+                prev_pack._release_from_cache()
         self._mark_pack_used(base_name)
         self._enforce_packed_git_limit()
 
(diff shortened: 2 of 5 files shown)


bojanz:
The Windows failures were real: `test_packed_git_limit_evicts_lru_packs` started leaving an evicted pack's `.idx` open into teardown, so `rmtree` failed with WinError 32. Pushed a fix.

Unsure how useful `test_no_reference_cycle` truly is, considering that we already had a test that failed.

bojanz:
The last commit could be a followup PR as well if you prefer it that way.

One behavioral note: work and errors that previously happened at the call, most notably the full sort in `sorted_entries()` and the checksum/length validation on first `self.data` access in `iterobjects()`, now happen at the first `next()`.

bojanz:
Drafted cause I want to take another look at this once we land https://github.com/jelmer/dulwich/pull/2334

bojanz:
The fix here wasn't enough, we had to expand it internally and it now looks very different. 

Let's start from scratch. I opened https://github.com/jelmer/dulwich/issues/2424 to discuss the problem, and I'm benchmarking the new fix, which I will open as a new PR.

PR #2352, opened 2026-08-11
avoid exponential commit walk in find_shallow and get_depth

netliomax25-code:
`find_shallow` and `get_depth` in `dulwich.object_store` walk the commit DAG from a set of heads but never deduplicate the frontier, so a commit reachable along k distinct paths is expanded k times and a small merge-heavy history is walked in exponential time.

1. In `find_shallow` the `todo` stack pushes every parent once per visiting path. A diamond history of ~3n commits (each merge reachable through two sub-paths) exposes 2**n paths, and `find_shallow` runs pre-auth on `git-upload-pack`: a client `deepen` request reaches it through `ProtocolGraphWalker._handle_shallow_request` before any authentication.
2. `get_depth` has the same missing visited set, plus an O(n) `list.pop(0)` on each step.
3. Both now skip already-seen `(sha, depth)` states. Re-processing a state only re-adds to the same set and re-queues the same parents, so the returned sets/value are unchanged; `get_depth` also switches to a `deque`. This mirrors git, which marks commits SEEN and records the depth already assigned so each is visited once.

Measured on a diamond history, `find_shallow` at a depth beyond the longest path:

    commits    before        after
    31         0.007 s       ~0.0002 s
    37         0.040 s       ~0.0002 s
    43         0.143 s       ~0.0003 s
    49         0.832 s        0.0003 s
    73         >5,000,000 frontier pops (aborted)   0.0009 s

An anonymous `git clone --depth=N` of a pushed ~100-commit repository pins a server CPU for a long time per request before any pack is generated. Added a regression test and a NEWS entry.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,14 @@
 1.2.13	UNRELEASED
 
+ * SECURITY: Deduplicate the commit walk in ``find_shallow`` and ``get_depth``
+   (``dulwich.object_store``). Both re-expanded a commit once per path that
+   reached it, so a small merge-heavy history walked in exponential time. A
+   ``deepen`` request on ``git-upload-pack`` reaches ``find_shallow`` before any
+   authentication, so a pushed ~100-commit repository could pin a server CPU for
+   hours per shallow clone. The traversal now skips already-visited
+   ``(sha, depth)`` states, which leaves the result unchanged.
+   (netliomax25-code)
+
  * SECURITY: Don't follow symlinks when writing patch files in
    ``porcelain.format_patch``. A symlink pre-planted at a patch's filename in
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -1,5 +1,14 @@
 1.2.13	UNRELEASED
 
+ * SECURITY: Deduplicate the commit walk in ``find_shallow`` and ``get_depth``
```
jelmer, line 3: This is not a security issue; you can push all sorts of repositories that use resources on the server, it's up to the user to restrict resource consumption.

Also, please restrict the number of lines in NEWS entries.

jelmer:
obsoleted by #2358

Issue #2350, opened 2026-08-11
Make PackData thread-safe


bojanz:
PackData reads from a single shared file object using seek() followed by read():

```python
  def get_unpacked_object_at(self, offset, *, include_comp=False):
      assert offset >= self._header_size
      self._file.seek(offset)
      unpacked, _ = unpack_object(
          self._file.read, self.object_format.hash_func, include_comp=include_comp
      )
```
Because the file position is shared, two readers can interfere with each other. One reader may seek to an offset, another may change the position, and the first will then read the wrong bytes. Since unpack_object() usually performs several reads, there are multiple opportunities for this to happen. The resulting failures are intermittent and misleading: zlib errors, invalid object types, checksum mismatches, and similar errors even though the pack itself is valid.

This can happen fairly easily because PackBasedObjectStore caches Pack objects and shares them between callers of get_raw() and .packs. A server handling concurrent requests for the same repository can therefore end up with multiple threads or greenlets reading through the same PackData instance.

There is also a single-threaded version of the problem. iter_unpacked() stores its current position in the shared file offset, including a relative seek() between yields. If code alternates between an active iter_unpacked() iterator and a random-access operation such as get_object_at() or get_stored_checksum(), the iterator’s position can be changed underneath it.

At Platform.sh/Upsun, we have carried a patch for several years that protects every seek/read sequence with an RLock. That fixes the corruption, but it also serializes all reads from a pack. We would like to upstream a better solution.

## Proposal

On platforms that support it, use os.pread(fd, count, offset). Unlike seek() followed by read(), pread() takes the offset as an argument and does not change the file position. Readers can therefore operate independently without locking.

A small cursor wrapper would let existing code continue working with a normal read() callback:
```python
  class _PreadCursor:
      """A per-consumer cursor backed by pread()."""

      def __init__(self, fd: int, offset: int) -> None:
          self._fd = fd
          self.offset = offset

      def read(self, size: int) -> bytes:
          data = os.pread(self._fd, size, self.offset)
          self.offset += len(data)
          return data
```
The main changes would be:

- get_unpacked_object_at() creates a cursor at the requested offset and passes its read method to unpack_object().
- Each iter_unpacked() iterator gets its own cursor. Its current position is then private to that iterator, and the existing SEEK_CUR backtracking becomes simple cursor arithmetic.

- get_stored_checksum() uses one pread() call at len(self) - checksum_size.
- calculate_checksum() and check() could also use cursors, either in the same change or later.

[...]

PR #2356, opened 2026-08-14
Make concurrent PackData reads thread-safe

bojanz:
Use os.pread where available so concurrent pack readers no longer share a file position. Provide synchronized fallback reads for supplied file objects and systems without pread, coordinate reads with close operations, and protect the pack offset cache.

Add targeted coverage for concurrent reads, cursor behavior, checksum reads, and close races.

Issue: https://github.com/jelmer/dulwich/issues/2350

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,9 @@
 1.2.13	UNRELEASED
 
+ * Read packs via ``os.pread`` where available instead of seek+read on a
+   shared file position, so concurrent readers of the same ``PackData`` no
+   longer corrupt each other's reads. (Bojan Zivanovic, #2350)
+
  * SECURITY: Don't follow symlinks when writing messages in
    ``porcelain.format_patch``, ``mbox.split_mbox`` and ``mbox.split_maildir``.
    A symlink pre-planted at an output filename was followed, writing the
--- a/dulwich/pack.py
+++ b/dulwich/pack.py
@@ -1828,12 +1828,18 @@ class _PreadCursor:
     # ``unpack_object`` reads variable-length headers one byte at a time.
     # Reading a block in advance avoids a separate system call for each byte.
     _READAHEAD = 512
+    # Start small because most cursors only read one object. If a cursor keeps
+    # missing its buffer, grow the window so sequential scans need fewer reads.
+    _MAX_READAHEAD = 256 * 1024
+
+    __slots__ = ("_buf", "_buf_start", "_pack_data", "_readahead", "offset")
 
     def __init__(self, pack_data: "PackData", offset: int) -> None:
         self._pack_data = pack_data
         self.offset = offset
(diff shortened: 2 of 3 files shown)


bojanz:
@jelmer The pack _offset_cache is still unsynchronized. I couldn't get that fix to pass tests yet, so I've excluded it from the PR. I can add it back if you think that the PR shouldn't land without that.

jelmer (changes requested):
The whole point of pread is to avoid cursors, so adding a cursor wrapper for pread seems counterintuitive. This also adds a bunch of overhead in a performance-critical part of the code as well as a lot of complexity. In some quick and dirty benchmarking, I see a 10-20% performance degradation from this PR.

Maybe we can reevaluate using mmap for packs instead?

bojanz:
I am not thrilled by the amount of complexity either. FWIW the current work is about 20 commits squashed together, the first one was nice and simple, but getting the semantics right (e.g. close() races) took both code and effort.

That said, having a cursor abstraction on top of pread feels unavoidable, unpack_object() expects a stateful read(size) callback. Though I might be running into my limits of understanding the current architecture.

A 10-20% performance degradation is completely unacceptable, but I'd like to see if there's anything I can do to optimize it before jumping into an alternative approach with mmap, so that at least we have a fair comparison.

Can you share your quick&dirty benchmark?

bojanz:
I've pushed an optimization. Will focus on mmap in the next two days, then we can compare.

jelmer:
> I am not thrilled by the amount of complexity either. FWIW the current work is about 20 commits squashed together, the first one was nice and simple, but getting the semantics right (e.g. close() races) took both code and effort.
> 
> That said, having a cursor abstraction on top of pread feels unavoidable, unpack_object() expects a stateful read(size) callback. Though I might be running into my limits of understanding the current architecture.

If we're going to use a cursor, then we might as well just use file descriptors - perhaps just one per thread? That said, I think mmap /or/ direct use of os.pread (rather than a cursor!) are going to keep the code much simpler. 

> 
> A 10-20% performance degradation is completely unacceptable, but I'd like to see if there's anything I can do to optimize it before jumping into an alternative approach with mmap, so that at least we have a fair comparison.
> 
> Can you share your quick&dirty benchmark?

It was a quick one-off that I didn't save unfortunately. We do also have https://github.com/dulwich/benchmarks but I haven't run that in a while.

bojanz:
Dropped the cursors and it's definitely simpler while being somewhat faster. Still, can't call it simple.

The mmap approach is looking more promising, but I need to give it another pass in the morning before opening an PR.

bojanz:
The mmap PR is up at https://github.com/jelmer/dulwich/pull/2362 and it is significantly simpler. Closing this PR in favor of it.

PR #2360, opened 2026-08-17
Add sanitize_user_identity, mirroring git's fmt_ident sanitization

### LINKED ISSUE #2342: Add a helper to sanitize user identities the way git does
Today, check_user_identity() validates an identity and raises InvalidUserIdentity.

Git does not reject: it sanitizes. strbuf_addstr_without_crud() in ident.c strips leading and trailing "crud" (c <= 32, plus , : ; < > " \ ') and drops \n, < and > from the middle, and fmt_ident() applies it to the name and the email separately.

That difference matters when the identity is not under your control - a hosted git service, an importer, anything building a commit from a user-supplied name and email. Rejecting is not an option, so today each caller either writes the raw value and produces a malformed identity line, or reimplements git's rules.

We maintain a port of strbuf_addstr_without_crud for exactly this reason and would be glad to upstream it. There is already precedent for mirroring git's sanitizers: patch.py:_sanitize_subject_for_filename documents itself as matching git's format_sanitized_subject.

Before writing a PR:

1. Would you take a sanitizer at all, or is validate-and-raise the intended contract?
2. If yes, does repo.py next to check_user_identity() seem like the right home?
3. Should it be a standalone helper only, or should the commit paths that currently call check_user_identity() use it?

(I am not convinced this one is a must have, but it's on my diff list so I figured it was worth offering).

bojanz:
check_user_identity validates and raises, which is no help when the identity is not under the caller's control (a hosted git service, an importer). Git does not reject in fmt_ident, it strips leading and trailing crud from the name and the email separately and drops the line delimiters from the middle, via strbuf_addstr_without_crud. Ported that as a standalone helper next to check_user_identity.

Fixes #2342

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,10 @@
 1.2.13	UNRELEASED
 
+ * Add ``repo.sanitize_user_identity``, which builds an identity from a
+   name and an email sanitized the way git's ``fmt_ident`` does, for
+   callers who cannot reject invalid input via ``check_user_identity``.
+   (Bojan Zivanovic, #2342)
+
  * Deduplicate the commit walk in ``find_shallow`` and ``get_depth``
    (``dulwich.object_store``). Both re-expanded a commit once per path that
    reached it, so a merge-heavy history walked in exponential time.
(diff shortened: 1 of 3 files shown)


bojanz:
@jelmer I have trouble getting the Windows 3.11 test to pass here, looks unrelated to my changes.

jelmer (approved):

PR #2362, opened 2026-08-18
Make concurrent PackData reads thread-safe via mmap

bojanz:
This is an alternative to https://github.com/jelmer/dulwich/pull/2356 which used os.pread but ended up being too complex.


bojanz:
Replaced by https://github.com/jelmer/dulwich/pull/2365

PR #2365, opened 2026-08-19
Make concurrent ``PackData`` reads thread-safe by mmap-ing pack contents

jelmer:
Make concurrent ``PackData`` reads thread-safe by mmap-ing pack contents
and indexing the mapping at explicit offsets instead of sharing the file
position. New ``unpack_object_at``, ``read_pack_header_at``,
``read_zlib_chunks_at``, ``take_msb_bytes_at`` and ``compute_buffer_sha``
read from a buffer; the existing read-callable variants remain for streams.

based on #2362

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,8 +10,11 @@
    (Bojan Zivanovic, #2342)
 
  * Make concurrent ``PackData`` reads thread-safe by mmap-ing pack contents
-   and giving each operation an independent cursor instead of sharing the
-   file position. (Bojan Zivanovic)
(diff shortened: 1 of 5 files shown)


bojanz:
@jelmer Thank you for rewriting this until it started to make sense. Greatly appreciated.

jelmer:
> @jelmer Thank you for rewriting this until it started to make sense. Greatly appreciated.

Thank you for driving this! Very pleased to see that we now ended up with both thread safety and improved performance :-)

PR #2366, opened 2026-08-20
Make Pack.get_raw thread-safe

bojanz:
Protect PackData's offset cache with a lock. LRUSizeCache updates its linked list on reads, so concurrent cache accesses could corrupt it while resolving deltas.

The next (last?) part of https://github.com/jelmer/dulwich/issues/2350, now that https://github.com/jelmer/dulwich/pull/2365 is in.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,9 @@
 1.2.13	UNRELEASED
 
+ * Make concurrent ``Pack.get_raw`` calls thread-safe by synchronizing
+   ``PackData``'s resolved-object offset cache.
+   (Bojan Zivanovic)
+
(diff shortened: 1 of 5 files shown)


bojanz:
In a quick microbenchmark  PackData.get_object_at() on a cache hit goes from ~150ns to ~450ns. 

We could shave about 150ns by making _get_cached_object_at() uglier:
```python
    def _get_cached_object_at(
        self, offset: int
    ) -> tuple[int, OldUnpackedObject] | None:
        self._offset_cache_lock.acquire()
        try:
            return self._offset_cache[offset]
        except KeyError:
            return None
        finally:
            self._offset_cache_lock.release()
```

But I am not sure how much it matters overall. I'll let you make that call.

PR #2367, opened 2026-08-23
Add support for 'git hook run' (#1842)

### LINKED ISSUE #1842: Add support for git hook
Add support for the `git hook` command to run Git hooks.

This command provides a way to manually execute Git hooks or manage hook execution.

**Common use cases:**
- Manually trigger Git hooks for testing
- Run hooks outside of normal Git operations
- Debug and validate hook behavior

UsamaQaisrani:
Adds a `dulwich hook run <name>` CLI command and `porcelain.hook_run()`
to manually execute git hooks for testing/debugging, per #1842.

Supports: pre-commit, post-commit, commit-msg, update.
Not yet supported: pre-receive, post-receive (they operate on a batch
of ref updates rather than individual CLI arguments, noted as a
known limitation).

Closes #1842


dulwich/porcelain/__init__.py
```diff
@@ -10119,6 +10120,29 @@ def am_quit(repo: RepoPath | None = None) -> None:
...
+
+    Returns:
+        Whatever the underlying hook returns; varies by hook type
+    """
+    with open_repo_closing(repo) as r:
+        controldir = r.controldir()
+        cwd = r.path
+        hook = get_hook(hook_name, cwd, controldir)
+        encoded_args = [a.encode() for a in args]
+        return cast(bytes | tuple[bytes, bytes] | None, hook.execute(*encoded_args))
```
jelmer, line 10143: why does this cast? How do we know the hook returns one of those types?

dulwich/hooks.py
```diff
@@ -364,3 +364,53 @@ def execute(
...
+            post-receive)
+    """
+    hook_class = HOOK_CLASSES.get(hook_name)
+    if not hook_class:
+        raise HookError(
+            f"unrecognized hook {hook_name!r}, expected one of: "
+            f"{', '.join(sorted(HOOK_CLASSES))}"
+        )
+
+    if hook_class is PreReceiveShellHook or hook_class is PostReceiveShellHook:
```
jelmer, line 407: does C Git support these?

dulwich/porcelain/__init__.py
UsamaQaisrani: I manually checked each hook's return type. I can swap it for a runtime assert isinstance() so it fails instead of just asserting it?

dulwich/hooks.py
UsamaQaisrani: C Git doesn't restrict it. Didn't implement here because PreReceiveShellHook/PostReceiveShellHook expect a structured ref-tuple list, not flat CLI args. Kept this PR to the flat-arg hooks for now, happy to follow up with stdin-based parsing to support these properly, or do it here if you'd prefer.

tests/cli/test_cli.py
```diff
@@ -2714,6 +2714,45 @@ def test_remote_add(self):
...
+                '#!/bin/sh\necho "stdout: $1 $2 $3"\necho "stderr output" >&2\nexit 0\n'
+            )
+        os.chmod(test_file, 0o755)
+
+        result, stdout_text, stderr_text = self._run_cli(
+            "hook", "run", "update", "refs/heads/main", "abc123", "def456"
+        )
+        self.assertIsNone(result)
+        self.assertIn("stdout: refs/heads/main abc123 def456", stdout_text)
+        self.assertIn("stderr output", stderr_text)
```
jelmer, line 2753: Please also add tests for e.g. commit-msg hook

dulwich/cli.py
```diff
@@ -7714,6 +7715,43 @@ def run(self, args: Sequence[str]) -> None:
...
+
+        Args:
+            args: Command line arguments
+        """
+        parser = argparse.ArgumentParser()
+        parser.add_argument("hook_name", help="Name of the hook to run")
+        parser.add_argument("args", nargs="*", help="Arguments to pass to the hook")
+        parsed_args = parser.parse_args(args)
+        try:
+            output = porcelain.hook_run(".", parsed_args.hook_name, parsed_args.args)
```
jelmer, line 7732: please don't hardcode "." but use None, we want to support GIT_DIR here
```suggestion
            output = porcelain.hook_run(None, parsed_args.hook_name, parsed_args.args)
```

dulwich/hooks.py
```diff
@@ -364,3 +364,53 @@ def execute(
...
+
+    Returns:
+        A constructed hook instance ready to be executed
+
+    Raises:
+        HookError: If hook_name does not match a known hook, or refers to
+            a hook not supported by manual execution (e.g. pre-receive,
+            post-receive)
+    """
+    hook_class = HOOK_CLASSES.get(hook_name)
```
jelmer, line 400: Having a dictionary here but also special casing is a bad smell. Furthermore, we're casting return values etc. 

If we're having to write specialized code for every hook anyway, perhaps it's better to not pretend we can run arbitrary hooks and instead just have ``if hook = "commit-msg"`.., etc in cmd_hook

jelmer (changes requested):
Sorry for the delay in reviewing.

I feel like this is trying to rely on an abstraction that doesn't really work - each hook has a different higher level API.

So I think there are two options:

 * special case each hook in cmd_hook and have bespoke handling of inputs/results depending on the hook specified
 * call out to the lower level ShellHook invocation code; you'll need to skip non-shell hooks though

I think the latter is more in line with the idea of being compatible with C git.

jelmer:
@UsamaQaisrani are you still planning to work on this?

UsamaQaisrani:
> @UsamaQaisrani are you still planning to work on this?

Hey @jelmer really sorry for the delay, I'll try to complete it by tomorrow eod...

UsamaQaisrani:
Hey @jelmer, just pushed, changed it the way you suggested: dropped the dict and get_hook and the cast, hooks are dispatched explicitly in cmd_hook now, switched to None for GIT_DIR, and added a commit-msg test.

dulwich/cli.py
```diff
@@ -7714,6 +7721,55 @@ def run(self, args: Sequence[str]) -> None:
...
+        parser.add_argument("args", nargs="*", help="Arguments to pass to the hook")
+        parsed_args = parser.parse_args(args)
+
+        hook_name = parsed_args.hook_name
+        hook_args = [a.encode() for a in parsed_args.args]
+
+        try:
+            with porcelain.open_repo_closing(None) as r:
+                controldir = r.controldir()
+                if hook_name == "pre-commit":
```
jelmer, line 7744: please add tests for each of these three hooks

dulwich/cli.py
```diff
@@ -7714,6 +7721,55 @@ def run(self, args: Sequence[str]) -> None:
...
+                    PreCommitShellHook(r.path, controldir).execute(*hook_args)
+                elif hook_name == "post-commit":
+                    PostCommitShellHook(controldir).execute(*hook_args)
+                elif hook_name == "commit-msg":
+                    msg = CommitMsgShellHook(controldir).execute(*hook_args)
+                    if msg is not None:
+                        sys.stdout.buffer.write(msg)
+                elif hook_name == "update":
+                    out, err = UpdateShellHook(controldir).execute(*hook_args)
```
jelmer, line 7753: What happens if fewer arguments are specified than execute() takes?

dulwich/cli.py
```diff
@@ -7714,6 +7721,55 @@ def run(self, args: Sequence[str]) -> None:
...
+        """Run a hook manually.
+
+        Args:
+            args: Command line arguments
+        """
+        parser = argparse.ArgumentParser()
+        parser.add_argument("hook_name", help="Name of the hook to run")
+        parser.add_argument("args", nargs="*", help="Arguments to pass to the hook")
+        parsed_args = parser.parse_args(args)
+
```
jelmer, line 7737: leading-dash hook arguments are eaten by argparse. nargs="*" means dulwich hook run commit-msg --amend fails with "unrecognized arguments". Real git forwards everything after the hook name. Use parse_known_args, argparse.REMAINDER, or -- handling.

dulwich/cli.py
```diff
@@ -7714,6 +7721,55 @@ def run(self, args: Sequence[str]) -> None:
...
+        hook_name = parsed_args.hook_name
+        hook_args = [a.encode() for a in parsed_args.args]
+
+        try:
+            with porcelain.open_repo_closing(None) as r:
+                controldir = r.controldir()
+                if hook_name == "pre-commit":
+                    PreCommitShellHook(r.path, controldir).execute(*hook_args)
+                elif hook_name == "post-commit":
```
jelmer, line 7746: commit-msg output concatenation. A message passed on argv has no trailing newline, so a hook appending Signed-off-by: yields Initial commitSigned-off-by: T on stdout. The PR's test hides this by passing "Initial commit\n" explicitly and using assertIn.

PR #2369, opened 2026-08-24
collapse consecutive ** segments in wildmatch translation

netliomax25-code:
1. wildmatch translate emits one quantifier per `**` segment, so a run of consecutive `**` produces adjacent unbounded quantifiers like `(?:[^/]+/)*(?:[^/]+/)*`.
2. `dulwich.ignore` and `dulwich.attrs` compile that regex and run it against every working-tree path during status/checkout, so a short pattern such as `a/**/**/**/z` from a cloned repository's `.gitignore` or `.gitattributes` backtracks catastrophically on a long non-matching path (a 57-byte line took about 14s for a single path through `IgnoreFilter.is_ignored`).
3. Git treats a run of `**` as a single `**`, so collapsing consecutive `**` segments before translation is behavior-preserving; I checked that match results are unchanged across single vs repeated `**` and that the existing wildmatch/ignore/attrs suites pass, and added a regression test.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,14 @@
 1.2.13	2026-08-24
 
+ * Collapse a run of consecutive ``**`` segments when translating a
+   wildmatch pattern (``dulwich.wildmatch``). Git treats a run of
+   directory-spanning ``**`` as a single ``**``, but each segment emitted
+   its own ``(?:[^/]+/)*``, so a short ``.gitignore``/``.gitattributes``
+   pattern such as ``a/**/**/**/z`` from an untrusted repository built a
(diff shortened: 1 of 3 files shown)


NEWS
```diff
@@ -1,5 +1,14 @@
 1.2.13	2026-08-24
 
+ * Collapse a run of consecutive ``**`` segments when translating a
```
jelmer: Please keep news entries limited to 1-3 lines

jelmer (changes requested):

jelmer (dismissed):
.

PR #2386, opened 2026-09-06
Honor GIT_OPTIONAL_LOCKS in status

### LINKED ISSUE #1861: Add support for GIT_OPTIONAL_LOCKS environment variable
## Description
Add support for the `GIT_OPTIONAL_LOCKS` environment variable, which allows skipping operations that require locks.

## Background
The `GIT_OPTIONAL_LOCKS` environment variable, when set to 0, tells Git to skip operations that would normally require file locks. This is useful for read-only scenarios or when multiple processes need to access a repository simultaneously.

## Use Case in Dulwich
- Run read-only operations in environments where locking might fail
- Need concurrent access to repositories from multiple processes
- Work in containerized or restricted environments where file locking is problematic

## Implementation Notes
This would require checking the environment variable before attempting to acquire locks on repository files, gracefully skipping lock-dependent operations when set to 0.

be-student:
Fixes #1861.

`dulwich status` now refreshes unchanged index stat metadata by default, matching Git's optional status side effect. `GIT_OPTIONAL_LOCKS=0` skips that write, and lock contention never fails the read-only command. The porcelain API remains side-effect-free unless `update_index=True` is requested.

Validation:
- 598 porcelain tests passed (26 skipped)
- 300 CLI tests passed (1 skipped)
- 174 index tests passed (6 skipped)
- repository-wide Ruff, format, mypy, codespell, and diff checks passed


dulwich/cli.py
```diff
@@ -105,6 +105,15 @@
 logger = logging.getLogger(__name__)
 
 
+def _optional_locks_enabled(env: Mapping[str, str] | None = None) -> bool:
```
jelmer, line 108: I think this belongs in porcelain rather than here - see other code for examples

dulwich/cli.py
```diff
@@ -105,6 +105,15 @@
...
+    """Return whether Git permits optional operations that require locks."""
+    if env is None:
+        env = os.environ
+    if "GIT_OPTIONAL_LOCKS" not in env:
```
jelmer, line 112: No need to query env twice

dulwich/index.py
```diff
@@ -3602,6 +3603,61 @@ def refresh_index(index: Index, root_path: bytes) -> None:
             index[path] = entry
 
 
+def refresh_index_stat(
```
jelmer, line 3606: The whole point is to store the data we've already gathered in the index; this is doing the same work again from scratch

tests/cli/test_cli.py
```diff
@@ -641,6 +641,40 @@ def test_log_reverse(self):
...
+            staged={"add": [], "delete": [], "modify": []},
+            unstaged=[],
+            untracked=[],
+        )
+
+        self._run_cli("status")
+
+        mock_status.assert_called_once_with(None, update_index=True)
+
+    @patch("dulwich.porcelain.status")
```
jelmer, line 656: Please don't use mocks

tests/cli/test_cli.py
```diff
@@ -641,6 +641,40 @@ def test_log_reverse(self):
...
+            unstaged=[],
+            untracked=[],
+        )
+        self.overrideEnv("GIT_OPTIONAL_LOCKS", "0")
+
+        self._run_cli("status")
+
+        mock_status.assert_called_once_with(None, update_index=False)
+
+    @patch("dulwich.porcelain.status")
```
jelmer, line 669: Please don't use mocks

jelmer (changes requested):
Same comment for the PR description as in the last PR.

jelmer:
I'm going to close this - it's slop and wrong in a very obvious (to humans) way and clearly AI generated. That doesn't add any value - I can generate some LLM to generate this myself if that's what I wanted to.

PR #2388, opened 2026-09-06
Add support for git bugreport

dchaudhari7177:
Adds `dulwich bugreport`.

`--diagnose` is not implemented, so `docs/c-git-compatibility.txt` records the command as partial.


dulwich/porcelain/__init__.py
```diff
@@ -1406,6 +1408,140 @@ def var(
         raise KeyError(f"Variable {variable} has no value")
 
 
+# The hook names githooks(5) defines. `git bugreport` lists which of these are
+# present and executable, because a hook firing unexpectedly is a common cause
+# of behaviour a reporter cannot explain.
```
jelmer: This is super brittle, hard to keep up to date

jelmer (changes requested):
Please trim PR description.

See https://jelmer.uk/pages/pr-advice.html

dchaudhari7177:
Description trimmed — thanks for the pointer, the advice page is fair and I had clearly overdone it. Happy to cut it further if it is still long.

dchaudhari7177:
Both addressed.

**The brittle list.** Dropped. The enabled-hooks section no longer filters against a 29-name copy of githooks(5) — it lists what is actually in the hooks directory, minus the `*.sample` templates `git init` leaves behind, sorted so the output does not depend on the filesystem.

That is less code and no maintenance, and it stops hiding the two cases a bug report most needs to show: a hook that is misspelled and therefore never fires, and one from a git newer than whatever list dulwich was carrying. Both used to come back as `<none>`.

Two tests moved with the behaviour — an unrecognised name is now reported rather than dropped, and the order is alphabetical rather than githooks(5) order.

**Description.** Trimmed to the two facts that are not in the commit message.

`tests/porcelain/test_bugreport.py`: 11 passed, 1 skipped (the execute-bit test, on Windows). `tests/cli/test_cli.py -k bugreport`: 6 passed.

dchaudhari7177:
Rebased on current \`main\` (it had gone conflicting on NEWS) and two things brought into line with your comments on the sibling PRs:

- The tests were in \`tests/porcelain/test_bugreport.py\`. They are now in \`tests/porcelain/__init__.py\` with the other porcelain tests — same objection as #2390, so it applies here too.
- NEWS entry cut from eight lines to three.

\`tests/porcelain/__init__.py\` and \`tests/cli/test_cli.py\`: 863 passed, 40 skipped. The nine failures are the pre-existing \`AddTests\`/\`StatusTests\` symlink cases, which need symlink privileges and fail identically on an unmodified tree. \`ruff\` 0.16.5 check and format clean.

jelmer:
> Rebased on current `main` (it had gone conflicting on NEWS) and two things brought into line with your comments on the sibling PRs:
> 
>     * The tests were in `tests/porcelain/test_bugreport.py`. They are now in `tests/porcelain/**init**.py` with the other porcelain tests — same objection as [Add clone --no-tags #2390](https://github.com/jelmer/dulwich/pull/2390), so it applies here too.
> 
>     * NEWS entry cut from eight lines to three.
> 
> 
> `tests/porcelain/**init**.py` and `tests/cli/test_cli.py`: 863 passed, 40 skipped. The nine failures are the pre-existing `AddTests`/`StatusTests` symlink cases, which need symlink privileges and fail identically on an unmodified tree. `ruff` 0.16.5 check and format clean.

This is AI slop, closing as such. https://github.com/jelmer/dulwich/blob/main/CONTRIBUTING.rst#ai-slop

Issue #1827, opened 2025-08-23
Add support for git credential


jelmer:
Add support for the `git credential` command, which retrieves and stores user credentials.

**Description:**
The `git credential` command provides a standardized interface for credential helpers. It can fill, approve, or reject credentials used for authentication with remote repositories.

**Common use cases:**
- Integration with system credential stores (keychain, credential manager)
- Automated authentication in CI/CD pipelines
- Secure credential handling without storing passwords in plain text
- Custom credential helper implementations

This would improve dulwich's authentication capabilities and security for remote operations.

**Note:** There are existing credential-related issues (#1227, #1145) but they focus on specific proxy/URL credential problems rather than implementing the full `git credential` interface.

PR #2389, opened 2026-09-06
Implement the git credential format and helper protocol

dchaudhari7177:
Part of #1827. Implements the credential description format and the `credential.helper` protocol in `dulwich.credentials`.

Split as requested; this is the library half only. #2394 (`StackedConfig.encoding`) is independent and worth reviewing first, and #2395 adds the `dulwich credential` CLI on top of this.


NEWS
```diff
@@ -1,5 +1,20 @@
 1.2.15	UNRELEASED
 
+ * Add ``dulwich credential fill|approve|reject``, implementing the git
```
jelmer: Please keep news entries to 3-4 lines

dulwich/cli.py
```diff
@@ -5073,6 +5082,66 @@ def run(self, args: Sequence[str]) -> None:
         logger.info(porcelain.describe(None))
 
 
+class cmd_credential(Command):
```
jelmer: This should use the porcelain pattern used elsewhere

dulwich/config.py
```diff
@@ -1558,6 +1558,21 @@ def __repr__(self) -> str:
         """Return string representation of StackedConfig."""
         return f"<{self.__class__.__name__} for {self.backends!r}>"
 
+    @property
+    def encoding(self) -> str | None:
```
jelmer, line 1562: Doesn't git assume encoding is utf8? Why is taking the first encoding acceptable, doesn't that lead to issues ogmmif encodings vary?

jelmer (changes requested):
Please trim PR description. 

Please split this PR itself, it's too long.

See https://jelmer.uk/pages/pr-advice.html

dchaudhari7177:
Split and trimmed. This PR is now the library half only — 676 lines rather than 851, and the two pieces that were not the feature are their own PRs:

- #2394 — `StackedConfig.encoding`. Independent of the rest; it is the `AttributeError` I mentioned, and it stands alone.
- #2389 (this) — the credential description format and the `credential.helper` protocol.
- #2395 — `dulwich credential fill|approve|reject`, the CLI wiring, stacked on this.

They are stacked in that order, so #2394 can go in on its own and the other two follow. Descriptions cut down on all three.

dulwich/config.py
dchaudhari7177: Yes, and that question killed the property rather than refined it — it is gone in 647d4b09 (and in #2394, where it originated).

git treats config files as UTF-8, and every value this module decodes is either a credential subsection (a URL, which RFC 3986 restricts to ASCII) or a `helper` / `username` value. So `dulwich.credentials` now decodes as UTF-8 throughout and never reads `config.encoding`:

- `urlmatch_credential_sections` — the subsection
- `helpers_for_url` — the `helper` command
- `fill_credential` — the configured `username`

`dulwich/config.py` already hard-codes `utf-8` for the `GIT_CONFIG_KEY_*` keys, so this is the convention that was already there.

And you are right that varying encodings were the actual problem, not a detail of which one wins. The bytes come from different backends, so applying any single codec across a stack mis-decodes some of them — there is no correct value for an aggregate `StackedConfig.encoding` to return. Since nothing reads the attribute now, a `StackedConfig` works because the question does not arise.

The three annotations widen from `ConfigDict` to `Config` as a result, since `Repo.get_config_stack()` is the realistic caller.

`tests/test_credentials.py` and `tests/test_config.py`: 187 passed. `ruff` 0.16.5 and `mypy` clean. Rebased on current `main`.

dulwich/cli.py
dchaudhari7177: Done, in #2395 where the CLI ended up after the split — 46ac57f8.

`cmd_credential` now parses its arguments, reads stdin and calls `porcelain.credential_fill` / `credential_approve` / `credential_reject`. All three are new; they hold the config resolution and the helper calls, so the behaviour is reachable from the library rather than only from the command line.

Config resolution moved with it: the repository's own stack when there is one (so a per-repository `credential.helper` is honoured), the user/system stack plus `GIT_CONFIG_*` outside a repository — which is how `git credential` is usually driven, and the case the CLI tests could not reach.

`cli.py` is down to 50 added lines and no longer imports `StackedConfig`, `fill_credential`, `helpers_for_url` or `url_for_credential`. `commands` was also out of alphabetical order; fixed.

New `CredentialTests` in `tests/porcelain/__init__.py` covers the three functions directly, including the no-repository fallback. `tests/cli/test_cli.py` and `tests/test_credentials.py`: 349 passed, 4 skipped. `ruff` 0.16.5 and `mypy` clean.

NEWS
dchaudhari7177: Trimmed a while back — the entry is 8 lines now rather than the original block, and the bulk of it is the behaviour a reader needs (helper resolution, the reset-on-empty rule, the newline/NUL refusal). Say the word if you want it cut to the 3-4 lines literally and the rest left to the docstrings.

dchaudhari7177: Cut it properly rather than leaving it to you to decide — 7c09609a. Both entries are 3-4 lines now; the helper-resolution and validation detail lives in the docstrings, where it does not go stale in a changelog.

jelmer:
AI slop. https://github.com/jelmer/dulwich/blob/main/CONTRIBUTING.rst#ai-slop

Issue #1846, opened 2025-08-23
Enhance git clone with missing options


jelmer:
## Summary
The git clone command in dulwich is missing several useful options that are available in standard git.

## Missing Options
- `--shallow-since`: Create a shallow clone with history after a specific date
- `--shallow-exclude`: Create a shallow clone excluding commits reachable from specified refs
- `--single-branch`: Clone only the history leading to the tip of a single branch
- `--no-tags`: Don't clone any tags
- `--recurse-submodules`: Initialize and clone submodules recursively

## Benefits
These options would improve dulwich's git clone functionality by:
- Enabling shallow clones with specific date constraints using `--shallow-since`
- Providing more control over shallow clone boundaries with `--shallow-exclude`
- Reducing clone size and time with `--single-branch` and `--no-tags`
- Supporting projects with submodules through `--recurse-submodules`

This would bring dulwich's clone command closer to feature parity with standard git and better support modern development practices like CI/CD pipelines that benefit from faster, smaller clones.

PR #2390, opened 2026-09-06
Add clone --no-tags

dchaudhari7177:
Part of #1846. Clones `HEAD` and the branches only, and records `remote.<origin>.tagOpt = --no-tags` in the new repository as git does, so the next bare `fetch` does not pull the tags back in.

`ls-refs`' `ref-prefix` has no negative form, so the flag narrows to `[HEAD, refs/heads/]` rather than subtracting `refs/tags/`. Stacked on #2393, which fixes `LocalGitClient.fetch` ignoring `ref_prefix`; without it the flag is a no-op for local clones.


tests/porcelain/test_clone_no_tags.py
```diff
@@ -0,0 +1,177 @@
+# test_clone_no_tags.py -- tests for porcelain.clone(no_tags=...)
```
jelmer, line 1: This shouldn't be in a separate file

NEWS
```diff
@@ -1,5 +1,15 @@
 1.2.15	UNRELEASED
 
+ * Add ``dulwich clone --no-tags``, which clones only ``HEAD`` and the
+   branches and records ``remote.<origin>.tagOpt = --no-tags`` so a later
+   bare ``fetch`` does not pull the tags in either.
+   (Dipak Chaudhari, #1846)
+
+ * ``LocalGitClient.fetch`` now honours ``ref_prefix``. It accepted the
```
jelmer: This should be a separate PR

dulwich/cli.py
```diff
@@ -2044,6 +2044,11 @@ def run(self, args: Sequence[str]) -> int | None:
             action="store_true",
             help="Initialize and clone submodules",
         )
+        parser.add_argument(
```
jelmer, line 2047: This doesn't have any tests

dulwich/client.py
```diff
@@ -241,6 +241,7 @@ def __call__(
 )
 from .refs import (
     HEADREF,
+    LOCAL_BRANCH_PREFIX,
```
jelmer, line 244: The changes in this file are untested.

jelmer (commented):
Please also fix the PR description. See also https://jelmer.uk/pages/pr-advice.html

jelmer (changes requested):
Please also fix the PR description

dchaudhari7177:
Description rewritten, and the unrelated fix is now #2393 with this PR stacked on it — the flag is a silent no-op for local clones without it, which is why they were together. Also trimmed the comments in the diff.

jelmer:
> Description rewritten, and the unrelated fix is now #2393 with this PR stacked on it — the flag is a silent no-op for local clones without it, which is why they were together. Also trimmed the comments in the diff.

Please follow up to PR comments, and trigger re-review.

tests/porcelain/test_clone_no_tags.py
dchaudhari7177: Done in 2e4864e — the porcelain cases are now three methods on `CloneTests` in `tests/porcelain/__init__.py` and `tests/porcelain/test_clone_no_tags.py` is gone. The `LocalGitClient.fetch` `ref_prefix` cases that were in it went with it; those belong with the other `LocalGitClient` tests and are in #2393.

dulwich/cli.py
dchaudhari7177: Added in 2e4864e: `CloneCommandTest` in `tests/cli/test_cli.py`, which runs `clone` through `cli.main` rather than calling the porcelain.

Two cases — the default (tags cloned) and `--no-tags` (no `refs/tags/*`, `refs/heads/master` present, `remote.origin.tagOpt = --no-tags` recorded). There was no clone CLI test class at all before, so this is a new one.

One thing it turned up: `cmd_clone` discarded the repository `porcelain.clone` returns without closing it, so its pack files stayed open. On Windows that locks the freshly cloned directory, and the test could not clean up after itself. Fixed in the same commit by closing it — one line of real change, and it is why the test can run on Windows.

dulwich/client.py
dchaudhari7177: Fair — after the `ref_prefix` fix moved to #2393, what was left of `client.py` here was the `no_tags` plumbing, and nothing touched it directly. Covered in 2e4864e:

- `test_clone_no_tags` — a `LocalGitClient.clone(no_tags=True)` has no `refs/tags/*`, keeps `refs/heads/master`, `refs/remotes/origin/master` and `HEAD`, does **not** have the tag object in its object store, and records `tagOpt`.
- `test_clone_no_tags_keeps_an_explicit_ref_prefix` — `no_tags` only supplies a *default* prefix. A caller that passed its own keeps it, so the flag cannot silently widen someone else's request. That is the branch of the condition nothing else reaches.

Plus the porcelain and CLI layers noted on the other two threads.

Verification: `tests/porcelain/__init__.py`, `tests/test_client.py` and `tests/cli/test_cli.py` together are 1097 passed / 36 skipped. Nine failures are the pre-existing `AddTests`/`StatusTests` symlink cases, which need symlink privileges and fail the same way on an unmodified tree. `ruff` 0.16.5 check and format clean over `dulwich/` and `tests/`; `mypy` clean.

dchaudhari7177:
@jelmer all review comments are now replied to inline and addressed, here and on #2388, #2389, #2393 and #2394 (pushed 2026-09-12). Could you re-review when you have a moment? I can't trigger the re-request myself from a fork.

jelmer:
https://github.com/jelmer/dulwich/blob/main/CONTRIBUTING.rst#ai-slop

PR #2393, opened 2026-09-06
Honour ref_prefix in LocalGitClient.fetch

dchaudhari7177:
`LocalGitClient.fetch` accepts `ref_prefix` and ignores it, so a local fetch returns every ref. Its docstring promises filtering is done "by the server if supported, and client side otherwise", and a local fetch has no server.

The wants are filtered as well as the returned refs, so objects reachable only from an unrequested ref are not copied.

Split out of #2390 at your request; that PR needs this fix to be testable.


dulwich/client.py
```diff
@@ -3117,12 +3117,25 @@ def fetch(
 
         """
         with self._open_repo(path) as r:
+            if ref_prefix is not None and determine_wants is None:
```
jelmer, line 3120: this should probably raise an exception if both ref_prefix and determine_wants are specified

tests/test_client.py
```diff
@@ -1391,6 +1391,27 @@ def test_fetch_into_empty(self) -> None:
...
+
+        result = c.fetch(s.path, t, ref_prefix=[b"refs/heads/"])
+
+        self.assertEqual(
+            {b"refs/heads/master": b"a90fa2d900a17e99b433217e988c4eb4a2e9a097"},
+            result.refs,
+        )
+        # The wants are filtered too, so an unrequested ref's objects are not
+        # transferred -- not merely hidden from the result.
+        self.assertNotIn(b"28237f4dc30d0d462658d6b937b08a0f0b6ef55a", t.object_store)
```
jelmer, line 1413: this should be checking refs as well

jelmer (changes requested):

dulwich/client.py
dchaudhari7177: Composed them instead of rejecting the combination, in 85721de — happy to switch to raising if you would rather.

The prefix is now applied to the refs *before* `determine_wants` sees them, which is what `ls-refs`' `ref-prefix` does on a remote fetch: the server advertises the admitted refs and `determine_wants` chooses among those. So the two arguments compose with the same meaning locally as remotely, and the silent half-honouring you spotted is gone — the prefix restricts the wants whether or not `determine_wants` was passed.

Raising would make `LocalGitClient` the only client that refuses a pair `TraditionalGitClient` and `HttpGitClient` both accept, and would break a caller passing a `determine_wants` alongside a prefix that it already expects the server to apply. That is the only reason I did not take the suggestion literally.

tests/test_client.py
dchaudhari7177: Done in 85721de.

The test asserted a hardcoded expected dict without ever showing what it was filtering *out*, so it would have passed against a fixture whose only ref was `refs/heads/master`. The setup now asserts the source repo's full ref set (`HEAD`, `refs/heads/master`, `refs/tags/mytag`, `refs/tags/mytag-packed`) before fetching, so the prefix is demonstrably excluding something, and the object-store assertions are two-sided: the wanted commit is present, both tag objects are absent.

Two more cases on the same helper: a prefix matching nothing (empty refs, nothing transferred) and a prefix combined with `determine_wants`, which records exactly which refs it was offered.

`LocalGitClientTests`: 17 passed. `ruff` 0.16.5 and `mypy` clean.

jelmer: Please don't add all of this LLM generated crap, write PR comments yourself and keep them brief and to the point.

PR #2394, opened 2026-09-06
Decode credential subsections as UTF-8

dchaudhari7177:
You are right that the aggregate property was wrong, so the fix moved rather than got patched.

`urlmatch_credential_sections` reads one `encoding` off the config and applies it to every section. In a stack that decodes one backend's bytes with another backend's codec. `StackedConfig` also has no `encoding` at all, so passing it the type `Repo.get_config_stack()` returns raised `AttributeError`.

git documents config files as UTF-8, and a credential subsection is a URL, which RFC 3986 restricts to ASCII — so the value is decoded as UTF-8 and the annotation widened to `Config`. `StackedConfig.encoding` is gone.

Split out of #2389 at your request.


tests/test_config.py
```diff
@@ -870,6 +870,16 @@ class StackedConfigTests(TestCase):
...
 
+    def test_encoding_no_backends(self) -> None:
+        self.assertIs(None, StackedConfig([]).encoding)
+
+    def test_encoding_from_first_backend_declaring_one(self) -> None:
+        first = ConfigDict()
+        first.encoding = None
+        second = ConfigDict(encoding="latin-1")
+        third = ConfigDict(encoding="utf-8")
+        self.assertEqual("latin-1", StackedConfig([first, second, third]).encoding)
```
jelmer, line 881: This is going to just silently corrupt config from the other ConfigDict.

jelmer (changes requested):

tests/test_config.py
dchaudhari7177: Right, and that kills the idea rather than the implementation — so `StackedConfig.encoding` is gone in 5b5d92b.

Applying any single codec across a stack is wrong by construction: the bytes come from different backends, and picking one backend’s codec mis-decodes the others. There is no correct aggregate to return.

To your question on #2389: yes, git treats config as UTF-8, and a credential subsection is a URL, which RFC 3986 restricts to ASCII — so `urlmatch_credential_sections` now decodes the subsection as UTF-8 and ignores `config.encoding` entirely. That makes `StackedConfig` work because nothing reads the attribute any more, not because it grew one. `dulwich/config.py` already hard-codes `utf-8` for the `GIT_CONFIG_KEY_*` keys, so this matches what is there.

Two details:

- The annotation widens from `ConfigDict` to `Config`, since a config stack is the realistic caller.
- Decoding uses `errors="replace"` rather than strict, like the `includeIf` path handling above it. One junk subsection then fails to match instead of raising out of the whole credential lookup; `U+FFFD` cannot match a real host. Say the word if you would rather it raise.

Tests: a `StackedConfig` across two backends, the mixed latin-1/utf-8 stack in both orders, and an undecodable subsection. `tests/test_credentials.py` and `tests/test_config.py` are 146 passed; `ruff` 0.16.5 and `mypy` clean.

jelmer: This comment was very clearly AI generated, overly verbose and contains useless slop.

jelmer:
https://github.com/jelmer/dulwich/blob/main/CONTRIBUTING.rst#ai-slop

PR #2396, opened 2026-09-07
Support GIT_OPTIONAL_LOCKS in status

### LINKED ISSUE #1861: Add support for GIT_OPTIONAL_LOCKS environment variable

sudhanshuskarn:
Fixes #1861.

Adds support for GIT_OPTIONAL_LOCKS to status operations. The environment is handled at the porcelain/CLI layer and passed to the lower-level index logic as an explicit option.

With GIT_OPTIONAL_LOCKS=0, status avoids optional index refresh writes. With optional locking enabled or unset, unchanged entries with stale stat information can be refreshed and persisted.

Tests cover the default behavior, disabled and explicitly enabled optional locks, and invalid values.


dulwich/index.py
```diff
@@ -3227,13 +3227,28 @@ def _stat_matches_entry(
     return True
 
 
+def _refresh_index_entry_stat(entry: IndexEntry, st: os.stat_result) -> IndexEntry:
+    """Return an index entry with refreshed filesystem stat information."""
+    fresh_entry = index_entry_from_stat(st, entry.sha, mode=entry.mode)
```
jelmer, line 3232: this seems backwards; can't we copy the missing data from entry into freshentry rather than the other way around?

sudhanshuskarn: Updated, thanks. I now use the fresh entry as the base and only preserve `flags` and `extended_flags` from the existing entry.

dulwich/index.py
```diff
@@ -3344,22 +3360,36 @@ def get_unstaged_changes(
...
+
+            if update_index and refreshed_entry is not None:
+                index[tree_path] = refreshed_entry
+                index_updated = True
+
+            if changed_path is not None:
+                yield changed_path
+
+    if update_index and index_updated:
+        index.write()
```
jelmer, line 3392: status() now fails on a read-only repo

This makes the index write unconditional when locks are enabled, and never catches failure:

```
  status with read-only .git RAISED: PermissionError [Errno
  13] .../.git/index.lock
```

On main the same case returns [] cleanly. Git tolerates this.

dulwich/index.py
```diff
@@ -3301,8 +3314,11 @@ def get_unstaged_changes(
       trust_ctime: If True, use ctime for change detection (default: True)
       max_stat: If set, limit the number of stat operations performed.
         When the limit is reached, remaining files are assumed unchanged.
+      update_index: Whether to refresh unchanged entries with updated stat information.
     Returns: iterator over paths with unstaged changes
     """
+    index_updated = False
```
jelmer, line 3320: status() reads the index, holds it while scanning the worktree, then writes it back with no re-read or staleness check. A concurrent git add during the scan is lost.

jelmer (changes requested):

dulwich/index.py
```diff
@@ -3344,22 +3360,36 @@ def get_unstaged_changes(
...
+                        index[tree_path] = refreshed_entry
+                        index_updated = True
+
+                    if changed_path is not None:
+                        yield changed_path
 
     if not preload_index:
         # Serial processing
         for tree_path, entry in index.iteritems():
             if max_stat is not None and stat_count >= max_stat:
```
jelmer, line 3377: This corrupts the index I think?

sudhanshuskarn: Thanks, I see the issues. I was treating the index refresh/write as unconditional once `status()` detected changes, but I hadn't accounted for read-only repositories or the possibility of the index changing while the worktree is being scanned. I'll rework this rather than trying to patch the individual failures.

sudhanshuskarn:
## Closing this PR

I'm closing this PR for now because I wasn't able to get to a reliable fix for the issue without making assumptions about how Dulwich is expected to handle the index lock in this case.

The main difficulty I ran into was the interaction between the optional lock behavior and the existing index refresh/write flow. The relevant path goes through `get_unstaged_changes()`, where the index can be refreshed and subsequently written back. I tried to handle the case where the index cannot be locked, but the changes I tested either did not resolve the original behavior or moved the problem to the index write path.

What I wasn't able to determine confidently is where the optional-lock behavior is supposed to be handled:

- when acquiring the index lock,
- in the caller before attempting to update/write the index, or
- in the index write path itself.

I don't want to work around this by simply catching and ignoring a `PermissionError`, because that could hide a real index write failure and potentially leave the behavior incorrect.

I think a follow-up would be better served by tracing the complete path from the optional-lock option through the index refresh/update and finally the index lock/write operation, and then comparing that with the expected Git behavior. That should clarify which layer is responsible for handling an unavailable optional lock.

I'm closing this rather than leaving an incomplete fix open. Thanks for the review and feedback.

PR #2397, opened 2026-09-07
Name a symlink by its own path in path_to_tree_path

### LINKED ISSUE #2402: porcelain.status reports a tracked symlink as untracked when its target is untracked
`porcelain.status()` reports a tracked symlink as untracked when its target is untracked or missing.

Reproducer:

```python
from pathlib import Path
from dulwich import porcelain

porcelain.init("repo")
Path("repo/link").symlink_to("target")
porcelain.add("repo", paths=["link"])
porcelain.commit("repo", message=b"add link")
Path("repo/target").write_text("data")
print(porcelain.status("repo", untracked_files="all").untracked)
```

Actual: `[b"link", b"target"]`

Expected, matching `git status --porcelain -uall`: `[b"target"]`

A modified tracked symlink can likewise appear in both `unstaged` and `untracked`.

undeemed:
Fixes #2402.

`path_to_tree_path()` followed symlinks through `Path.resolve()`, so callers received the target's tree path instead of the link's path.

Resolve only the parent directory for symlinks. This keeps the symlink's own index key and makes `porcelain.status()` match Git when the target is untracked or missing.

Tests cover the helper, a tracked link to an untracked file, and a modified link to a missing target.

Verified with the Dulwich test suites, `ruff`, `mypy`, and Git 2.53.0.

This change was made with AI assistance.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,11 @@
    ``ssh://user@host/git/repo.git`` as the remote URL, which points at a
    different repository. (Jelmer Vernooĳ, #2375)
 
+ * Name a symlink by its own path in ``porcelain.path_to_tree_path``, rather
+   than by the path of its target. ``porcelain.status`` listed a tracked
+   symlink as untracked whenever the file it pointed at was itself untracked
+   or missing, so a modified symlink was reported twice. (Jerry Xiao)
+
 1.2.14	2026-08-28
 
  * Collapse consecutive `**` segments in wildmatch translation, matching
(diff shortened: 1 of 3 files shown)


jelmer (changes requested):
Please file a bug ahead of sending a bug fix PR; also, your PR description is way too long and contains LLM-generated nonsense.

undeemed:
Filed the bug as #2402 and replaced the PR description with a shorter version focused on the cause, fix, and tests. Thanks for the direct feedback.

jelmer:
> Filed the bug as #2402 and replaced the PR description with a shorter version focused on the cause, fix, and tests. Thanks for the direct feedback.

Thanks!

jelmer (approved):

Issue #2399, opened 2026-09-08
A `.gitignore` in a subdirectory loses to one closer to the root: 22 of 28 divergences from git on a 4,463-query bench


KaizenShogun:
I ran dulwich against an outside conformance bench for `.gitignore` *trees* and it comes out at
**28 divergences from git in 4,463 queries over 33 real repositories**. 22 of them are one
mechanism and I can point at the line; the other 6 are a different one, at the bottom.

This is about behaviour, not a code review: I read `ignore.py` to form the hypothesis and then
measured it. Everything below is a number rather than an opinion, and every repro was reduced by
re-asking both git and dulwich at each step, so the small trees are measured, not deduced.

### Setup

* **Subject:** dulwich 1.2.14 from PyPI, and the tip of `main` (`2d728c2a`). For this file they are
  the same code: `dulwich/ignore.py` is byte-identical in both (md5
  `7f49c383cad6a0c13dd0031bc0bfec9e`), so nothing merged since the release changes any of it.
* **Oracle:** `git check-ignore -v --non-matching --no-index` (git 2.55.0) over the *identical*
  tree. `rc=0` means *decided*, not *ignored*, so the verdict comes from whether the winning
  pattern starts with `!`.
* **Corpus:** every `.gitignore` of 33 popular repositories, materialised as a real tree, queried
  for paths in five positions relative to each rule file (the rule file's own directory, deeper,
  from the root, a sibling, the repo root). Each repository is asked twice — once about files, once
  about the directories on the way to them — which is where the 66 cases and 4,463 queries come
  from.
* **Entry points:** `IgnoreFilterManager.is_ignored` and, separately,
  `porcelain.check_ignore(..., no_index=True)`. **Both produce the same 28**, so this is not an
  internal-API detail — it is what `dulwich check-ignore` prints.
* No `core.excludesFile` and no `~/.config/git/ignore` exist on the measuring machine, so global
  excludes are not in play on either side.

### The mechanism: between two rule files, the shallower one decides

git gives a deeper `.gitignore` precedence over one closer to the root. `find_matching`
accumulates filters with `filters.insert(0, ...)` (`ignore.py:601`), so the inner loop walks them
deepest-first, the match list ends with the *shallowest* file's match, and `is_ignored` takes
`matches[-1]`.

Reduced from `nodejs/node`:

```
.gitignore                                        !deps/v8/**
deps/v8/third_party/ittapi/ittapi-rs/.gitignore   Cargo.lock
```

```
[...]

PR #2404, opened 2026-09-09
Ignore gitignore lines that carry no pattern

### LINKED ISSUE #2398: A bare `!` line re-includes an ignored directory
A `.gitignore` line consisting of a single `!` makes dulwich stop ignoring a directory that an
earlier pattern in the same file ignores. git makes nothing of that line.

```
$ printf 'build/\n!\n' > .gitignore && mkdir build

$ git check-ignore -v --no-index build/
.gitignore:1:build/	build/
```

```python
>>> from dulwich.ignore import IgnoreFilterManager
>>> from dulwich.repo import Repo
>>> m = IgnoreFilterManager.from_repo(Repo("."))
>>> m.is_ignored("build/")
False
>>> list(porcelain.check_ignore(".", ["build/"], no_index=True))
[]
```

**Control, one variable changed:** replace the `!` line with `#c`, another line git turns into no
pattern at all, and those two come back `True` and `['build/']`, agreeing with git. Put the `!`
back and they flip again. So it is the `!` line, not the directory pattern.

It only bites directory queries: with `x` and `!` in the file, `is_ignored("x")` is `True`, which
matches git.

Measured with dulwich 1.2.14 from PyPI; `dulwich/ignore.py` is byte-identical on the tip of `main`
(`2d728c2a`), md5 `7f49c383cad6a0c13dd0031bc0bfec9e`, so this is not fixed there. Oracle is
`git check-ignore --no-index`, git 2.55.0, on the same tree.

Not a hypothetical input, either. `psf/black` ships
`tests/data/invalid_nested_gitignore_tests/a/.gitignore` whose only line is `!`, and I hit this
while running dulwich over trees built from real repositories. The Rust `ignore` crate has a bug of
the same family ([ripgrep#3527](https://github.com/BurntSushi/ripgrep/issues/3527)), where a bare
`!` compiles to a whitelist that matches everything — different consequence, same corner.

Found with a `.gitignore` conformance bench:
https://github.com/KaizenShogun/gitignore-conformance. The tree-precedence findings from the same
run are in a separate issue; this one is independent of those.

jelmer:
A line like a bare `!` or `/` has nothing left once the negation prefix and the trailing slash are stripped. Git keeps such a line in its pattern list but it can never match, since the empty pattern is matched against a basename. Dulwich instead compiled it to a regex that matched any path ending in a slash, so a bare `!` after `build/` in the same file re-included the directory.

Fixes #2398

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -31,6 +31,10 @@
    symlink as untracked whenever the file it pointed at was itself untracked
    or missing, so a modified symlink was reported twice. (Jerry Xiao)
 
+ * Ignore gitignore lines that carry no pattern, such as a bare ``!`` or
+   ``/``. A bare ``!`` previously re-included a directory that an earlier
+   pattern in the same file had ignored. (Jelmer Vernooĳ, #2398)
+
 1.2.14	2026-08-28
 
  * Collapse consecutive `**` segments in wildmatch translation, matching
--- a/dulwich/ignore.py
+++ b/dulwich/ignore.py
@@ -283,6 +283,11 @@ class Pattern:
         # Check if this is a directory-only pattern
         self.is_directory_only = pattern.endswith(b"/")
 
+        # Lines like "!" or "/" carry no pattern once the negation prefix and
+        # the directory suffix are stripped. Git keeps them in its pattern list
+        # but they can never match a path, so neither do we.
+        self.is_empty = not pattern.rstrip(b"/")
+
         flags = 0
         if self.ignorecase:
             flags = re.IGNORECASE
@@ -334,6 +339,9 @@ class Pattern:
           path: Path to match (relative to ignore location)
         Returns: boolean
         """
+        if self.is_empty:
+            return False
+
         # For negation directory patterns (e.g., !dir/), only match directories
         if self.is_directory_only and not self.is_exclude and not path.endswith(b"/"):
             return False
--- a/tests/test_ignore.py
+++ b/tests/test_ignore.py
@@ -424,6 +424,16 @@ class IgnoreFilterTests(TestCase):
         self.assertEqual([Pattern(b"!c.c")], list(filter.find_matching(b"c.c")))
         self.assertEqual([], list(filter.find_matching(b"d.c")))
 
+    def test_empty_pattern_never_matches(self) -> None:
+        # Lines that carry no pattern once the negation prefix and the
+        # directory suffix are stripped are inert in git; a bare "!" in
+        # particular does not re-include an earlier ignored directory.
+        for line in [b"!", b"!/", b"/", b"//", b"!//"]:
+            filter = IgnoreFilter([b"build/", line])
+            self.assertIs(True, filter.is_ignored(b"build/"), line)
+            self.assertIs(True, filter.is_ignored(b"build/f"), line)
+            self.assertEqual([], list(filter.find_matching(b"other/")), line)
+
     def test_include_exclude_include(self) -> None:
         filter = IgnoreFilter([b"a.c", b"!a.c", b"a.c"])
         self.assertTrue(filter.is_ignored(b"a.c"))


KaizenShogun:
I opened #2398, so read this as an interested party's measurement rather than a review: I ran the
branch against an outside bench before it lands, because the number worth having about a matcher
change is the one that doesn't come from the tests shipping in the same commit.

**Setup.** Two trees of this repo pulled the same day, `main` and `issue-2398-bare-negation`, both
reporting `1.2.14`, driven through the same adapter (`IgnoreFilterManager.is_ignored`). Oracle:
`git check-ignore --no-index`, git 2.55.0, over the identical tree. Corpus: the rule files of 33
large repositories as written, 8,953 queries, plus a separate 1,010 asking about the intermediate
directories of each path.

| tree | divergences from git, 8,953 | 1,010 |
|---|---:|---:|
| `main` | 68 | 1 |
| this PR | **66** | 1 |

Compared as sets rather than counts, because 68 → 66 is also what "fixes 6, breaks 4" looks like:
**2 fixed, 0 broken.** Both fixed cases come from `psf/black`'s
`tests/data/invalid_nested_gitignore_tests/`, whose `a/.gitignore` is a single bare `!` line, with
the exclusions it appeared to undo (`.coverage/`, `.venv/`) coming from higher up the tree — real
code, not a constructed case.

Two is small because a bare `!` is rare in the wild; the row that earns the patch is the zero on
the other side. Nothing else in 9,963 queries moves.

What it doesn't reach, so the number isn't read too widely: the 66 that remain are other families —
21 on the excluded directory itself, 38 on what's inside it, 7 deeper. Most of those are #2399
(a subdirectory's `.gitignore` losing to one closer to the root), which this patch doesn't claim.

Bench: https://github.com/KaizenShogun/gitignore-conformance

— Midas

jelmer:
> I opened #2398, so read this as an interested party's measurement rather than a review: I ran the branch against an outside bench before it lands, because the number worth having about a matcher change is the one that doesn't come from the tests shipping in the same commit.
> 
> **Setup.** Two trees of this repo pulled the same day, `main` and `issue-2398-bare-negation`, both reporting `1.2.14`, driven through the same adapter (`IgnoreFilterManager.is_ignored`). Oracle: `git check-ignore --no-index`, git 2.55.0, over the identical tree. Corpus: the rule files of 33 large repositories as written, 8,953 queries, plus a separate 1,010 asking about the intermediate directories of each path.
> tree 	divergences from git, 8,953 	1,010
> `main` 	68 	1
> this PR 	**66** 	1
> 
> Compared as sets rather than counts, because 68 → 66 is also what "fixes 6, breaks 4" looks like: **2 fixed, 0 broken.** Both fixed cases come from `psf/black`'s `tests/data/invalid_nested_gitignore_tests/`, whose `a/.gitignore` is a single bare `!` line, with the exclusions it appeared to undo (`.coverage/`, `.venv/`) coming from higher up the tree — real code, not a constructed case.
> 
> Two is small because a bare `!` is rare in the wild; the row that earns the patch is the zero on the other side. Nothing else in 9,963 queries moves.
> 
> What it doesn't reach, so the number isn't read too widely: the 66 that remain are other families — 21 on the excluded directory itself, 38 on what's inside it, 7 deeper. Most of those are #2399 (a subdirectory's `.gitignore` losing to one closer to the root), which this patch doesn't claim.
> 
> Bench: https://github.com/KaizenShogun/gitignore-conformance
> 
> — Midas

Hi Midas,

While I appreciate the research in general, please don't post overly verbose LLM-generated comments.

KaizenShogun:
Understood, and taken. You're the second maintainer to tell me this in a week, so it's me, not the
house — I'll keep it to the delta and a link, and let anyone who wants the setup ask for it.

Thanks for the fix.

PR #2405, opened 2026-09-11
Honor per-path eol attributes

### LINKED ISSUE #2403: gitattributes `eol=crlf` / `eol=lf` is ignored on checkout
> Disclaimer: AI generated, but I verified the problem with an E2E test reproducing the bug we'd seen in Production. I will answer any questions personally

## Summary

The `eol` gitattribute has no effect. On checkout, a path matching `text eol=crlf` is written with LF, and — when `core.autocrlf=true` — a path matching `text eol=lf` is written with CRLF. In both cases the working tree gets the opposite of what the attribute asks for.

The string `eol` never appears as an attribute name anywhere in the source:

```console
$ grep -rn "b\"eol\"\|b'eol'" dulwich/
$ 
```

Only `text` is consulted, and the single `text` driver derives its smudge behavior from `core.autocrlf` alone.

## Versions

- dulwich `main` @ `e27b0cc` (reports `1.2.14`) — reproduces
- dulwich `1.2.12` from PyPI — reproduces
- Python 3.13.9, Linux (x86_64)
- git 2.53.0 used for the reference behavior below

## Reproduction

```python
import tempfile
from pathlib import Path

from dulwich import porcelain

CONTENT_LF = b"first line\nsecond line\n"
CONTENT_CRLF = b"first line\r\nsecond line\r\n"


def check(*, autocrlf: bytes | None) -> None:
    with tempfile.TemporaryDirectory() as temp_dir:
        repo_path = Path(temp_dir)
        repo = porcelain.init(str(repo_path))
        if autocrlf is not None:
            config = repo.get_config()
            config.set((b"core",), b"autocrlf", autocrlf)
            config.write_to_path()

        (repo_path / ".gitattributes").write_bytes(b"*.crlffile text eol=crlf\n*.lffile text eol=lf\n")
        (repo_path / "sample.crlffile").write_bytes(CONTENT_LF)
        (repo_path / "sample.lffile").write_bytes(CONTENT_LF)
        porcelain.add(repo, paths=[".gitattributes", "sample.crlffile", "sample.lffile"])
        porcelain.commit(repo, message=b"add files", author=b"a <a@e.com>", committer=b"a <a@e.com>")

        (repo_path / "sample.crlffile").unlink()
        (repo_path / "sample.lffile").unlink()
        porcelain.reset(repo, mode="hard", treeish=b"HEAD")

        print(f"core.autocrlf={autocrlf!r}")
        print(f"  sample.crlffile  actual={(repo_path / 'sample.crlffile').read_bytes()!r}  expect={CONTENT_CRLF!r}")
        print(f"  sample.lffile    actual={(repo_path / 'sample.lffile').read_bytes()!r}  expect={CONTENT_LF!r}")


check(autocrlf=None)
check(autocrlf=b"true")
```

### Actual

```
core.autocrlf=None
  sample.crlffile  actual=b'first line\nsecond line\n'  expect=b'first line\r\nsecond line\r\n'
  sample.lffile    actual=b'first line\nsecond line\n'  expect=b'first line\nsecond line\n'
core.autocrlf=b'true'
  sample.crlffile  actual=b'first line\r\nsecond line\r\n'  expect=b'first line\r\nsecond line\r\n'
  sample.lffile    actual=b'first line\r\nsecond line\r\n'  expect=b'first line\nsecond line\n'
```

Two wrong outcomes:

- `core.autocrlf` unset (default): `eol=crlf` gets LF.
- `core.autocrlf=true`: `eol=lf` gets CRLF.

The `eol` attribute is documented to be independent of `core.autocrlf` and `core.eol`, so `eol=crlf` should yield CRLF and `eol=lf` should yield LF under both configurations.

### Expected (git 2.53.0, same repo layout)

```
core.autocrlf=false  sample.crlffile: ...0d0a...  sample.lffile: ...0a...
core.autocrlf=true   sample.crlffile: ...0d0a...  sample.lffile: ...0a...
```

git honors the attribute in both cases, and the attribute wins over `core.autocrlf`.

`porcelain.checkout()` switching between two branches reproduces this identically — it is not specific to `reset --hard`.

## Analysis

1. `Repo.get_blob_normalizer()` (`dulwich/repo.py:2631`) returns a plain `filters.FilterBlobNormalizer` built on a bare `FilterRegistry`.
2. `get_filter_for_path()` (`dulwich/filters.py:811`) reads only `attributes.get(b"text")`. When it is `True` it fetches the single driver named `"text"`. `eol` is never read.
3. That driver comes from `FilterRegistry._create_text_filter()` (`dulwich/filters.py:778`) → `LineEndingFilter.from_config(config, for_text_attr=True)` (`dulwich/line_ending.py:194`), where `smudge_conversion = get_smudge_filter(core_eol, autocrlf)`.
4. `get_smudge_filter()` (`dulwich/line_ending.py:390`) delegates to `get_smudge_filter_autocrlf()` (`dulwich/line_ending.py:406`), which returns `None` unless `core.autocrlf == b"true"`. The `core_eol` argument is accepted and discarded.

So the smudge side is driven entirely by `core.autocrlf`, and a single shared `"text"` driver serves every `text` path regardless of its `eol` value — which is why the two failure directions above are mirror images of each other.

Secondary observation: `line_ending.BlobNormalizer` (`dulwich/line_ending.py:441`) is exported in `__all__` but never instantiated anywhere in dulwich; `Repo.get_blob_normalizer` uses `FilterBlobNormalizer` instead. It would not fix this on its own, though — its `"text"` driver hardcodes `smudge_conversion=smudge_filter or convert_lf_to_crlf`, which would return CRLF for `eol=lf` paths too.

## Suggested direction

Resolve the smudge/clean conversion per path from the `eol` attribute when present, falling back to `core.eol` / `core.autocrlf` only when it is absent. Since `get_filter_for_path` already has the matched attribute dict in hand, that likely means selecting (or parameterizing) a per-path `LineEndingFilter` rather than reusing one registry-wide `"text"` driver. Git's precedence is `eol` attribute > `text=auto` + `core.eol`/`core.autocrlf`.

## Impact

This surfaces in repositories that pin per-extension line endings because a downstream consumer requires them. A tree holding two file types that need different endings — one CRLF, one LF, side by side — is expressed exactly as gitattributes intends:

```
some/dir/**/*.ext-a text eol=crlf
some/dir/**/*.ext-b text eol=lf
```

A `git` checkout produces CRLF for the `.ext-a` files; a dulwich checkout of the same commit produces LF. Every one of those files then reads as modified against the bytes the downstream consumer expects, and writing the working tree back out would silently strip the CRLFs it requires.

GruffElixir:
## Summary

- honor per-path `eol=lf` and `eol=crlf` attributes during checkout
- make those attributes override `core.autocrlf`
- add regression coverage and a NEWS entry

Fixes #2403

## Tests

- `uv run --extra dev --with pytest python -m pytest tests/test_line_ending.py -q`
- `uv run --extra dev ruff check dulwich/filters.py dulwich/line_ending.py tests/test_line_ending.py`
- porcelain checkout reproduction with both `eol` attributes and `core.autocrlf=true`


jelmer (changes requested):
Please fix mypy, tests, etc.

NEWS
```diff
@@ -31,6 +31,9 @@
    symlink as untracked whenever the file it pointed at was itself untracked
    or missing, so a modified symlink was reported twice. (Jerry Xiao)
 
+ * Honor per-path ``eol=lf`` and ``eol=crlf`` attributes during checkout,
+   independent of ``core.autocrlf``. (#2403)
```
jelmer, line 35: Please add author here

jelmer (changes requested):
As far as I can tell, eol without text attribute is still not handled?

dulwich/line_ending.py
```diff
@@ -252,7 +260,12 @@ def from_config(
         if for_text_attr:
             # For text attribute: always normalize to LF on checkin
             # Smudge behavior depends on core.eol and core.autocrlf
-            smudge_filter = get_smudge_filter(core_eol, autocrlf)
+            if eol in ("crlf", b"crlf"):
+                smudge_filter = convert_lf_to_crlf
```
jelmer, line 264: This bypasses any other filters!

dulwich/line_ending.py
```diff
@@ -252,7 +260,12 @@ def from_config(
         if for_text_attr:
             # For text attribute: always normalize to LF on checkin
             # Smudge behavior depends on core.eol and core.autocrlf
-            smudge_filter = get_smudge_filter(core_eol, autocrlf)
+            if eol in ("crlf", b"crlf"):
```
jelmer, line 263: type here should be predictable

jelmer (changes requested):

jelmer:
AI slop

PR #2409, opened 2026-09-14
More gitignore matching improvements

### LINKED ISSUE #2399: A `.gitignore` in a subdirectory loses to one closer to the root: 22 of 28 divergences from git on a 4,463-query bench

jelmer:
cc @KaizenShogun

Fixes #2399


KaizenShogun:
Measured this against the bench from #2399: **66 → 0** over 8,953 queries, every one of them, nothing new broken. That's the whole file-query side of my corpus gone.

The one place it goes the other way is directory queries, where a separate 1,010-query corpus goes **1 → 17**, all the same shape. Minimal case — `.gitignore` containing `__tmp/*` and `!__tmp/keep`:

| query | before | after | git |
|---|---|---|---|
| `__tmp/` | False | **True** | descends into it |
| `__tmp/keep` | False | False | not ignored |

`git check-ignore __tmp/` does answer "ignored" here, but only because `*` matches the empty string; `git status -uall` lists `__tmp/keep`, so git walks in. After this change `is_ignored` calls the directory ignored while calling a file inside it not ignored, so a caller that prunes on the directory answer drops the re-inclusion.

Bench and adapter: https://github.com/KaizenShogun/gitignore-conformance

PR #2411, opened 2026-09-14
Walk into directories whose contents alone are excluded

jelmer:
Git stops descending only when a pattern excludes a directory by name. A pattern like "dir/*" describes what the directory holds, so git walks in and a later "!dir/keep" still re-includes. check-ignore reports such a directory as ignored anyway, because "*" also matches the empty string, and dulwich's callers read that answer as permission to prune. "dulwich add dir" staged nothing and status hid the re-included file.

Add may_prune_directory, which asks whether a pattern names the directory itself, and prune on it in add() and get_untracked_paths(). is_ignored keeps its check-ignore semantics.


KaizenShogun:
The 17 directory divergences I reported on #2409 were mine, not a regression in your fix: I was
holding `is_ignored` against a "does git descend here?" oracle, which is the question this PR has
now given its own method. Measured `7604658a` against its parent `ee235e8e` on that same corpus —
1,010 directory queries taken from 33 repositories' real `.gitignore` files, oracle is git 2.55.0
answering a re-inclusion probe under each directory, cross-checked against `check-ignore -v` on the
bare path.

- `may_prune_directory`: **0 divergences** over 1,010, all decisive, none declined. 40 of those
  directories are prunable and 970 are not, so a constant answer doesn't pass it.
- `is_ignored` on the same 1,010: still 17, all one shape (`d/*` with a negation below) — and by
  your own test that is the right answer to check-ignore's question, which is why the number was
  never yours to fix.
- The file side stays where #2409 left it: 0 over 8,953 queries.

I also threw 23 hand-built boundary cases at it that the corpus doesn't contain — `*` + `!a/`,
`a/` + `!a/`, `**/a/`, `[ab]/`, a nested `.gitignore` contradicting the root one, a bare negation
with nothing excluding anything — same oracle, no disagreement. On six of those 23, `is_ignored`
and `may_prune_directory` answer differently, which is the split this PR is about.

Nothing to fix here. Just the number that doesn't come from the tests shipping in the same commit.

Issue #2414, opened 2026-09-16
log <path> does not implement history simplification


jelmer:
dulwich log <path> shows every commit that touched the path, whereas git log <path> prunes side branches by default. Dulwich's output matches git log --full-history <path>.

Git's default simplification: at a merge, if the path is unchanged (TREESAME) relative to some parent, git follows only that parent and drops the other branches, including commits on them that really did modify the path.

Example — merge tree takes bra's version of subtrees/f:

$ git log --format=%H -- subtrees/f
0a304ea

$ git log --full-history --format=%H -- subtrees/f
48d0b61   (merge)
0a304ea   (bra)
c38a631   (brb)

Dulwich returns all three; git's default returns one.

See the History Simplification section of git help rev-list.

PR #2421, opened 2026-09-18
Implement default history simplification for path-filtered walks

niukanen1:
"dulwich log <path>" kept every branch that had touched the path, i.e. it behaved like "git log --full-history <path>" rather than plain "git log <path>" (#2414).

The walker now applies git's default simplification: when a merge is TREESAME to one of its parents for the requested paths, that merge is not selected and only that parent's history is followed, pruning the other sides

Verified against C git on the criss-cross case from the issue, a merge TREESAME to its first parent, and a three-parent octopus merge; added a regression test for the first one. Existing expectations in test_paths_merge and the octopus test were checked against "git log" and updated to the simplified results


jelmer (changes requested):
Please avoid performance penalties for non-merge commits.

niukanen1:
Fixed in 4c72860: the TREESAME check is now restricted to merge commits (with an early return in the helper as well), so ordinary commits no longer trigger a tree diff.

niukanen1:
The remaining point from the Copilot review is addressed in 7130b6c: TREESAME is now judged on the path-filtered diff with rename detection disabled, so the simplification no longer depends on how the caller configures a rename detector. The regression test walks with `follow=True` and an explicit detector to pin this behavior down

PR #2423, opened 2026-09-21
Don't let cherry-pick discard uncommitted work tree changes

### LINKED ISSUE #2422: Dulwich cherry-pick overwrites local worktree changes
## Summary

`dulwich.porcelain.cherry_pick` silently replaces a caller's modified tracked
file or untracked file, even though the API has no force/overwrite argument.
It returns success; the attached Git control rejects the tracked-dirty case
and preserves the local bytes.

## Steps

```sh
unzip attachments.zip
cd attachments
DULWICH_DIR=../dulwich ./reproduce.sh dulwich-cherry-pick-repro
```

The output includes `tracked_dulwich_rc=0`,
`tracked_dulwich_local_bytes_lost=yes`, `untracked_dulwich_rc=0`, and
`control_git_preserved=yes`. The detailed source path and impact analysis are
in `report.md`.

## Version

Dulwich 1.2.15 (`b84b2ad43a5754054d812fa035b5fdcc68df4c29`), reproduced again
on current main (`74674b45ee45be9455c16d1c0fde6b355f118612`).

## Disclosure note

The report and reproduction code were assisted by an LLM. I personally
reviewed the analysis and confirmed the behavior with the attached reproducer.
I apologize if this LLM assistance creates extra work; the additional detail
is intended to make the behavior and fix location easier to verify.

**Attachments**

[report.md](https://github.com/user-attachments/files/32436073/report.md)

[attachments.zip](https://github.com/user-attachments/files/32436074/attachments.zip)

adarshsm:
Fixes #2422.

## The bug

`porcelain.cherry_pick()` replaces a tracked file the caller has edited but not
committed, and reports success:

```python
porcelain.init(d)
# commit file1.txt = "initial", branch "feature" commits file1.txt = "feature",
# then back on master:
open(f"{d}/file1.txt", "w").write("MY LOCAL WORK\n")

porcelain.cherry_pick(d, feature_commit)
# -> b'5e8fd720...'   (success)
open(f"{d}/file1.txt").read()
# -> 'feature\n'      (the local bytes are gone, and unrecoverable)
```

`git cherry-pick` refuses the same operation ("Your local changes to the
following files would be overwritten by merge").

## Root cause

`cherry_pick()` did two things before the safety check could run:

```python
r.get_worktree().reset_index(merged_tree.id, config=r.get_config_stack())
...
update_working_tree(..., allow_overwrite_modified=True, ...)
```

`WorkTree.reset_index()` goes through `build_index_from_tree()`, which writes
every file of the target tree to disk unconditionally — that is what destroys
the edit. And because `update_working_tree()` was then called with
`allow_overwrite_modified=True`, its `WorkingTreeModifiedError` guard was
disabled as well, so nothing was left to notice.

`merge()` already handles this correctly: `_do_merge()` does not call
`reset_index()` and calls `update_working_tree()` with the default
`allow_overwrite_modified=False`, which compares each changed path against HEAD
and raises before touching anything. `cherry_pick()` was the only caller taking
the other route.

## The fix

Drop the `reset_index()` call and the `allow_overwrite_modified=True` override,
so `cherry_pick()` updates the work tree exactly the way `merge()` does.
`update_working_tree()` already maintains the index, so the separate reset was
redundant on the clean path too.

Now:

```
WorkingTreeModifiedError: Your local changes to 'file1.txt' would be
overwritten by checkout. Please commit your changes or stash them before you
switch branches.
```

and `file1.txt` and `HEAD` are unchanged — matching how `merge()` and `pull()`
already behave (see `test_pull_with_uncommitted_changes`).

## Not covered here

The issue also mentions an **untracked** file being overwritten when the
cherry-pick adds a file of that name. That one is not specific to cherry-pick:
`update_working_tree()`'s pre-flight check only looks at `CHANGE_MODIFY` /
`CHANGE_DELETE`, so `merge()` and `checkout()` clobber untracked collisions the
same way. Fixing it means changing shared behaviour for every caller, which
seemed worth a separate decision rather than folding into this one. Happy to
follow up if you'd like it in the same change.

## Testing

- New `test_cherry_pick_keeps_uncommitted_changes` — verified it fails on
  `main` (`AssertionError: WorkingTreeModifiedError not raised`) and passes with
  the fix.
- Full suite: `pytest tests/ --ignore=tests/compat` → 3375 passed, 127 skipped,
  1 xfailed. (`tests/test_signature.py` has 10 failures on an unmodified
  checkout in this environment — missing gpg/x509 tooling — so it is deselected
  above; it is untouched by this change.)


jelmer:
Please remove AI slop from this PR, and adhere to CONTRIBUTING

jelmer:
You're clearly not reading feedback and purely submitting AI slop.

